PR #772, opened 2013-06-18
Remove warning caused by changes in minitest 5

vanstee:
Fixes the warning reading: "Warning: you should require
'minitest/autorun' instead."

Minitest 5 moved a few things around, such as the "MiniTest" to
"Minitest" module change, so those were updated as well.


allcentury:
Thanks @dpickett !  Worked for me as well.

PR #1141, opened 2014-08-10
Check for ActiveRecord configurations for fixture support

owenbendavies:
Currently ActiveRecord fixtures are included if ActiveRecord is defined,
but not set up. On Rails < 4.1.0 this was not a problem, as it had a
test in ActiveRecord::TestFixtures to see if ActiveRecord was
configured. On Rails 4.1.0 this check was removed, causing rspec-rails
to always try to connect to a Database even if there is not one.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -2,7 +2,7 @@ module RSpec
   module Rails
     # @private
     module FixtureSupport
-      if defined?(ActiveRecord::TestFixtures)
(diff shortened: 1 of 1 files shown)


bkimble:
I had this same issue, and after reading the bit about CarrierWave doing an explicit include of ActiveRecord, I looked deeper at what my gems were doing. After racking my brain, I noticed that someone added this:

gem 'activerecord-jdbcsqlite3-adapter', :platform => :jruby

To the Gemfile, which flipped ActiveRecord on, despite us not actually using it anywhere. Hopefully this will help others by giving them another (very obvious) place to look :)

PR #1232, opened 2014-11-24
Refactored HelperExampleGroup to use view context instances from a controller

crismali:
This way it's possible to call helper methods defined on a controller in the helper's methods.


samphippen:
@crismali I'm going to close this because it's both stale and conflicting. If you'd still like to see this merged, feel free to bring it up to date and we'll happily discuss whether it's ready to merge then :)

PR #1238, opened 2014-11-29
Make `instance_double` et al work with ActiveRecord columns.

JonRowe:
When we create an `instance_double` of an ActiveRecord object, we traditionally haven't been able to verify methods created by database columns. By calling a method on the class we can actually force Rails to create the dynamic methods on the class, and thus an `instance_double` will now work with a column method. This utilises a new extension hook in `rspec-mock`s that allows us to customise the creation process of verified doubles.

This comes out of work with @maxim-filimonov at Railscamp.


JonRowe:
This doesn't fail for me locally, any hints @cupakromer et al?

myronmarston:
I'm not sure what's causing the failure, but I have a suggested change that may fix it. I think you should wrap your config code in a `config.before(:suite)` block -- that way, it delays setting up the rspec-mocks callback until right before specs start to run.  That also gives the user a chance to configure an alternate `config.mock_with` option.  We used a similar technique here:

https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/extensions/active_record/proxy.rb

Also, I commented on this above, but I think this config should go in a self-contained file that we allow users to load w/o the rest of rspec-rails.

cupakromer:
Looks like all of the builds are failing. I'm looking into it today.

cupakromer:
I get the travis error locally after pulling the latest `master` branch for all repos. I'll dig in more a bit later.

cupakromer:
I'm not completely sure what's going on just yet. However, `defined?(::RSpec::Mocks)` returns `"constant"`. In rspec-core [the check](https://github.com/rspec/rspec-core/blob/c22e9023da/lib/rspec/core/configuration.rb#L1624-L1626) for if rspec-mocks is loaded is:

```ruby
def rspec_mocks_loaded?
  defined?(RSpec::Mocks.configuration)
end
```

Running this instead returns `nil`. It seems rspec-mocks isn't fully loaded until some time later. I haven't gotten to why yet. I probably won't have time to look at this again until tomorrow.

JonRowe:
Changed the hook as per @myronmarston's suggestion but now getting a different error?

lib/rspec/rails/configuration.rb
```diff
@@ -48,6 +48,15 @@ def self.initialize_configuration(config)
       config.include RSpec::Rails::ViewExampleGroup,       :type => :view
       config.include RSpec::Rails::FeatureExampleGroup,    :type => :feature
 
+      config.before :suite do
+        return unless defined?(ActiveRecord) && defined?(::RSpec::Mocks)
```
myronmarston: `return` can only be used within a method that hasn't yet exited, and that's not the case here -- `initialize_configuration` has exited long before this hook runs.   In this situation, it triggers the `LocalJumpError` you are seeing on travis.  You can use `break` to exit the block instead.

JonRowe: Then why does it work locally?

myronmarston: Is the `return` being called locally?  Try changing it to `return if true` and I imagine it will fail in the same way.

JonRowe: Fair enough, only changed it because I've seen rubocop complaining about it elsewhere lol :)

myronmarston:
BTW, @JonRowe, what do you think of my suggestion to move this bit of config into a self-contained file that can be loaded w/o the rest of rspec-rails in order to support people who sometimes run activerecord specs w/o all of rspec-rails loaded?

JonRowe:
Something to look at later perhaps? You have to do a lot of manual work to do that at the moment anyway...

myronmarston:
Seems trivial to put it in another file -- why wait?  I'm concerned that the primary case where you want to use test doubles (isolated specs) is the case where users may not have this loaded, leading to confusing differences in behavior between when rspec-rails is loaded and when it's not loaded.

cupakromer:
Is it possible to move the ActiveRecord check outside the block? That way if there's no AR at all we don't even create essentially empty hook?

myronmarston:
> Is it possible to move the ActiveRecord check outside the block? That way if there's no AR at all we don't even create essentially empty hook?

I think that would be possible, but I'm not sure it's desirable.  If the AR check happened outside the hook than this would only work for situations where AR is already loaded when this config is loaded.  Are we sure that's always the case?  Seems better to not require AR to be loaded first (which is not a requirement of the current implementation), given how light-weight `before(:suite)` hooks are, particularly since they only run once.

cupakromer:
> Are we sure that's always the case?

It's what we've done so far with all of the other checks. See just below for the [`ActionMailer` check](https://github.com/rspec/rspec-rails/blob/33099b2/lib/rspec/rails/configuration.rb#L60-L62).

rspec-rails expects the Rails environment to be loaded before it is. The default `spec/rails_helper.rb` loads it before loading rspec-rails. I have helped other debug a few issues where they were not loading Rails until later. We'd need to do a code audit to support delaying the Rails loading until later. I'm sure there are lots of places this could causes issues.

myronmarston:
Fair enough!

JonRowe:
@myronmarston's point is valid with the separate file though, as then we probably won't have Rails loaded.

lib/rspec/rails/configuration.rb
```diff
@@ -48,6 +48,15 @@ def self.initialize_configuration(config)
...
       config.include RSpec::Rails::FeatureExampleGroup,    :type => :feature
 
+      config.before :suite do
+        # This allows dynamic columns etc to be used on ActiveRecord models when creating instance_doubles
+        unless defined?(ActiveRecord) && defined?(::RSpec::Mocks)
+          ::RSpec::Mocks.configuration.when_declaring_verifying_double do |possible_model|
+            possible_model.target.define_attribute_methods if possible_model.target.respond_to?(:define_attribute_methods)
+          end
+        end
+      end
```
myronmarston: I'm surprised you left this in place -- now this config is duplicated in two places.  Can we just delegate to the version of this you put in the other file from here?

JonRowe: Er... oops? :)

cupakromer:
> @myronmarston's point is valid with the separate file though, as then we probably won't have Rails loaded.

Agreed. :+1:

cupakromer:
I fixed the broken cuke. Hopefully this will help things pass.

cupakromer:
Ok, still failing but I see why. Two reasons:

1. With the latest changes the cuke does not have the mocks configuration set to define attribute methods.
2. Updates to rspec-core's output if failing the feature cuke.

Issue 2 is fixed on `master`, so we just need to rebase this PR.

I think the best option for issue 1 is a two-step approach:

- Add the following to the cuke showing how enable the setting (for existing projects):

  ```ruby
  RSpec.configuration.mock_with(:rspec) do |config|
    config.when_declaring_verifying_double do |reference|
      reference.target.define_attribute_methods
    end
  end
  ```

- For new projects we should update the default `spec/rails_helper.rb` so that it includes this configuration option if `ActiveRecord` is available. We already have several hooks checking for AR so this should be easy to include.

Thoughts?

myronmarston:
> With the latest changes the cuke does not have the mocks configuration set to define attribute methods.

It's unclear to me why this is, given that loading rspec-rails loads `rspec/rails/activerecord` (the new file) and in that file we configure this when it is loaded.  Can you explain why it's not set for the cuke?

> Add the following to the cuke showing how to enable the setting (for existing projects):

There's no guarantee that `reference.target` is an ActiveRecord::Base subclass.  I expect users to use verifying doubles for a variety of ruby classes including many with no superclass (besides the default of `Object`).  If we were to recommend a snippet of code like this, it would need to do a `respond_to?` check or a `ActiveRecord::Base == reference.target` check or something.

> For new projects we should update the default spec/rails_helper.rb so that it includes this configuration option if ActiveRecord is available. We already have several hooks checking for AR so this should be easy to include.

It sounds like you are recommending that we move in the direction of not configuring this automatically for users but instead put that responsibility on them in their `spec/rails_helper.rb`.  Why is putting the responsibility on them better?  I would think that it's something rspec-rails should take care of for users.  After all, it's not necessary to use rspec-rails to test rails applications with rspec....it's simply that rspec-rails takes care of these kind of integration points for the user so they don't have to.  Once we start moving the responsibility for these sorts of things back on the user it seems to go against the entire reason for rspec-rails to exist.

cupakromer:
I some how missed the line where we loaded the config. I was expecting it at the end of the file not in the module definition. Having now found that I see the root cause. We need to change the `unless` to an `if` check.

Since everything will now be auto-configured no need for any other changes.

JonRowe:
Derp :)

cupakromer:
LGTM :shipit:

PR #1239, opened 2014-11-30
Extend rspec-mocks to see store_accessor methods

taryneast:
Rails has a number of ways of adding dynamic methods to models.

Verifying doubles in mocks currently fail on all dynamic methods that aren't explicitly set in method_missing, because the model does not recognise them as a method at the time the mock verifies.

However - many of Active Record's dynamic methods have a very fixed and easy-to-check list... so this extension does that and adds these dynamic methods to the allowed set of valid methods

This commit starts that ball rolling by recognising the "store_accessor" methods (which (de)serialise columns on an Active Record)

Given that an extension to rspec-mocks is something that looks like it hasn't been done in this repo before, I took a wild guess at a location for it that would be appropriate. I am totally ok with moving/renaming or otherwise updating this code to match a preferred location/naming-structure/whatever... 

Note: this change came out of an issue that I first mentioned on rspec-mocks where I was directed to rspec-rails as a better location: https://github.com/rspec/rspec-mocks/issues/830


cupakromer:
@taryneast just a friendly :smile_cat: ping to see how things are going on this and if you need any help.

samphippen:
Going to close this because it hasn't been updated since jan 17. @taryneast Feel free to submit this as a **new pr** if you want to get this feature integrated.

Issue #773, opened 2013-06-19
Undefined method `message`


mattconnolly:
I'm seeing some undefined method `message` errors using rspec with rails 4.0.0rc2. I'm using spec 2.13.

I'm not sure if this is the right place to flag this... I saw the issue closed in the rails project:

https://github.com/rails/rails/issues/9085

mkolenda:
I'm in a Catch-22 situation with this issue.  

With gem 'test-unit' I get the error undefined method `message' for....  
Without the 'test-unit' gem I get the error ..2.2.0/gems/activesupport-4.0.13/lib/active_support/dependencies.rb:229:in `require': cannot load such file -- test/unit/assertions (LoadError). (this issue https://github.com/rspec/rspec-rails/issues/1262)

I'm running: Ruby 2.2.0, rspec-rails (2.13.1), test-unit (3.0.8), rails (4.0.13)

cupakromer:
Support for Ruby 2.2 on Rails 4.0 does not start until RSpec 3.2.

cupakromer:
The other issue is specifically about using a routing matcher in a controller spec. Moving the spec into a route spec should resolve the issue for the time being.

PR #1264, opened 2014-12-30
Ruby 2.2 on Rails 4.0 support

### LINKED ISSUE #1262: Routing specs undefined method 'message' on Ruby 2.2
When upgrading from Ruby 2.1.5 to Ruby 2.2 I see test failures in all our routing specs.

```
Failure/Error: expect(get("/time_sheets")).to route_to("time_sheet/time_sheets#day")
     NoMethodError:
       undefined method `message' for #<#<Class:0x007fa9ab213160>:0x007fa998fe2948>
```

This is using rspec-rails 3.1.0 and **Rails 4.0.12**. I tried testing against rspec-rails master but it appears to be a bit convoluted.

```
Could not find gem 'rspec-core (= 3.2.0.pre) ruby', which is required by gem 'rspec-rails (>= 0) ruby', in any of the
sources.
```

I see you are already testing against Ruby 2.2 and Travis CI is passing. Curious.

cupakromer:
In Ruby 2.2 Minitest has been removed from ruby core. However, Rails 4.0.x includes it as a gem dependency. In order to not fail on loading our adapter this performs a second check for Ruby 2.2+ which then handles the Minitest 4.x gem.

Resolve #1262

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -14,6 +14,12 @@ module RSpec
       # Constant aliased to either Minitest or TestUnit, depending on what is
       # loaded.
       Assertions = Minitest::Assertions
+    elsif RUBY_VERSION >= '2.2.0'
+      # Minitest has been removed from ruby core. However, we must be on an
+      # old Rails version and must load the Minitest 4.x gem
+      gem 'minitest' if defined?(Kernel.gem)
+      require 'minitest/unit'
+      Assertions = MiniTest::Assertions
     else
       begin
         require 'test/unit/assertions'


lib/rspec/rails/adapters.rb
```diff
@@ -14,6 +14,12 @@ module Rails
       # Constant aliased to either Minitest or TestUnit, depending on what is
       # loaded.
       Assertions = Minitest::Assertions
+    elsif RUBY_VERSION >= '2.2.0'
+      # Minitest has been removed from ruby core. However, we must be on an
+      # old Rails version and must load the Minitest 4.x gem
+      gem 'minitest' if defined?(Kernel.gem)
```
raj454raj: I am getting 
```
undefined method `message' for #<RSpec::Core::ExampleGroup::Nested_3::Nested_1:0x007fbb75fc4c38>
```
I am on Rails 4.1.15, Ruby 2.2.5 and RSpec 2.12.0 (Yes it is too old. I am migrating from Rails 3.2.22).
I have added minitest(5.9.0) and test-unit(3.2.0) both because RSpec won't run without them.
At some [places](https://github.com/rspec/rspec-rails/issues/773#issuecomment-20241398) it is written to remove test-unit, but then I won't be able to run RSpec at all. Any help ?
Also RSpec 2.12.0 is not a problem for Rails 4.1.15 yeah ?

JonRowe: RSpec 2.12 is not supported so it's quite possible Rails 4.1.15 won't work.

raj454raj: Cool Thanks ! 👍

Issue #1187, opened 2014-10-07
DEPRECATION WARNING: `named_routes.helpers` is deprecated, please use `route_defined?(route_name)` to see if a named route was defined. (called from block in it_requires_authentication


Vorob-Astronaut:
Hi guys. I wrote `rspec spec` And I had:

```
DEPRECATION WARNING: `named_routes.helpers` is deprecated, please use `route_defined?(route_name)` to see if a named route was defined. (called from block in it_requires_authentication at /Users/vorob/Dropbox/ruby/upnetwork/spec/support/controllers/errors_helper.rb:7)
```

See my _/Users/vorob/Dropbox/ruby/upnetwork/spec/support/controllers/errors_helper.rb_

``` ruby
module Controllers
  module ErrorsHelper
    def it_requires_authentication(&block)
      it 'requires authentication' do
        instance_exec(&block)

        expect(response).not_to be_successful
        expect(flash[:alert]).to eq(I18n.t('devise.failure.unauthenticated'))
      end
    end

    def it_raises_authorization_error(in_case = nil, &block)
      it "raises authorization error #{in_case}" do
        instance_exec(&block)

        expect(response).not_to be_successful
        expect(flash[:alert]).to eq(I18n.t('errors.not_authorized'))
      end
    end
  end
end
```

I use

```
  gem 'rspec-rails', '~> 3.0'
```

arthurnn:
is this on rspec 2 too?

cupakromer:
@arthurnn it is not, we are not supporting Rails 4.2 on RSpec 2. If you are looking for help upgrading to RSpec 3 check out the main [upgrade guide](https://relishapp.com/rspec/docs/upgrade) the [rails specific changes](https://www.relishapp.com/rspec/rspec-rails/v/3-2/docs/upgrade) and the amazing tool [transpec](http://yujinakayama.me/transpec/) which automates a huge portion of the process.

arthurnn:
@cupakromer thanks. I will do that.

PR #1266, opened 2014-12-31
Update gemspec capping Rails at 4.2

cupakromer:
According to the Rails blog the 4.2 release is the last in the 4.x line.
Plans for Rails 5.x support are still in the works and may not be on
rspec-rails 3.x.

To inform users and prevent accidental usage this limits the allowed
Rails versions in the gemspec.

- http://weblog.rubyonrails.org/2014/12/19/Rails-4-2-final/

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -27,9 +27,9 @@ Gem::Specification.new do |s|
     s.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]
   end
 
-  s.add_runtime_dependency(%q<activesupport>, [">= 3.0"])
-  s.add_runtime_dependency(%q<actionpack>, [">= 3.0"])
-  s.add_runtime_dependency(%q<railties>, [">= 3.0"])
+  s.add_runtime_dependency(%q<activesupport>, [">= 3.0", "<= 4.2"])
+  s.add_runtime_dependency(%q<actionpack>, [">= 3.0", "<= 4.2"])
+  s.add_runtime_dependency(%q<railties>, [">= 3.0", "<= 4.2"])
   %w[core expectations mocks support].each do |name|
     if RSpec::Rails::Version::STRING =~ /[a-zA-Z]+/ # prerelease builds
       s.add_runtime_dependency "rspec-#{name}", "= #{RSpec::Rails::Version::STRING}"


rspec-rails.gemspec
```diff
@@ -27,9 +27,9 @@ Gem::Specification.new do |s|
     s.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]
   end
 
-  s.add_runtime_dependency(%q<activesupport>, [">= 3.0"])
-  s.add_runtime_dependency(%q<actionpack>, [">= 3.0"])
-  s.add_runtime_dependency(%q<railties>, [">= 3.0"])
+  s.add_runtime_dependency(%q<activesupport>, [">= 3.0", "<= 4.2"])
+  s.add_runtime_dependency(%q<actionpack>, [">= 3.0", "<= 4.2"])
+  s.add_runtime_dependency(%q<railties>, [">= 3.0", "<= 4.2"])
```
myronmarston: While the rails team has publicly announced that 4.2 will be the last 4.x release, wouldn't it be slightly safer to do `< 5.0`?  If for some reason they decided to release 4.3 that would allow that version.  Unlikely, I know, but if your intention here is to explicitly not support rails 5+ that seems more semantically correct.  (OTOH, if you explicitly would not want to support 4.3 if that was ever released, this is fine).

cupakromer: Recently, we've needed to add some updates for the new releases. I believe they have been most ERs, but one or two deprecations came up. I would feel safer locking this down to avoid the "I updated and it broke" scenario. This way our support is explicit and Bundler / Rubygems will inform people of mismatches ahead of time. See history in #1187 and #1192.

cupakromer: If say a 4.3 does come out and we don't need code changes, I see no problems in releasing patch dot release bumping the gem support.

jasnow:
Can't label this issue's label "rails5"

cupakromer:
This is merged. I don't consider it a "rails5" issue

jasnow:
ok

jasnow:
Just a FYI: The reason I found this is because
 * when I upgrade my repos to Rails 5 and 
 * they have rspec-rails in Gemfile

they use **rspec-rails 3.1.0** because it was the last one to not have a "<= 4.x" dependency.

JonRowe:
Theres not a whole heap we can do about that, maybe we'd be better off without the gem dependency but have a runtime warning thats printed when Rails 5 is detected?

cupakromer:
I still think it's better to have the dependency locked in. It's unfortunate that we didn't do that initially with the 3.x release. Given the nature of rspec-rails, where we do interact with various Rails private APIs, I am in favor of erroring on the side of being explicit in our version support. This way users don't accidentally update Rails and things "break" because we are not supporting that Rails version on their current rspec-rails version.

I feel Gem dependencies are the proper tool for this situation.

JonRowe:
> I feel Gem dependencies are the proper tool for this situation.

So do I, but I feel we can't do that given we have releases out there which don't have this dependency, it'll just force them backwards...

cupakromer:
> I feel we can't do that given we have releases out there which don't have this dependency

I understand that. Though those releases are 3.0 and 3.1. The only current Rails version that those do not technically support if 4.2; which has support in 3.2 and 3.3. Rails 5 is a bit different and not out yet; nor in beta.

I don't see the benefit in changing this behavior currently as only edge Rails has this issue. Which is something I am currently looking into.

JonRowe:
The benefit is we won't spend so much time responding to issues with "Rails 5 isn't supported" :/

PR #1267, opened 2014-12-31
Backport setup changes

cupakromer:


myronmarston:
Given our gem signing cert expired, and the confusion that could ensue from some 3.1.x versions being signed with one cert and others with a different cert, I've been thinking we won't have any more 3.1.x releases.  As long as we maintain SemVer in 3.2+ that shouldn't be a problem for users.

All that's to say you probably don't need to put any significant effort into this here :).

cupakromer:
Agreed. The build also failed due to 2.2 needing additional support which is listed as an ER.

Closing this.

PR #1276, opened 2015-01-13
Fix ignoring default_url_options

### LINKED ISSUE #1275: Default_url_options is being ignored
In my rspec_helper.rb I defined;

```
Rails.application.routes.default_url_options[:host] = 'lvh.me'
```

It works fine for non-feature specs but doesn't work for feature specs. Host for urls are equal to "www.example.com" in feature specs. 

I've debugged and in _feature_example_group.rb_ file, _default_url_options_ is always an empty hash.

https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/example/feature_example_group.rb#L18

 found the following solution for this issue;
https://github.com/beydogan/rspec-rails/commit/b590c82573b27bd8a1be967811b4835119260f09

Is this valid?

beydogan:
Fixes #1275


cupakromer:
Closing per the discussion in the original issue.

PR #1277, opened 2015-01-13
Prepare to support Ruby 2.2 on Rails 3.2.22+

cupakromer:
This is in preparation for the next patch release. Similar steps are
being taken by Rails core. See:

https://github.com/rails/rails/commit/8f92edb2b4b6beb5c778283be7cbcef6bf7e596c

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -1,6 +1,7 @@
 require 'delegate'
 require 'active_support'
 require 'active_support/concern'
+require 'active_support/core_ext/string'
 
 module RSpec
   module Rails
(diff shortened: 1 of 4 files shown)


lib/rspec/rails/adapters.rb
```diff
@@ -15,11 +15,31 @@ module Rails
...
-      require 'minitest/unit'
-      Assertions = MiniTest::Assertions
+      # Minitest / TestUnit has been removed from ruby core. However, we are
+      # on an old Rails version and must load the appropriate gem
+      if ::Rails::VERSION::STRING >= '4.0.0'
+        # ActiveSupport 4.0.x has the minitest '~> 4.2' gem as a dependency
+        gem 'minitest' if defined?(Kernel.gem)
+        require 'minitest/unit'
+        Assertions = MiniTest::Assertions
+      elsif ::Rails::VERSION::STRING >= '3.2.21'
```
myronmarston: 3-part versions compare differently than their string form does, particular when you have parts with multiple digits.  I think it's OK here because the 2-digit part is the last part...but thought I'd bring it up as something to be aware of and careful of.

myronmarston:
LGTM

keepcosmos:
+1

cupakromer:
An odd `Test::Unit` runner failer I was not previously seeing locally. Will need to dig into that. Also seems  a Ruby bug with something Travis CI has installed :frowning:

keepcosmos:
Please check, I have same issue too

cupakromer:
Yep I get that error now. An update on one of the other repos is likely related. I'll look into and fix this today.

rohandaxini:
:+1:

PR #1278, opened 2015-01-13
Fix failing cuke with new rspec-core changes.

cupakromer:
Minor update with new output changes.


cupakromer:
Will merge once the sanity check is green.

myronmarston:
Oh right, our rspec-core changes broke this.  Sorry about that.  We run the rspec-rails specs from the other repo builds, but not the cukes (since they are so much more involved and slower).  If this becomes a frequent occurrence we can consider changing our build to run the cukes, but I'd rather not do that just yet.

PR #1283, opened 2015-01-17
Cache bundle directory.

cupakromer:
This also re-orders the listing of the settings to align with the
`rspec-dev` order. This makes manually comparing the files easier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,36 @@
 language: ruby
 
+sudo: false
+
+# We deviate from the rspec-dev cache setting here.
+#
+# Travis has special bundler support where it knows to run `bundle clean`
+# before backing up the bundle so it doesn't grow indefinitely.
+#
+# For the other repos we changed to specifying the directory specifically to
+# avoid the `bundle clean` travis was doing, because Myron found that, when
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,5 +1,22 @@
 language: ruby
 
+sudo: false
+
+cache:
+  directories:
+    - ../bundle
```
myronmarston: You may want to do `cache: bundler` here rather than listing the directory.  Travis has special bundler support where it knows to run `bundle clean` before backing up the bundle so it doesn't grow indefinitely.

For the other repos we changed to specifying the directory specifically to avoid the `bundle clean` travis was doing, because I found that, for example, when running the rspec-mocks build, the "run the rspec-core and rspec-rails spec suites" parts would take a while because it would _always_ have to install fresh gems and they weren't in the cache.  This was happening because travis was running `bundle clean` (due to our `cache: bundler` config) and running `bundle clean` against rspec-mocks was pruning gems installed for the other repos like rspec-core and rspec-rails.

In your case, if you're not running the spec suites for the other repos (which I doubt -- given they don't depend on rspec-rails, why would you?) then `cache: bundler` is probably a better option.

PR #1292, opened 2015-01-30
Change type meta tag in templates to Hash 1.9 syntax

mikeastock:
I often go into specs created by the generators and change the hash syntax. Does rspec-rails plan to support Ruby 1.8.7 much longer?

### DIFF
--- a/lib/generators/rspec.rb
+++ b/lib/generators/rspec.rb
@@ -7,6 +7,8 @@ module Rspec
   module Generators
     # @private
     class Base < ::Rails::Generators::NamedBase
+      include RSpec::Rails::FeatureCheck
+
       def self.source_root(path = nil)
         if path
           @_rspec_source_root = path
--- a/spec/generators/rspec/controller/controller_generator_spec.rb
+++ b/spec/generators/rspec/controller/controller_generator_spec.rb
@@ -19,7 +19,7 @@ describe Rspec::Generators::ControllerGenerator, :type => :generator do
       describe 'the spec' do
         it { is_expected.to exist }
         it { is_expected.to contain(/require 'rails_helper'/) }
-        it { is_expected.to contain(/^RSpec.describe PostsController, :type => :controller/) }
+        it { is_expected.to contain(/^RSpec.describe PostsController, #{type_metatag(:controller)}/) }
       end
     end
     describe 'skipped with a flag' do
@@ -72,13 +72,13 @@ describe Rspec::Generators::ControllerGenerator, :type => :generator do
           subject { file('spec/views/posts/index.html.erb_spec.rb') }
           it { is_expected.to exist }
           it { is_expected.to contain(/require 'rails_helper'/) }
-          it { is_expected.to contain(/^RSpec.describe "posts\/index.html.erb", :type => :view/) }
+          it { is_expected.to contain(/^RSpec.describe "posts\/index.html.erb", #{type_metatag(:view)}/) }
         end
         describe 'show.html.erb' do
           subject { file('spec/views/posts/show.html.erb_spec.rb') }
           it { is_expected.to exist }
           it { is_expected.to contain(/require 'rails_helper'/) }
-          it { is_expected.to contain(/^RSpec.describe "posts\/show.html.erb", :type => :view/) }
+          it { is_expected.to contain(/^RSpec.describe "posts\/show.html.erb", #{type_metatag(:view)}/) }
         end
       end
       describe 'with haml' do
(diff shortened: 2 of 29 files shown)


JonRowe:
We'll drop support for 1.8.7 when we drop support for Rails 3.2, that said I think we'd accept this if you either: added an option to make this opt-in, or, made it sensitive to the Rails version.

mikeastock:
By opt-in, do you mean a config you set?

JonRowe:
Yes

>

mikeastock:
So I went ahead and added a helper that uses 1.9 syntax is the ruby version is greater than `1.8.7`. Let me know if that's what you had in mind

lib/generators/rspec.rb
```diff
@@ -20,6 +20,16 @@ def module_namespacing
...
+
+      protected
+
+      def type_metatag(type)
+        if RUBY_VERSION > '1.8.7'
+          "type: :#{type}"
+        else
+          ":type => :#{type}"
+        end
+      end
```
cupakromer: Why is this listed as `protected`?

spec/support/helpers.rb
```diff
@@ -8,5 +8,13 @@ def with_isolated_config
     RSpec.configuration = original_config
   end
 
+  def type_metatag(type)
+    if RUBY_VERSION > '1.8.7'
+      "type: :#{type}"
+    else
+      ":type => :#{type}"
+    end
+  end
```
cupakromer: This should only be defined once. We have a place for similar checks: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/feature_check.rb

spec/generators/rspec/controller/controller_generator_spec.rb
```diff
@@ -19,7 +19,7 @@
       describe 'the spec' do
         it { is_expected.to exist }
         it { is_expected.to contain(/require 'rails_helper'/) }
-        it { is_expected.to contain(/^RSpec.describe PostsController, :type => :controller/) }
+        it { is_expected.to contain(Regexp.new("^RSpec.describe PostsController, #{type_metatag(:controller)}")) }
```
cupakromer: We use either `/regex/` or `%r{regex}` could you use one of those instead of `Regexp.new` please. :heart:

cupakromer:
@mikeastock Thanks for taking care of this. :blue_heart: :blue_heart: It looks like the specs are failing due to Rubocop with `protected`. I think if you move that into the shared `feature_check` module we can eliminate the `protected` call.

How do you feel about making the check based on the Rails version? It looks like Rails 4.0 uses the new hash syntax. I'd rather mirror the Rails version so we are consistent with the other generated Rails files.

Does Rails have a way to configure the hash syntax? I found http://stackoverflow.com/a/8565952 but I wasn't able to get that to work for me with Rails 4.0.

mikeastock:
@cupakromer Thanks for the comments. I like the idea of checking the Rails version vs the Ruby version so I'll make that swap tonight as well as address the rest of the comments.

Rails used to have a way to configure hash syntax but it looks like it was removed in Rails 4.0+. I guess we will need to keep our hash syntax implementation around.

lib/generators/rspec.rb
mikeastock: Thor throws an exception regarding arity if the method is public. Is it best to put this method in https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/feature_check.rb?

mikeastock:
@cupakromer let me know if the change fit what you had in mind. Thanks!

spec/generators/rspec/controller/controller_generator_spec.rb
```diff
@@ -72,13 +72,13 @@
           subject { file('spec/views/posts/index.html.erb_spec.rb') }
           it { is_expected.to exist }
           it { is_expected.to contain(/require 'rails_helper'/) }
-          it { is_expected.to contain(/^RSpec.describe "posts\/index.html.erb", :type => :view/) }
+          it { is_expected.to contain(/^RSpec.describe \"posts\/index.html.erb\", #{type_metatag(:view)}/) }
```
cupakromer: Is there a reason to escape the quotes when they weren't before?

cupakromer:
Looks good. Just posted one last question. Also, would be nice to just rebase this into a single commit to clean up the history a little.

mikeastock:
Everything should be good to go, removed the escaping and rebased.

cupakromer:
Awesome! Thanks again @mikeastock just in time for the next release too.

PR #1303, opened 2015-02-09
Updates from rspec-dev (2015-02-08)

myronmarston:


cupakromer:
This LGTM, which associated PR was this from?

myronmarston:
rspec/rspec-dev#115.

Although the diff is much larger here because there was a prior rspec-support PR that we chose not to merge there (couldn't get it to do what I was trying to have it do in the other repos) but it looks like it got merged here (but probably never affected the build given the build differences).

PR #1307, opened 2015-02-13
Fix `be_routable` description -- Renames BeRoutableMatcher to BeRoutable

tonyta:
**Problem**
Previously, using `be_routable` in a one-liner...
``` ruby
RSpec.describe 'a not-routable route', type: :routing do
  let(:sad_route) { '/this/route/makes/me/sad' }
  specify { expect(get: sad_route).not_to be_routable }
end
```
...will output the following:
```
a not-routable route
  should not be routable matcher
```
`BaseMatcher#description` turns `BeRoutableMatcher` into "be routable matcher".

**Solution**
The solution implemented here is to rename `BeRoutableMatcher` to `BeRoutable` in order to rely on `BaseMatcher#description` to correctly return the following description:
```
a not-routable route
  should not be routable
```
This looks like the precedence in other matchers like `BeValid`.

**Alternate Solution**
Alternatively, `#description` could be overridden in `BeRoutableMatcher`:
``` ruby
def description
  "be routable"
end
``` 
I'd be happy to change PR if that's preferred :smile_cat:

### DIFF
--- a/lib/rspec/rails/matchers/routing_matchers.rb
+++ b/lib/rspec/rails/matchers/routing_matchers.rb
@@ -63,7 +63,7 @@ module RSpec
         end
 
         # @private
-        class BeRoutableMatcher < RSpec::Matchers::BuiltIn::BaseMatcher
+        class BeRoutable < RSpec::Matchers::BuiltIn::BaseMatcher
           def initialize(scope)
             @scope = scope
           end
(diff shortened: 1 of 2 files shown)


JonRowe:
I'm not sure what @cupakromer will think but personally I'd rather see:

>**Alternate Solution**
Alternatively, `#description` could be overridden in `BeRoutableMatcher`:
``` ruby
def description
  "be routable"
end
```

cupakromer:
I agree with @JonRowe. Let's define `description`.

tonyta:
Thanks @JonRowe @cupakromer. I'm closing this PR and redefining `#description` in https://github.com/rspec/rspec-rails/pull/1310

PR #1310, opened 2015-02-13
Fix `be_routable` description -- Defines BeRoutableMatcher#description

tonyta:
Alternate implementation of https://github.com/rspec/rspec-rails/pull/1307
Thanks for the feedback @JonRowe @cupakromer! :joy_cat: 

**Problem**
Previously, using `be_routable` in a one-liner...
``` ruby
RSpec.describe 'a not-routable route', type: :routing do
  let(:sad_route) { '/this/route/makes/me/sad' }
  specify { expect(get: sad_route).not_to be_routable }
end
```
...will output the following:
```
a not-routable route
  should not be routable matcher
```
`BaseMatcher#description` turns `BeRoutableMatcher` into "be routable matcher".

**Solution**
`BeRoutableMatcher#description` is defined to override `BaseMatcher#description`...
``` ruby
class BeRoutableMatcher < RSpec::Matchers::BuiltIn::BaseMatcher
  # ...
  def description
    "be routable"
  end
end
``` 
...to correctly output:
```
a not-routable route
  should not be routable
```

### DIFF
--- a/lib/rspec/rails/matchers/routing_matchers.rb
+++ b/lib/rspec/rails/matchers/routing_matchers.rb
@@ -84,6 +84,10 @@ module RSpec
           def failure_message_when_negated
             "expected #{@actual.inspect} not to be routable, but it routes to #{@routing_options.inspect}"
           end
+
+          def description
+            "be routable"
+          end
         end
 
         # Passes if the route expression is recognized by the Rails router based on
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks @tonyta

tonyta:
Thanks, @JonRowe! Happy to help! :octopus:

PR #1314, opened 2015-02-18
Patch mocks to properly handle JSON monkey patches

### LINKED ISSUE #1309: calling .to_json on double.as_null_object causes stack level too deep in rails 4.1+
Having this issue in rails 4.2 on my own application. promised @samphippen I'd write an issue.

at https://github.com/chav02/rspec_to_json I've prepared a minimal rails application that showcases this for rails 4.0.x (working) and 4.1.x (splosions!)

cupakromer:
Fix #1309

Rails adds a nearly global monkey patch to make converting objects into
JSON easier. It adds both `to_json` and `as_json` methods on nearly all
core object types; including `Object` but not `BasicObject`.

That prevents a RSpec double or spy from working as expected. Since the
double/spy object inherits these patches from `Object` they are not able
to intercept them in the normal `method_missing` manner.

We work around that by applying a monkey patch directly on the RSpec
mocks modules. This is necessary, instead of creating a new module and
including it, because Ruby does not recognize new modules included on an
existing module for classes which have already included the base module.

Additionally, we must copy some of the `private` internals of the mock
in order to preserve behavior. This is not a desirable approach but is
necessary.

/cc @chav02 @JonRowe @myronmarston

### DIFF
--- a/lib/rspec/rails/extensions/active_support/json.rb
+++ b/lib/rspec/rails/extensions/active_support/json.rb
@@ -1,0 +1,67 @@
+RSpec.configure do |c|
+  c.before(:suite) do
+    if defined?(::RSpec::Mocks)
+      # Need to require the base file to setup internal support require helpers
+      require 'rspec/mocks'
+      require 'rspec/mocks/test_double'
+      ::RSpec::Mocks::TestDouble.module_exec do
+        # @private
+        def as_json(*args, &block)
+          return method_missing(:as_json, *args, &block) unless null_object?
+          __mock_proxy.record_message_received(:as_json, *args, &block)
+          nil
+        end
+
+        # @private
+        def to_json(*args, &block)
+          return method_missing(:to_json, *args, &block) unless null_object?
+          __mock_proxy.record_message_received(:to_json, *args, &block)
+          "null"
+        end
+
+        # @private
+        remove_method(:respond_to?)
+        def respond_to?(message, incl_private = false)
+          return true if null_object?
+          if [:to_json, :as_json].include?(message)
+            public_methods(false).include?(message)
+          else
+            super
+          end
+        end
+      end
+
+      require 'rspec/mocks/verifying_double'
+      ::RSpec::Mocks::VerifyingDouble.module_exec do
+        # @private
+        def as_json(*args, &block)
+          verify_on_mock_proxy(:as_json, *args)
+          super
+        end
+
+        # @private
+        def to_json(*args, &block)
+          verify_on_mock_proxy(:to_json, *args)
+          super
+        end
+
+      private
+
+        def verify_on_mock_proxy(message, *args)
+          # Null object conditional is an optimization. If not a null object,
+          # validity of method expectations will have been checked at
+          # definition time.
+          return unless null_object?
+
+          if @__sending_message == message
+            __mock_proxy.ensure_implemented(message)
+          else
+            __mock_proxy.ensure_publicly_implemented(message, self)
+          end
+
+          __mock_proxy.validate_arguments!(message, args)
+        end
+      end
+    end
+  end
+end
--- a/spec/rspec/rails/extensions/active_support/json_spec.rb
+++ b/spec/rspec/rails/extensions/active_support/json_spec.rb
@@ -1,0 +1,196 @@
+require "spec_helper"
+require "active_support"
+if ::Rails::VERSION::STRING > '4.1'
+  require "active_support/core_ext/object/json"
+else
+  require "active_support/core_ext/object/to_json"
+end
+require "rspec/rails/extensions/active_support/json"
+
+RSpec.describe "Patch mocks to work with Active Support's json patches" do
+
+  shared_examples "verifies the patch arguments" do
+    it "verifies arguments" do
+      allow(subject).to receive(patch)
+      expect { subject.send(patch, :too, :many) }.to raise_error(ArgumentError)
+    end
+  end
+
+  shared_examples "intercepts Active Support's patch" do |message, opts|
+    opts ||= {}
+    context "intercepts Active Support's ##{message}" do
+      defaults = { :to_json => "null", :as_json => nil }
+
+      let(:patch) { message }
+      let(:default) { defaults[patch] }
+
+      it "responds to the message" do
+        expect(subject).to respond_to(patch)
+      end
+
+      it "defines a sane default implementation" do
+        expect(subject.send(patch)).to eq(default)
+      end
+
+      it "records the message being received" do
+        subject.send(patch)
+        expect(subject).to have_received(patch)
+      end
+
+      it "doesn't interfere with a stubbed response" do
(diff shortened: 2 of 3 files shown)


cupakromer:
FWIW locally I added this code directly into rspec-mocks and run the specs, it was all green there.

lib/rspec/rails/extensions/active_support/json.rb
```diff
@@ -0,0 +1,63 @@
+if defined?(::RSpec::Mocks)
```
myronmarston: rspec-core waits to load rspec-mocks until just before the first example group is defined:

https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core/example_group.rb#L376-L384
https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core/example_group.rb#L435-L443
https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core/configuration.rb#L1209-L1212
https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core/configuration.rb#L412
https://github.com/rspec/rspec-core/blob/v3.2.0/lib/rspec/core/configuration.rb#L535

Thus, rspec-mocks may not be loaded at the point this file is loaded.  I think this could go in a `before(:suite)` hook in order to be delayed until after the point where rspec-mocks will have been loaded if it is going to be.

lib/rspec/rails/extensions/active_support/json.rb
```diff
@@ -0,0 +1,63 @@
...
+  # Need to require the base file to setup internal support require helpers
+  require 'rspec/mocks'
+  require 'rspec/mocks/test_double'
+  ::RSpec::Mocks::TestDouble.module_exec do
+    # @private
+    def as_json(*args, &block)
+      return method_missing(:as_json, *args, &block) unless null_object?
```
myronmarston: I think this can be `return super unless null_object?` (and same below in `to_json`) because `method_missing` would be called naturally.

lib/rspec/rails/extensions/active_support/json.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    # @private
+    def to_json(*args, &block)
+      return method_missing(:to_json, *args, &block) unless null_object?
+      __mock_proxy.record_message_received(:to_json, *args, &block)
+      "null"
+    end
+
+    # @private
+    remove_method(:respond_to?)
```
myronmarston: I'm concerned that if/when we change the `respond_to?` logic in rspec-mocks, it'll get ignored here in rspec-rails because you remove and redefine it.

lib/rspec/rails/extensions/active_support/json.rb
cupakromer: It won't. Rails adds `as_json` and `to_json` onto `Object`. That will get called before `method_missing`.

lib/rspec/rails/extensions/active_support/json.rb
cupakromer: I agree. I thought about that and almost created a method alias, but dropped it since it was a single conditional check in the original. I can go back to the alias method instead of `super`.

lib/rspec/rails/extensions/active_support/json.rb
cupakromer: Ah right, as we did in the other PR. Sounds good to me.

myronmarston:
There's a fair bit of duplication for `as_json` vs `to_json` here.  Also, as you noted, this is coupled to rspec-mocks internals in a way that I'm uncomfortable with.

I think that we can solve both issues by defining some kind of public rspec-mocks API that would allow rspec-rails to add these methods to test doubles.  Maybe something like this:

``` ruby
RSpec::Mocks.configure do |c|
  c.add_method_to_test_doubles(:as_json) { nil }
  c.add_method_to_test_doubles(:to_json) { "null" }
end
```

Thoughts?

cupakromer:
> I think that we can solve both issues by defining some kind of public rspec-mocks API that would allow rspec-rails to add these methods to test doubles.

That would be nice. There's a lot of corner cases around `respond_to?` and the call chain hierarchy that needs to get covered. If the mocks helpers could be sure to cover those that would be really great.

JonRowe:
> There's a fair bit of duplication for `as_json` vs `to_json` here.  Also, as you noted, this is coupled to rspec-mocks internals in a way that I'm uncomfortable with.

> I think that we can solve both issues by defining some kind of public rspec-mocks API that would allow rspec-rails to add these methods to test doubles.  Maybe something like this:

> ``` ruby
RSpec::Mocks.configure do |c|
  c.add_method_to_test_doubles(:as_json) { nil }
  c.add_method_to_test_doubles(:to_json) { "null" }
end
```

> Thoughts?

:+1:

**Comment by samphippen** [comment]
@cupakromer @JonRowe is this still valid?

**Comment by JonRowe** [comment]
Yes, but the build is failing, maybe rebase it and see if you can get it to run?

File: spec/rspec/rails/extensions/active_support/json_spec.rb [hunk 4]
```diff
@@ -0,0 +1,196 @@
...
+        basic_object = Class.new do
+          undef :as_json
+          undef :to_json
+        end
+        stub_const("BasicObject", basic_object)
+      end
+    else
+      # Sanity check to ensure Rails didn't change this on us
+      before(:context) do
+        expect(BasicObject).not_to respond_to(:as_json)
```
samphippen: https://travis-ci.org/rspec/rspec-rails/jobs/120927983#L543

Unless I'm mistaken, BasicObject will never exist here because we defined checked it above. Do we know why the undefined constant is happening here?

cupakromer: Hmm, I think this was a combination of attempting to use scope to determine if `BasicObject` was from Ruby or [patched in by Rails](https://github.com/rails/rails/blob/v3.2.22.2/activesupport/lib/active_support/basic_object.rb).

I probably was thinking that at file load time Rails had not patched in `BasicObject`, for older Rubies, but by the time the example group was run Rails's patch was loaded. So this was left in as a sanity check in-case Rails ever changed the behavior and my precondition assumptions later became invalid (per the comment).

cupakromer: So to answer the question: No. I'm not sure yet why the undefined constant is happening.

cupakromer: Looks like I wasn't paying attention to the Rails patch scoping. It's defining `ActiveSupport::BasicObject` - not `::BasicObject`. Which is likely why the constant isn't defined.

aselder:
Hi all? Is there anything I can do to help this along? We're running into this issue and would love to help get it solved.

samphippen:
@aselder this PR just needs to be pushed to completion. All the RSpec people are pretty busy so if you're able to take this branch, and get it to pass, then I'll gladly review your work.

PR #1315, opened 2015-02-20
Remove need to get fixture_path from ActiveSupport::TestCase

seanknox:
No longer needed since https://github.com/rails/rails/pull/1861.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -10,17 +10,6 @@ module RSpec
         include ActiveRecord::TestFixtures
 
         included do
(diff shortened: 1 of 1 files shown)


cupakromer:
I think this is still required for us. We still support Rails 3.0.x and that does not have the fix: https://github.com/rails/rails/blob/3-0-stable/actionpack/lib/action_dispatch/testing/test_process.rb#L41

Please let me know if I'm missing something else.

seanknox:
Ah, that pesky yet kind act of supporting older versions. Totally makes sense. Going to drop this for now.

PR #1318, opened 2015-02-21
Relax `rails` dependency constraint to support the `4.2.x` releases.

lucasmazza:
The previous constraint would only go as up as `4.2.0`, thus making `rspec-rails`
incompatible with any of the patch releases of the rails 4.2.x series.

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -27,9 +27,9 @@ Gem::Specification.new do |s|
     s.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]
   end
 
-  s.add_runtime_dependency(%q<activesupport>, [">= 3.0", "<= 4.2"])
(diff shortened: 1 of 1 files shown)


lucasmazza:
This change is required to use `rspec-rails` with the 4.2.1 rc released today (http://weblog.rubyonrails.org/2015/2/20/Rails-4-2-1-rc1-and-4-1-10-rc1-have-been-released/), would be awesome to have a new release of the `3.2.x` series of rspec-rails.

cupakromer:
:heart: Thanks!!!

lucasmazza:
@cupakromer thanks for the merge! Any plans on releasing a new version (either something like `3.2.1` or the `3.3.x`) with this?

cupakromer:
@lucasmazza yes, we plan on getting a 3.2.1 patch release out in the next day or two.

PR #1320, opened 2015-02-21
Disable the test-unit gem autorunner.

### LINKED ISSUE #1312: invalid option: --format
```bash
$ rspec spec --format documentation
... test results
invalid option: --format
Usage: ...
```
Why rspec warn me about this option? Test results printed exactly in this format "documentation", but with this warning and usage.
```bash
 $ bundle show | grep rspec
  * rspec-core (3.2.0)
  * rspec-expectations (3.2.0)
  * rspec-mocks (3.2.0)
  * rspec-rails (3.2.0)
  * rspec-support (3.2.1)
```
This option is in the official documentation. https://relishapp.com/rspec/rspec-core/v/3-2/docs/command-line/format-option

cupakromer:
For some reason the test-unit gem loads it's autorunner by default.
Though the test-unit included with Ruby does not. The original code for
loading test-unit pre-dates widespread usage of the test-unit gem.

We took care of disabling the autorunner for Ruby 2.2+ since the
test-unit gem is now required for Rails 3.2. For older Ruby versions we
attempt to only load the `test/unit/assertions` module. When using the
`test-unit` gem however, Rails will auto require it for us, which loads
the main test-unit code, thus enabling the autorunner.

Resolve #1312

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,8 @@ Bug Fixes:
   autoload does not find it in some environments. (Aaron Kromer, #1305)
 * `be_routable` matcher now has the correct description. (Tony Ta, #1310)
 * Fix dependency to allow Rails 4.2.x patches / pre-releases (Lucas Mazza, #1318)
+* Disable the `test-unit` gem's autorunner on projects running Rails < 4.1 and
+  Ruby < 2.2 (Aaron Kromer, #1320)
 
 ### 3.2.0 / 2015-02-03
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.1.0...v3.2.0)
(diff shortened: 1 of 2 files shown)


myronmarston:
LGTM.  A spec covering this would be nice or maybe a travis build matrix entry we currently lack?) but if you've looked into it and concluded the effort of such a thing would exceed the benefit that's fine.

cupakromer:
I'm hesitant to expand the build matrix even further just for this. I'm not sure of a way to conditionally include the gem without overriding the Ruby implementation. I'm also not sure how common it is to have the `test-unit` gem loaded with older Ruby + Rails versions.

myronmarston:
Makes sense to me.  Merge away.

PR #1324, opened 2015-02-25
Fix permissions.

voxik:
These files does not look executable.


cupakromer:
Why do these need to be executable?

cupakromer:
Oh sorry. I miss read that. Thanks! :-D :heart:

Issue #1325, opened 2015-02-27
Action Mailer preview_path warning


kamen-hursev:
Is it really necessary to output a warning: 'Action Mailer `preview_path` is not the RSpec default.' when the preview_path is changed to something different than its default value?

cupakromer:
This is being removed, is part of a PR that will be submitted later today

kamen-hursev:
Great, thanks! :smile:

cupakromer:
Closing per #1327

PR #1327, opened 2015-03-02
Fix action mailer preview path setup.

### LINKED ISSUE #1313: 3.1.0 -> 3.2.0 breaks url generation
When I updated from rspec-rails 3.1.0 to 3.2.0, url generation (as in, `url_for`) broke. I get errors like:
```
ActionView::Template::Error: Missing host to link to! Please provide the :host parameter, set default_url_options[:host], or set :only_path to true
from /[...]/[...]/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/actionpack-4.2.0/lib/action_dispatch/http/url.rb:45:in `full_url_for'
```

I am not setting default_url_options. I could do so, but the point is that the code works fine with the previous version of rspec-rails, and it breaks with the new one.

cupakromer:
This removes the unnecessary warning about a preview path set to a
location other than `/spec/mailers/preview`.

This does not force `ActionMailer` to load when the Railtie runs by
removing the `defined?` check. This is necessary because the
`ActionMailer` class only loads it's settings once, after it is first
loaded. If we load the class in our Railtie any custom settings declared
in a config block in an initializer will be ignored.

An improved spec suite for the different permutations of availability of
action mailer previews is included as well.

Resolve #1313, #1325

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,7 @@ Bug Fixes:
 
 * Fix auto-including of generic `Helper` object for view specs sitting in the
   `app/views` root (David Daniell, #1289)
+* Remove pre-loading of ActionMailer in the Railtie (Aaron Kromer, #1327)
 
 ### 3.2.1 / 2015-02-23
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.2.0...v3.2.1)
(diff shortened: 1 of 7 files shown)


olance:
Hi,

rspec-rails 3.2.2 is broken with Rails 3.2.21 (and probably Rails 3.x) as the test to configure `preview_path` default value seems to have been overly simplified.
`preview_path` is indeed not available in Rails 3.x (not even before Rails 4.1.8), however it will be added to ActionMailer configuration if Rails' version is less than '4.2' and in development mode.

I believe this PR is responsible for this break?

tagliala:
same here

PR #1336, opened 2015-03-08
Add PATCH HTTP method to scaffold routing generator

biow0lf:
Add PATCH HTTP method to scaffold routing generator.


JonRowe:
Thanks @biow0lf

PR #1350, opened 2015-04-09
Check `need_auto_run=` method is defined also

orien:
To prevent errors such as:

```
undefined method `need_auto_run=' for Test::Unit::AutoRunner:Class (NoMethodError)
/Users/orien/src/marketplace/.bundle/ruby/2.1.0/gems/rspec-rails-3.2.1/lib/rspec/rails/adapters.rb:65:in `<module:Rails>'
/Users/orien/src/marketplace/.bundle/ruby/2.1.0/gems/rspec-rails-3.2.1/lib/rspec/rails/adapters.rb:7:in `<module:RSpec>'
/Users/orien/src/marketplace/.bundle/ruby/2.1.0/gems/rspec-rails-3.2.1/lib/rspec/rails/adapters.rb:6:in `<top (required)>'
```

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -37,7 +37,7 @@ module RSpec
           # date). If so, we turn the auto runner off.
           require 'test/unit'
           require 'test/unit/assertions'
-          Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner)
+          Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner.need_auto_run = ())
         rescue LoadError => e
           raise LoadError, <<-ERR.squish, e.backtrace
             Ruby 2.2+ has removed test/unit from the core library. Rails
@@ -62,7 +62,7 @@ module RSpec
         require 'test/unit/assertions'
       end
       # Turn off test unit's auto runner for those using the gem
-      Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner)
+      Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner.need_auto_run = ())
       # Constant aliased to either Minitest or TestUnit, depending on what is
       # loaded.
       Assertions = Test::Unit::Assertions
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -5,6 +5,19 @@ require 'active_support/core_ext/string'
 
 module RSpec
   module Rails
+    def self.disable_testunit_autorun
+      # `Test::Unit::AutoRunner.need_auto_run=` was introduced to the test-unit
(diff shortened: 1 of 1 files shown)


lib/rspec/rails/adapters.rb
```diff
@@ -37,7 +37,7 @@ module Rails
           # date). If so, we turn the auto runner off.
           require 'test/unit'
           require 'test/unit/assertions'
-          Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner)
+          Test::Unit::AutoRunner.need_auto_run = false if defined?(Test::Unit::AutoRunner.need_auto_run = ())
```
JonRowe: Huh... TIL... I would have expected this to blow up.

JonRowe:
So why wouldn't this exist if `Test::Unit::AutoRunner` exists, if the method has been renamed across versions we should attempt to pick its replacement (if one exists) instead?

orien:
[This version](https://github.com/ruby/ruby/blob/ruby_2_1/lib/test/unit.rb#L807-L846) from the Ruby codebase doesn't have that method. It has no alternative that I can see. Compare to the test-unit gem [version](https://github.com/test-unit/test-unit/blob/3.0.9/lib/test/unit/autorunner.rb).

cupakromer:
So do we just need this check on line 65? As the first one _should_ be guaranteed to use the gem?

orien:
@cupakromer Actually, that's the line I experienced the error.
FYI: in an app running Ruby v2.1.5 and Rails v3.2.21.

cupakromer:
@orien Good question. I think it was being defensive for changes in the future. In which case, your check is actually better. It seems the method was introduced sometime between the test-unit 2.4.0 and 2.4.9 gems. :+1:

cupakromer:
> we should attempt to pick its replacement (if one exists) instead?

@JonRowe I didn't see a clean replacement for the Ruby version. The gem version seems like it has an alternative in: [`Test::Unit.run=`](http://www.rubydoc.info/gems/test-unit/2.4.0/Test/Unit#run%3D-class_method). We would need to set it to `true`. However, several of the gems with the newer version also have this method, but have simply deprecated it. So we would need to check for it after the newer method.

lib/rspec/rails/adapters.rb
orien: `defined?` is quite magical!

cupakromer:
@orien if you want to add the 2nd fallback for `run=` I think this will be good. Given we now have an additional check I think we should move the logic into a helper. Something like `disable_testunit_autorun!`. We can then call it from both sections of the code and let it handle all the checks.

orien:
@cupakromer I used a non bang version of that method name. As per http://dablog.rubypal.com/2007/8/15/bang-methods-or-danger-will-rubyist

JonRowe:
This looks good to me :)

cupakromer:
Thanks for this @orien :blue_heart:

PR #1354, opened 2015-04-15
Provide a summary for humans

jgorset:
"rspec-rails-3" doesn't really add anything I don't already know (and especially if you use something like [label](https://github.com/jgorset/label) it looks pretty silly) — could we change it to something like "RSpec for Rails" and use the description for something even more descriptive? :cake:


JonRowe:
Don't see why not

JonRowe:
Thanks

jgorset:
No, thank you!

PR #1361, opened 2015-05-05
Add JobExampleGroup to support :type => :job

gabetax:
`:type => :job` metadata is already being written by the [job_spec template](https://github.com/rspec/rspec-rails/blob/a09a6231ceecefa177ec08b27c3066d5947e5899/spec/generators/rspec/job/job_generator_spec.rb#L6). If it's already being defined, then `infer_spec_type_from_file_location!` should support it too.


JonRowe:
We should probably do this conditionally like we do with mailers, in case we ever extend this to do something more specific...

gabetax:
The `if defined?(ActionMailer)` check is only when modules are mixed in. I don't think a conditional is necessary until behavior is introduced within rspec-rails itself.

JonRowe:
Actually I would consider it surprising behaviour for `:type => :job` to do anything when there is no active job so I think it is necessary.

cupakromer:
Thanks for submitting this @gabetax :blue_heart: I'm with Jon on this, we should have the check to ensure active job is available as we do for action mailer.

lib/rspec/rails/configuration.rb
```diff
@@ -42,6 +43,7 @@ def self.initialize_configuration(config)
 
       config.include RSpec::Rails::ControllerExampleGroup, :type => :controller
       config.include RSpec::Rails::HelperExampleGroup,     :type => :helper
+      config.include RSpec::Rails::JobExampleGroup,        :type => :job
```
JonRowe: I'd rather this was conditional...

gabetax:
Thanks for the fast feedback. Is this the parity you were looking for?

JonRowe:
Yeah that'll do, thanks.

PR #1372, opened 2015-05-19
Update fixture support

### LINKED ISSUE #1355: Fixtures error when tests requiring 'spec_helper' are run alongside tests requiring 'rails_helper'
A failure occurs when some tests require only a very barebones `spec_helper` for classes that have no dependency on rails, while other tests require `rails_helper` if they have dependencies on the rails side of the app.

When run individually, both types of tests run fine.  When run together, the `spec_helper` tests will raise an error.  It does not appear to matter which order the tests run in.

```
Failure/Error: Unable to find matching line from backtrace
     NoMethodError:
       undefined method `pre_loaded_fixtures' for SomeClass
     # ruby-2.1.5@avant_basic/gems/attr_encryptor-2.0.0/lib/attr_encryptor.rb:263:in `method_missing'
     # ruby-2.1.5@avant_basic/gems/activesupport-3.2.19/lib/active_support/core_ext/class/attribute.rb:98:in `pre_loaded_fixtures'
     # ruby-2.1.5@avant_basic/gems/activerecord-3.2.19/lib/active_record/fixtures.rb:836:in `setup_fixtures'
```

I've traced the issue to this line: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/fixture_support.rb#L10

activesupport (3.2.19)
rails (3.2.19)
rspec-core (3.2.1)
rspec-expectations (3.2.0)
rspec-mocks (3.2.1)
rspec-rails (3.2.1)
rspec-support (3.2.2)

cupakromer:
This fixes an "undefined method `fixture_path`" error caused in
`RSpec::Core::Configuration`. The issue occurs when a spec is created
before rspec-rails is loaded. The root cause is rspec-core attempts to
include the `RSpec::Rails::FixtureSupport` module into the existing spec
as soon as we tell the configuration about it, but have not told the
configuration about the new related options.

This moves the including of the module until after we have told the
configuration about the dependent options.

In order to test this, a custom spec and script are included which
attempt to ensure it is the first spec loaded by rspec-core. To ensure
we test the issue the run script is updated to also run this spec file
standalone.

This also includes full spec coverage for all rspec-rails configuration
options. These specs are written in an integration style; instead of a
more unit-collaborator approach. The reason is that users of rspec-rails
have specific expectations on how the configuration options are
available. If rspec-core changes anything related to these calls which
would break the rspec-rails user's expectations, these specs will catch
it.

This also adds metadata to ad-hoc include fixture support for any
examples or groups which may require it. It additionally, includes
fixture support into all default `RailsExampleGroup` modules. This
provides the current functionality for the majority of use cases.

Fix #1355

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -65,10 +65,10 @@ module RSpec
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
       config.add_setting :fixture_path
+      config.include RSpec::Rails::FixtureSupport, :use_fixtures
 
-      # TODO: We'll need to create a deprecated module in order to properly
-      #  report to gems / projects which are relying on this being loaded
-      #  globally.
+      # We'll need to create a deprecated module in order to properly report to
+      # gems / projects which are relying on this being loaded globally.
       #
       # See rspec/rspec-rails#1355 for history
       #
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -61,11 +61,20 @@ module RSpec
       config.add_setting :infer_base_class_for_anonymous_controllers, :default => true
 
       # fixture support
-      config.include RSpec::Rails::FixtureSupport
       config.add_setting :use_transactional_fixtures, :alias_with => :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
       config.add_setting :fixture_path
+
+      # TODO: We'll need to create a deprecated module in order to properly
+      #  report to gems / projects which are relying on this being loaded
+      #  globally.
+      #
+      # See rspec/rspec-rails#1355 for history
+      #
+      # @deprecated Include `RSpec::Rails::RailsExampleGroup` or
+      #   `RSpec::Rails::FixtureSupport` directly instead
+      config.include RSpec::Rails::FixtureSupport
 
       # This allows us to expose `render_views` as a config option even though it
       # breaks the convention of other options by using `render_views` as a
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
---color
 --warnings
+--color
+--require spec_helper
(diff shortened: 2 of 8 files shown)


JonRowe:
Looks sane to me

lib/rspec/rails/configuration.rb
```diff
@@ -56,11 +56,20 @@ def self.initialize_configuration(config)
...
+      config.include RSpec::Rails::FixtureSupport, :use_fixtures
+
+      # We'll need to create a deprecated module in order to properly report to
+      # gems / projects which are relying on this being loaded globally.
+      #
+      # See rspec/rspec-rails#1355 for history
+      #
+      # @deprecated Include `RSpec::Rails::RailsExampleGroup` or
+      #   `RSpec::Rails::FixtureSupport` directly instead
+      config.include RSpec::Rails::FixtureSupport
```
myronmarston: What does it mean to tag a line of code as deprecated rather than a method or class?  Will users see this deprecation in the YARD docs?  And why are you including the module above and here?

cupakromer: It was mostly a note for myself and others for 3.99. Line 72 is us globally patching fixture support onto all specs. Due to semver we can't just remove it. So this was a reminder that we need to warn users in 3.99.

I've added fixture support to the `RSpec::Rails:RailsExampleGroup` base module. So all Rails related specs which people expect fixtures to be available for will have it. Line 63 above provides an option for people to selectively include fixture support onto any example/group via metadata.

mnoack:
How do I globally disable RSpec::Rails::FixtureSupport?
It interferes with nulldb gem which can't insert anything into database. This wasnt a problem prior to rails 5 because it wasn't being included and MinitestLifecycle wasn't being included

samphippen:
@mnoack if this is presenting as an actual issue for you that broke compatibility, please open a new issue. We'd prefer it if you not comment on old issues. 

If you do open your issue, can you please provide clear steps to reproduce, following this guide:


> Thanks for the issue. We need a little more detail to be able to reproduce this.
> 
> Could you please provide us with a rails app that we can clone that demonstrates the issue. Specifically it'd be great if
> 
> 1)  you could `rails new` an application and commit
> 2) make all the changes necessary to reproduce the issue and commit
> 
> then, provide us with a description of how to clone your application and reproduce the issue.
> 
> Thanks :)

PR #1373, opened 2015-05-19
Update License.txt

bogdanvlviv:


myronmarston:
Thanks for pointing this out!

2012 is the year that @alindeman became lead maintainer, which is why that's the year listed there.  We should probably add a line for when @cupakromer took over (2014, I think?) -- e.g. `Copyright (c) 2015 David Chelimsky, Andy Lindeman, Aaron Kromer`.

I'm not a lawyer, but I don't think there's any need to update or add a copyright line with the current year in there.

Regardless, I'll leave @cupakromer to update it as he sees fit.

PR #1374, opened 2015-05-24
Switching to the real SEMVER format

krainboltgreene:
SEMVER describes "pre-release" information as being appended to the version by a dash.

### DIFF
--- a/lib/rspec/rails/version.rb
+++ b/lib/rspec/rails/version.rb
@@ -3,7 +3,7 @@ module RSpec
     # Version information for RSpec Rails.
     module Version
       # Current version of RSpec Rails, in semantic versioning format.
-      STRING = '3.3.0.pre'
+      STRING = '3.3.0-pre'
     end
   end
 end


JonRowe:
This would be inconsistent with how we've been releasing RSpec versions in the past, not sure wether this is important enough to change... Thoughts @myronmarston?

JonRowe:
We'd also need to update the build scripts so they look for the right prerelease versions of other gems.

cupakromer:
I have no issues making the change. Though I'd prefer to remain consistent with the rest of the rspec gems, so we'd need to consider making this change across the board.

myronmarston:
Thanks for the pointing this out @krainboltgreene!

When I was getting ready to cut some pre-releases for VCR 2.0 (the first beta, IIRC), I tried following the SemVer.org spec for prerelease versions and found that rubygems had some problems with using a dash to separate the normal version part from the prerelease part.  It looks like whatever issue I ran into has been fixed (building a gem with a `3.3.0-pre` does not error now like it used to).  However, rubygems still seems to treat it a bit oddly -- notice how it parses `3.3.0.pre` vs `3.3.0-pre`:

```
➜  rspec-expectations git:(try-diff-version-form) ✗ irb -rubygems
irb(main):001:0> Gem::Version.new("3.3.0.pre")
=> #<Gem::Version "3.3.0.pre">
irb(main):002:0> Gem::Version.new("3.3.0-pre")
=> #<Gem::Version "3.3.0.pre.pre">
```

If we use `-pre` instead of `.pre`, it seems to parse it as `.pre.pre` for some reason.  Reading the [docs](https://github.com/rubygems/rubygems/blob/v2.4.7/lib/rubygems/version.rb#L2-L23):

```
# The Version class processes string versions into comparable
# values. A version string should normally be a series of numbers
# separated by periods. Each part (digits separated by periods) is
# considered its own number, and these are used for sorting. So for
# instance, 3.10 sorts higher than 3.2 because ten is greater than
# two.
#
# If any part contains letters (currently only a-z are supported) then
# that version is considered prerelease. Versions with a prerelease
# part in the Nth part sort less than versions with N-1
# parts. Prerelease parts are sorted alphabetically using the normal
# Ruby string sorting rules. If a prerelease part contains both
# letters and numbers, it will be broken into multiple parts to
# provide expected sort behavior (1.0.a10 becomes 1.0.a.10, and is
# greater than 1.0.a9).
#
# Prereleases sort between real releases (newest to oldest):
#
# 1. 1.0
# 2. 1.0.b1
# 3. 1.0.a.2
# 4. 0.9
```

...it seems to suggest that rubygems only intends to handle versions with period separating the parts, and all the pre-release versions use periods instead of dashes.

I'd be interested in switching to follow the semver.org spec once rubygems gets these issues ironed out, but until then sticking with a period is going to work better.

krainboltgreene:
Ugh, I had hoped they had fixed this by now. I reported it to them ages ago.

On Tue, May 26, 2015 at 2:10 AM, Myron Marston <notifications@github.com>
wrote:

> Closed #1374 <https://github.com/rspec/rspec-rails/pull/1374>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rspec/rspec-rails/pull/1374#event-313478063>.
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

PR #1375, opened 2015-05-24
consistent quoting, more readable interpolation

josephgrossberg:
via http://refactorcop.com/rspec/rspec-rails


JonRowe:
Thanks.

PR #1378, opened 2015-05-25
Add request generator

### LINKED ISSUE #1377: rspec:request generator not shown in "bin/rails g --help"
The rspec:request generator is not being shown in the output of `bin/rails g --help`. Any pointers on what might cause this?

cupakromer:
This includes a generator for "request" specs. There already exists a
generator which creates these specs, but it is called "integration".
The existing name comes from the legacy implementations attempting to
mirror the Rails test style names for the RSpec equivalent specs.

Since these two generators are identical, except for the name, this
makes the new `RequestGenerator` a subclass of the existing
`IntegrationGenerator`. Alternatively, we could have decided to make
`RequestGenerator` the parent class, moving the existing functionality
into it. While this is arguably the "proper" way of thinking about the
generators, it could potentially break some existing applications; even
though these classes are marked `private`. There's no harm for now in
keeping semver compliant by making this generator additive.

Since the two generators are exactly the same this uses a shared example
to keep the specs in-sync.

Resolve #1377


JonRowe:
LGTM :)

PR #1379, opened 2015-05-25
Address yard issues

### LINKED ISSUE #1297: Address YARD warnings and get docs back to 100%
```
[warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
[warn]:         in file 'lib/rspec/rails/configuration.rb':73:

        73: def config.render_views=(val)

[warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
[warn]:         in file 'lib/rspec/rails/configuration.rb':77:

        77: def config.render_views

[warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
[warn]:         in file 'lib/rspec/rails/configuration.rb':81:

        81: def config.render_views?

[warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
[warn]:         in file 'lib/rspec/rails/configuration.rb':85:

        85: def config.infer_spec_type_from_file_location!

[warn]: Syntax error in `lib/generators/rspec/mailer/templates/preview.rb`:(1,1): syntax error, unexpected '<'
[warn]: in YARD::Handlers::Ruby::MixinHandler: Undocumentable mixin: YARD::Parser::UndocumentableError for class RSpec::Rails::RoutingExampleGroup
[warn]:         in file 'lib/rspec/rails/example/routing_example_group.rb':11:

        11: include RSpec::Rails::AssertionDelegator.new(ActionDispatch::Assertions::RoutingAssertions)

[warn]: in YARD::Handlers::Ruby::MixinHandler: Undocumentable mixin: YARD::Parser::UndocumentableError for class RSpec::Rails::ControllerExampleGroup
[warn]:         in file 'lib/rspec/rails/example/controller_example_group.rb':12:

        12: include RSpec::Rails::AssertionDelegator.new(ActionDispatch::Assertions::RoutingAssertions)

[warn]: Syntax error in `lib/generators/rspec/install/templates/spec/rails_helper.rb`:(23,1): syntax error, unexpected '<', expecting end-of-input
Files:          26
Modules:        27 (    0 undocumented)
Classes:         7 (    0 undocumented)
Constants:       4 (    0 undocumented)
Methods:        59 (    1 undocumented)
 98.97% documented
```

cupakromer:
This makes YARD available in the dev/test environment again, fixing the `bin/yard` failure. It also updates the docs to get minimal 100% coverage and addresses the outstanding warnings.

Fix #1297

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -84,23 +84,25 @@ module RSpec
       # but requires this workaround:
       config.add_setting :rendering_views, :default => false
 
-      def config.render_views=(val)
-        self.rendering_views = val
-      end
+      config.instance_exec do
+        def render_views=(val)
+          self.rendering_views = val
+        end
 
-      def config.render_views
-        self.rendering_views = true
-      end
+        def render_views
+          self.rendering_views = true
+        end
 
-      def config.render_views?
-        rendering_views
-      end
+        def render_views?
+          rendering_views
+        end
 
-      def config.infer_spec_type_from_file_location!
-        DIRECTORY_MAPPINGS.each do |type, dir_parts|
-          escaped_path = Regexp.compile(dir_parts.join('[\\\/]') + '[\\\/]')
-          define_derived_metadata(:file_path => escaped_path) do |metadata|
-            metadata[:type] ||= type
+        def infer_spec_type_from_file_location!
+          DIRECTORY_MAPPINGS.each do |type, dir_parts|
+            escaped_path = Regexp.compile(dir_parts.join('[\\\/]') + '[\\\/]')
+            define_derived_metadata(:file_path => escaped_path) do |metadata|
+              metadata[:type] ||= type
+            end
           end
         end
       end
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -6,6 +6,8 @@ module RSpec
     module ViewRendering
       extend ActiveSupport::Concern
 
+      # @!attribute [r]
+      # Returns the controller object instance under test.
       attr_accessor :controller
 
       # DSL methods
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -8,7 +8,11 @@ module RSpec
 
       # @!attribute [r]
       # Returns the controller object instance under test.
-      attr_accessor :controller
+      attr_reader :controller
+
+      # @private
+      attr_writer :controller
+      private :controller=
 
       # DSL methods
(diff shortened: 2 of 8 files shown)


lib/rspec/rails/configuration.rb
```diff
@@ -84,23 +84,25 @@ def self.initialize_configuration(config)
       # but requires this workaround:
       config.add_setting :rendering_views, :default => false
 
-      def config.render_views=(val)
-        self.rendering_views = val
-      end
+      config.instance_exec do
```
JonRowe: Is `instance_exec` really necessary to fix up the yard?

cupakromer:
YARD issues warnings about writing the methods directly on the object:

> [warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
> [warn]:         in file 'lib/rspec/rails/configuration.rb':73:
>
>         73: def config.render_views=(val)

Using `instance_exec` provides access to the same singleton scope allowing us to write the methods. Another option would be place all of those methods into a module and then access the singleton class directly (providing proper 1.8.7 backwards compat) to `extend` aforementioned module.

lib/rspec/rails/configuration.rb
cupakromer: YARD issues warnings about writing the methods directly on the object:

> [warn]: in YARD::Handlers::Ruby::MethodHandler: Undocumentable method defined on object instance
> [warn]:         in file 'lib/rspec/rails/configuration.rb':73:
>
>         73: def config.render_views=(val)

Using `instance_exec` provides access to the same singleton scope allowing us to write the methods. Another option would be place all of those methods into a module and then access the singleton class directly (providing proper 1.8.7 backwards compat) to `extend` aforementioned module.

lib/rspec/rails/view_rendering.rb
```diff
@@ -6,6 +6,8 @@ module Rails
     module ViewRendering
       extend ActiveSupport::Concern
 
+      # @!attribute [r]
```
myronmarston: I assume this is just here to declare the attribute is read-only, right?  Should we consider changing `attr_accessor` to `attr_reader` to actually make it read only?  And/or we could do `private :controller=` to make the writer private.

myronmarston:
LGTM

lib/rspec/rails/view_rendering.rb
cupakromer: Oops, that was a copy-paste mistake. But yes, I believe you are correct, we should note that `controller=` is a `private` API call. Thanks for catching it!

PR #1381, opened 2015-05-27
allow users to use fixture_path and transactional_fixtures settings

swaroopmurthy:
We have a project where we're using 
  config.fixture_path
  config.use_transactional_fixtures

Why was this removed?

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -9,6 +9,15 @@ module RSpec
         include RSpec::Rails::MinitestAssertionAdapter
         include ActiveRecord::TestFixtures
 
+        RSpec.configure do |c|
+          c.include RSpec::Rails::FixtureSupport
+          c.add_setting :use_transactional_fixtures
+          c.add_setting :use_transactional_examples, :alias => :use_transactional_fixtures
+          c.add_setting :use_instantiated_fixtures
+          c.add_setting :global_fixtures
+          c.add_setting :fixture_path
+        end
+
         included do
           # TODO: (DC 2011-06-25) this is necessary because fixture_file_upload
           # accesses fixture_path directly on ActiveSupport::TestCase. This is


c0mrade:
Not sure when or why at this point.

As stated in our conversation I think we should get rid of the fixtures, simply because the Factorygirl provides the same feature. Plus tons more, it's not a just dumb stub as fixture is, it can create a real records or list of them.

Swaroop opened a chore for this he can link, it's bit outside of the upgrade to rspec 3 to remove the fixtures.

If there is any opinions to keep the fixtures, please present them, we can keep the fixtures and loose the factorygirl if they're strong enough, we really don't need both.

cupakromer:
Was this meant to be added to `rspec/rspec-rails`? Those configuration options have not been removed. They are set in [`rspec/rails/configuration.rb`](https://github.com/rspec/rspec-rails/blob/03d8cbe21c/lib/rspec/rails/configuration.rb#L63-L67).

swaroopmurthy:
Apologies, we must be doing it in the wrong order. I'm getting this error:
uninitialized constant RSpec::Rails::ControllerExampleGroup

from the following sequence:
require 'rspec/rails/extensions/active_record/proxy'
require 'rspec/rails/adapters'
require 'rspec/rails/configuration'
require 'rspec/rails/fixture_support'

cupakromer:
You should just be using `require 'rspec/rails'` after loading your Rails configuration. The [default generated `spec/rails_helper.rb`](https://github.com/rspec/rspec-rails/blob/b913f6acfc2c4b7/lib/generators/rspec/install/templates/spec/rails_helper.rb) starts with:

```ruby
ENV['RAILS_ENV'] ||= 'test'
# Prevent database truncation if the environment is production
abort("The Rails environment is running in production mode!") if Rails.env.production?
require 'spec_helper'
require 'rspec/rails'
```

cupakromer:
Is this by any chance the issue reported in #1355 fixed in #1372?

cupakromer:
@swaroopmurthy please let us know if you're still seeing this issue. If so please make sure you've updated to the lasted to get the prior mentioned fix. Cheers!

Issue #1273, opened 2015-01-13
ruby 2.2.0 rails 4.0.13 and rspec 3.1.0 cannot load such file -- test/unit/assertions 


waynehoover:
Hello, I'm getting this error, any ideas why? `require': cannot load such file -- test/unit/assertions 

```
~/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/activesupport-4.0.13/lib/active_support/dependencies.rb:229:in `require': cannot load such file -- test/unit/assertions (LoadError)
```

The stack trace points to this line in my spec_helper:`require 'rspec/rails'`

waynehoover:
I figured it out.

You have to add the test-unit gem to your gemfile: `gem 'test-unit'`

cupakromer:
This is related to #1264. It should be fixed in the next release. You should just need the `minitest` gem, it sets up the proper test unit shims.

rohandaxini:
I had to add both, `test-unit` and `minitest' gems to make this work. So both are fine or am I doing anything wrong here?

cupakromer:
@rohandaxini you should not need the `test-unit` gem unless you have something else that is loading test unit specifically. The minitest gem handles setting up the proper constant aliases. What problem did you run into?

rohandaxini:
Thanks for your note @cupakromer
Actually I was facing this error while running my tests (rspec test suite) on local. 

/Users/Rohan/.rvm/gems/ruby-2.2.0/gems/activesupport-3.2.18/lib/active_support/dependencies.rb:251:in `require': cannot load such file -- test/unit/testcase (LoadError)

This has happened recently when I upgraded to Ruby2.2.0, earlier I was not facing this error on Ruby 2.1.2 or 2.1.5

I found this git issue and thought to add `test-unit` after other error went away by adding `minitest`
But I am sure, I am doing something wrong that has led me to add both these gems. Thoughts ?

btw, while upgrading to Ruby 2.2.0, I have no other code change except in Gemfile. My new entires in Gemfile are:
ruby '2.2.0'
gem 'eventmachine', '1.0.4'

Accordingly my Gemfile.lock has changes only related to these 2 gems.

cupakromer:
Thanks for the background information. Have you tried just removing `test-unit` from your `Gemfile` and only having `minitest`?

cupakromer:
[...]

PR #1388, opened 2015-06-04
Fix regression mailer preview path

### LINKED ISSUE #1386: Rails 3.2.1 does not start with rspec-rails 3.2.2
It should be a consequence of #1327, as per https://github.com/rspec/rspec-rails/pull/1327#issuecomment-108278230

```sh
$ rails c
~/.rvm/gems/ruby-2.1.5/gems/actionmailer-3.2.21/lib/action_mailer/base.rb:438:in `method_missing': undefined method `preview_path=' for ActionMailer::Base:Class (NoMethodError)
```

cupakromer:
This fixes the regression where setting the `ActionMailer` preview path leaked into versions which does not support it (Rails before 4.1).

The cause of the regression was a series of failures:

- The integration spec did not force `ActionMailer::Base` to load; thus the incorrect setting was not loaded exposing the issue
- In the case of older Rails versions, it was believed that the mailer spec's script defaulting to the "development" environment would be sufficient. However, when shelling out the existing Ruby `ENV` was already available, resulting in the `RAILS_ENV` already being set to "test". Thus the test was not explicitly testing the code as expected.
- The implementation code itself made a note about `config.action_mailer.show_previews` always returning `true`, yet code slightly further down was relying on `config.action_mailer.preview_path` being able to return `false` (which it never will)

This fixes each of these issues.

Based on minimal local testing it appears that `defined?(ActionMailer)`, while possibly loading `ActionMailer`, will _not_ force the configuration to load. The configuration is only with `ActionMailer::Base`.

I'm weary :weary: about switching out the logic yet again regarding: `defined?(ActionMailer)` vs `config.respond_to?(:action_mailer)`. It's possible that on any further version the behavior of the config being loaded or the config changing to always respond `true` can occur. For now, it seems that `config.respond_to?(:action_mailer)` is working and stable. Though I'm always open to further discussion.

Resolve #1386

/cc @tagliala, @olance, @myronmarston

### DIFF
--- a/example_app_generator/generate_action_mailer_specs.rb
+++ b/example_app_generator/generate_action_mailer_specs.rb
@@ -14,11 +14,25 @@ using_source_path(File.expand_path('..', __FILE__)) do
 
   initializer 'action_mailer.rb', <<-CODE
     if ENV['DEFAULT_URL']
-      Rails.application.configure do
-        config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+      if ::Rails::VERSION::STRING < '4.1'
+        ExampleApp::Application.configure do
+          config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+        end
+      else
+        Rails.application.configure do
+          config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+        end
       end
     end
+
+    if defined?(ActionMailer)
+      # This will force the loading of ActionMailer settings
+      ActionMailer::Base.smtp_settings
+    end
   CODE
+  gsub_file 'config/initializers/action_mailer.rb',
+            /ExampleApp/,
+            Rails.application.class.parent.to_s
 
   copy_file 'spec/support/default_preview_path'
   chmod 'spec/support/default_preview_path', 0755
--- a/lib/rspec-rails.rb
+++ b/lib/rspec-rails.rb
@@ -52,17 +52,17 @@ module RSpec
       end
 
       def supports_action_mailer_previews?(config)
-        # If the action mailer railtie isn't loaded the config will not
-        # respond.
+        # These checks avoid loading `ActionMailer`. Using `defined?` has the
+        # side-effect of the class getting loaded if it is available. This is
+        # problematic because loading `ActionMailer::Base` will cause it to
+        # read the config settings; this is the only time the config is read.
+        # If the config is loaded now, any settings declared in a config block
+        # in an initializer will be ignored.
         #
-        # This string version check avoids loading the ActionMailer class, as
-        # would happen using `defined?`. This is necessary because the
-        # ActionMailer class only loads it's settings once, at load time. If we
-        # load the class now any settings declared in a config block in an
-        # initializer will be ignored.
-        #
-        # We cannot use `config.action_mailer.respond_to?(:preview_path)` here
-        # as it will always return `true`.
+        # If the action mailer railtie has not been loaded then `config` will
+        # not respond to the method. However, we cannot use
+        # `config.action_mailer.respond_to?(:preview_path)` here as it will
+        # always return `true`.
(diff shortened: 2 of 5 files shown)


example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -14,11 +14,25 @@ def comment_lines(path, flag, *args)
...
+      else
+        Rails.application.configure do
+          config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+        end
       end
     end
+
+    if defined?(ActionMailer)
+      # This will force the loading of ActionMailer settings
+      ActionMailer::Base.smtp_settings
```
JonRowe: Is the same on all versions of Rails?

lib/rspec-rails.rb
```diff
@@ -59,8 +52,18 @@ def config_default_preview_path(options)
...
+        # respond.
+        #
+        # This string version check avoids loading the ActionMailer class, as
+        # would happen using `defined?`. This is necessary because the
+        # ActionMailer class only loads it's settings once, at load time. If we
+        # load the class now any settings declared in a config block in an
+        # initializer will be ignored.
+        #
+        # We cannot use `config.action_mailer.respond_to?(:preview_path)` here
+        # as it will always return `true`.
```
JonRowe: Rails is a ghetto

example_app_generator/generate_action_mailer_specs.rb
cupakromer: From what I can tell.

example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -14,11 +14,25 @@ def comment_lines(path, flag, *args)
...
-      Rails.application.configure do
-        config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+      if ::Rails::VERSION::STRING < '4.1'
+        ExampleApp::Application.configure do
+          config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+        end
+      else
+        Rails.application.configure do
+          config.action_mailer.default_url_options = { :host => ENV['DEFAULT_URL'] }
+        end
```
JonRowe: Just curious why this change is related to this PR? Shouldn't we always have been doing this?

cupakromer: Yes, it should always have been there. Rails changed how they provided access to the configuration in 4.1. Incidentally the mailer specs / script never executed with a code path testing a default url option on Rails < 4.1. Which is also partially related to why this issue wasn't noticed sooner.

After fixing the mailer specs to run better coverage for Rails < 4.1, reproducing the issue, more things started failing as this code path was actually run. Thus, a fix is included here. The commit 9522842 has a long explanation of several related things which got fixed in the spec suite.

myronmarston:
Thanks for getting on this promptly, @cupakromer.  I'm happy to defer to @JonRowe's review since I don't really have the context for this.

JonRowe:
Looks all good to me

olance:
:+1:

myronmarston:
@cupakromer Are you planning to work up a 3-2-maintenance backport for this?

cupakromer:
@myronmarston done: #1389

Just waiting for the sanity check to go green.

chrisbloom7:
FYI Currently blocked on upgrading a Rails 4.0 app to use Ruby 2.2. This bug is preventing me from upgrading to rspec-rails 3.2, but issue #1273 prevents the suite from running on rspec-rails 3.1 (recommended solution: upgrade to 3.2)

myronmarston:
@chrisbloom7 -- I released rspec-rails 3.2.3 with a fix for this yesterday, so you should be able to upgrade now.

chrisbloom7:
@myronmarston Thanks, everything is green on Rails 4.0, rspec-rails 3.2.2, and Ruby 2.2.2

PR #1389, opened 2015-06-05
Back port pull request #1388 from rspec/fix-regression-mailer-preview-path

cupakromer:
Just a sanity check on the rails versions.


cupakromer:
@myronmarston all merged and green

PR #1396, opened 2015-06-15
Fix verified partial doubles with abstract classes.

JonRowe:
As reported in rspec/rspec-mocks#975 abstract classes can't be verified currently due to the callback for defining attribute methods. This corrects that.

Fixes rspec/rspec-mocks#975.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ Gemfile-custom
 bundle
 .bundle
 specs.out
+spec/examples.txt
--- a/lib/rspec/rails/active_record.rb
+++ b/lib/rspec/rails/active_record.rb
@@ -11,7 +11,8 @@ module RSpec
             ::RSpec::Mocks.configuration.when_declaring_verifying_double do |possible_model|
               target = possible_model.target
 
-              if target.respond_to?(:define_attribute_methods) && ActiveRecord::Base != target
+              if target.respond_to?(:define_attribute_methods) && ActiveRecord::Base != target &&
+                 target.respond_to?(:abstract_class?) && !target.abstract_class?
                 target.define_attribute_methods
               end
             end
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,11 @@
 ### 3.4.0 Dev
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.3.1...master)
 
+Bug Fixes:
+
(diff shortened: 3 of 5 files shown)


lib/rspec/rails/active_record.rb
```diff
@@ -11,7 +11,7 @@ def self.initialize_activerecord_configuration(config)
             ::RSpec::Mocks.configuration.when_declaring_verifying_double do |possible_model|
               target = possible_model.target
 
-              if target.respond_to?(:define_attribute_methods) && ActiveRecord::Base != target
+              if target.respond_to?(:define_attribute_methods) && ActiveRecord::Base != target && !target.abstract_class
```
myronmarston: Why `abstract_class` instead of `abstract_class?`

http://api.rubyonrails.org/classes/ActiveRecord/Inheritance/ClassMethods.html#method-i-abstract_class-3F

Also, I don't think that you can assume `target` will respond to `abstract_class` just because it responds to `define_attribute_methods`.  Consider that ActiveModel provides` define_attribute_methods` as well:

http://api.rubyonrails.org/classes/ActiveModel/AttributeMethods/ClassMethods.html#method-i-define_attribute_methods

So this needs a `respond_to?(:abstract_class)` check.

On a side note, I was hoping that `ActiveRecord::Base.abstract_class?` would return `true` (it is abstract after all...) so that we could do away with the explicit `ActiveRecord::Base != target` check but sadly, it doesn't :(.

myronmarston:
LGTM besides the concerns raised in my comment above.

.gitignore
```diff
@@ -1,4 +1,4 @@
-tmp
+pmp
```
cupakromer: Did you mean to change this?

cupakromer:
LGTM. I agree with Myron's comment.

JonRowe:
Changed and added a spec for active model based classes

myronmarston:
LGTM

myronmarston:
@JonRowe, before merging this, take a look at the conversation here:

https://github.com/rspec/rspec-core/issues/1994#issuecomment-112325603

I think it could be worthwhile to make further changes to the check in the callback here.

JonRowe:
Possibly but I think there might be further discussion needed given a more complex interaction, this will allow the original issue to be solved in master (possivly even released if it takes us a while) while we do that

PR #1398, opened 2015-06-16
Refactor verifying doubles callback

JonRowe:
As suggested in rspec/rspec-core#1994 this refactor's the rspec-mocks callback to only run on active record subclasses without `respond_to` checks.

### DIFF
--- a/lib/rspec/rails/active_record.rb
+++ b/lib/rspec/rails/active_record.rb
@@ -11,8 +11,7 @@ module RSpec
             ::RSpec::Mocks.configuration.when_declaring_verifying_double do |possible_model|
               target = possible_model.target
 
-              if target.respond_to?(:define_attribute_methods) && ActiveRecord::Base != target &&
-                 target.respond_to?(:abstract_class?) && !target.abstract_class?
+              if Class === target && ActiveRecord::Base > target && !target.abstract_class?
                 target.define_attribute_methods
               end
             end


JonRowe:
@cupakromer can you shed any light on why this is failing? as far as I know that method should exist on subclasses of active record

cupakromer:
This is due to Ruby's definition of `===`. It asks the question: "Is this an _instance_ of me or on of my descendants?" I'm surprised more specs didn't fail. The new conditional will fail when the target is a class (meaning `define_attribute_methods` will never get called when we need it), but pass when the target is an instance of the class.

```ruby
Parent = Class.new
Child = Class.new(Parent)
child_inst = Child.new

Parent === Child
# => false

Child.is_a?(Parent)
# => false

Parent === child_inst
# => true

Class === Child
# => true

Child.is_a?(Class)
# => true

Parent === Parent
# => false

Parent < Parent
# => false

Child < Parent
# => true

child_inst < Parent
# => NoMethodError: undefined method `<' for #<Child:0x007fa011d0c9c0>
```

We'll need to use a conditional which checks for classes first:

```ruby
target.is_a?(Class) && target < ActiveRecord::Base && !target.abstract_class?
# OR
Class === target && target < ActiveRecord::Base && !target.abstract_class?
```

I've verified that both versions do not cause the deadlock issue reported. However, I would probably prefer the 2nd version (using `Class === target`) so that we aren't sending a message to target as the first action; just in case.

cupakromer:
Maybe even change that to use `ActiveRecord::Base > target` instead so that it's not sending `>` to `target`. Sadly, we still need that class check too :sob: 

```ruby
Parent > Parent
# => false

Parent > Child
# => true

Parent > child_inst
# => TypeError: compared with non class/module
```

JonRowe:
New version should  work, had a derp moment with `===` I guess it was running the callbacks against instances!

JonRowe:
Updated

cupakromer:
LGTM. Feel free to merge on green.

myronmarston:
Should we backport this to 3-3-maintenance?

JonRowe:
Done.

Issue #1347, opened 2015-04-04
view specs doesn't render "current_page?" condition?


PascalTurbo:
Hi There,

I'm trying to test a css class generated while using the current_page? helper:

    RSpec.describe 'my_day/index.html.erb', type: :view do
      it 'should have an active link to my_days' do
        render template: 'my_day/index', layout: 'layouts/application'
        expect(rendered).to have_css('li.active a', text: 'MyDays')
      end
    end

    <li class="<%= 'active' if current_page?(controller: 'my_day', action: 'index') %>"><%= link_to('MyDays', my_day_index_path) %></li>

But this will never match.

So I tried setting the condition true:

    <li class="<%= 'active' if true %>"><%= link_to('MyDays', my_day_index_path) %></li>

And this works. So I think that current_page? isn't evaluated correctly.

cupakromer:
:anguished: I can see why this behavior is surprising. TLDR; you will need to [stub this helper](https://www.relishapp.com/rspec/rspec-rails/v/3-2/docs/view-specs/view-spec#passing-view-spec-that-stubs-a-helper-method) in your spec :heart::

```ruby
allow(view).to receive(:current_page?).with(
  controller: 'my_day',
  action: 'index'
).and_return(true)
```

RSpec view specs are intended to allow testing view templates in isolation. When working with a view template in isolation the idea is to treat it as detached from a controller, action, or request path.

However, many applications tend to have a single template used by a single controller and action. RSpec tries to help with this by adding that information via a `controller` object: https://www.relishapp.com/rspec/rspec-rails/v/3-2/docs/view-specs/view-spec-infers-controller-path-and-action

The root of the issue is that `current_page?` doesn't use this information. Checking the [current source](https://github.com/rails/rails/blob/67bb09e28354d203f9c247fef3f4b52af30316ea/actionview/lib/action_view/helpers/url_helper.rb#L539) shows this implementation of the helper only checks the request path.

I think we may be able to adjust things to accommodate this.

samphippen:
Closing because of #1402

Issue #1332, opened 2015-03-05
Rails Test log says my templates were rendered when they actually weren't


tenderlove:
When I run a controller spec, the test log says that my templates were rendered when they actually weren't rendered, also the response body is an empty string.

Steps to reproduce:

```
$ rails new myapp
$ cd myapp
$ echo 'gem "rspec-rails", :group => [:development, :test]' >> Gemfile
$ bundle
$ rails generate scaffold Widgets name:string
$ rake db:migrate && rake db:test:prepare
$ rake spec
$ rm log/test.log
$ bundle exec rspec spec/controllers/widgets_controller_spec.rb
```

After the final run, cat `log/test.log`.  Many times the log file will say something like:

```
Rendered widgets/new.html.erb
```

When in fact my templates were not rendered at all.  Since the templates were not actually rendered, these logs should not be in the file.

Secondly, calling `response.body` in my spec always returns a blank string.  If the templates were not rendered, and there isn't actually a response body, I'd expect either a message in the response body telling me that _or_ the method to not be available on the response. (side note: if controller tests are meant to just test the controller, why is the response object available at all?)

I realize now that I'm supposed to add `render_views` to my test to get it to render views, but as someone new to RSpec, the behavior _without_ `render_views` in my spec is extremely confusing.

searls:
:+1:

sgrif:
Sounds like a Rails bug. Someone should inform the core team. :trollface:

cupakromer:
Sorry this was a confusing experience. :frowning: 

> I'd expect either a message in the response body telling me that or the method to not be available on the response.

I think a message in the body is probably the way to go. I don't feel comfortable removing / redefining a method on `ActionController::TestResponse`.

> I realize now that I'm supposed to add `render_views` to my test to get it to render views, but as someone new to RSpec, the behavior without `render_views` in my spec is extremely confusing.

I'd like to understand more why is this confusing behavior. Is this something we need to improve in the documentation? Or does the confusion stem from your long use of Minitest, where your expectations of how `ActionController::TestCase` works is conflicting with our intention of a "controller" spec?

While controller specs _are_ a thin shim on top of `ActionController::TestCase`. Some functionality, as in `render_views`, isn't run by default. The intent is treat the controller specs more in an "isolated" manner, only encouraging testing the controller's contract. We offer `view` specs and `request` specs for isolated view and fuller integration testing respectively.

cupakromer:
[...]

PR #1402, opened 2015-06-18
Expose path to view specs

sarahmei:
Fix for https://github.com/rspec/rspec-rails/issues/1347.

Expose inferred path data to view specs, since we already expose inferred controller & action. This will allow view specs to exercise views that depend on the current page.

We did two implementations - the first commit uses `rescue nil` - the second one has no rescues but uses somewhat awkward logic to iterate through the routes to find a matching one.

If you know of a better way to implement this without rescue...would love to hear about it.

### DIFF
--- a/features/view_specs/view_spec.feature
+++ b/features/view_specs/view_spec.feature
@@ -2,7 +2,7 @@ Feature: view spec
 
   View specs live in spec/views and render view templates in isolation.
 
-  Scenario: passing spec that renders the described view file
+  Scenario: View specs render the described view file
     Given a file named "spec/views/widgets/index.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -24,7 +24,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with before and nesting
+  Scenario: View specs can have before block and nesting
     Given a file named "spec/views/widgets/index.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -51,7 +51,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with explicit template rendering
+  Scenario: View specs can explicitly render templates
     Given a file named "spec/views/widgets/widget.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -73,7 +73,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with a description that includes the format and handler
+  Scenario: View specs can have description that includes the format and handler
     Given a file named "spec/views/widgets/widget.xml.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -105,7 +105,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with rendering of locals in a partial
+  Scenario: View specs can render locals in a partial
     Given a file named "spec/views/widgets/_widget.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -127,7 +127,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with rendering of locals in an implicit partial
+  Scenario: View specs can render locals in an implicit partial
     Given a file named "spec/views/widgets/_widget.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -149,7 +149,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing spec with rendering of text
+  Scenario: View specs can render text
     Given a file named "spec/views/widgets/direct.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -166,7 +166,7 @@ Feature: view spec
     When I run `rspec spec/views`
     Then the examples should all pass
 
-  Scenario: passing view spec that stubs a helper method
+  Scenario: View specs can stub a helper method
     Given a file named "app/helpers/application_helper.rb" with:
       """ruby
       module ApplicationHelper
@@ -199,7 +199,7 @@ Feature: view spec
     When I run `rspec spec/views/secrets`
     Then the examples should all pass
 
-  Scenario: request.path_parameters should match Rails by using symbols for keys
+  Scenario: View specs use symbols for keys in `request.path_parameters` to match Rails style
     Given a file named "spec/views/widgets/index.html.erb_spec.rb" with:
       """ruby
       require "rails_helper"
@@ -207,6 +207,37 @@ Feature: view spec
       RSpec.describe "controller.request.path_parameters" do
         it "matches the Rails environment by using symbols for keys" do
           [:controller, :action].each { |k| expect(controller.request.path_parameters.keys).to include(k) }
+        end
+      end
+      """
+    When I run `rspec spec/views`
+    Then the examples should all pass
+
+  Scenario: View spec actions that do not require extra parameters have `request.fullpath` set
+    Given a file named "spec/views/widgets/index.html.erb_spec.rb" with:
+    """ruby
+      require "rails_helper"
+
+      RSpec.describe "widgets/index" do
+        it "has a request.fullpath that is defined" do
+          expect(controller.request.fullpath).to eq widgets_path
+        end
+      end
+      """
+    When I run `rspec spec/views`
+    Then the examples should all pass
+
+  Scenario: View spec actions that require extra parameters have `request.fullpath` set when the developer supplies them
+    Given a file named "spec/views/widgets/show.html.erb_spec.rb" with:
+    """ruby
+      require "rails_helper"
+
+      RSpec.describe "widgets/show" do
+        it "displays the widget with id: 1" do
+          widget = Widget.create!(:name => "slicer")
+          controller.extra_params = { :id => widget.id }
+
+          expect(controller.request.fullpath).to eq widget_path(widget)
         end
       end
       """
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -1,4 +1,6 @@
 require 'rspec/rails/view_assigns'
+require 'rspec/rails/view_spec_methods'
+require 'rspec/rails/view_path_builder'
 
 module RSpec
   module Rails
@@ -159,6 +161,12 @@ module RSpec
           controller.controller_path = _controller_path
           controller.request.path_parameters[:controller] = _controller_path
           controller.request.path_parameters[:action]     = _inferred_action unless _inferred_action =~ /^_/
+          controller.request.path = ViewPathBuilder.new(::Rails.application.routes).path_for(controller.request.path_parameters)
+          ViewSpecMethods.add_to(::ActionView::TestCase::TestController)
+        end
+
+        after do
+          ViewSpecMethods.remove_from(::ActionView::TestCase::TestController)
         end
 
         let(:_default_file_to_render) do |example|
--- a/lib/rspec/rails/view_path_builder.rb
+++ b/lib/rspec/rails/view_path_builder.rb
@@ -1,0 +1,29 @@
+module RSpec
+  module Rails
+    # Builds paths for view specs using a particular route set.
+    class ViewPathBuilder
+      def initialize(route_set)
+        self.class.send(:include, route_set.url_helpers)
+      end
+
+      # Given a hash of parameters, build a view path, if possible.
+      # Returns nil if no path can be built from the given params.
+      #
+      # @example
+      #     # path can be built because all required params are present in the hash
+      #     view_path_builder = ViewPathBuilder.new(::Rails.application.routes)
+      #     view_path_builder.path_for({ :controller => 'posts', :action => 'show', :id => '54' })
+      #     # => "/post/54"
+      #
+      # @example
+      #     # path cannot be built because the params are missing a required element (:id)
+      #     view_path_builder.path_for({ :controller => 'posts', :action => 'delete' })
+      #     # => nil
+      def path_for(path_params)
+        url_for(path_params.merge(:only_path => true))
+      rescue => e
+        e.message
+      end
+    end
+  end
+end
--- a/lib/rspec/rails/view_spec_methods.rb
+++ b/lib/rspec/rails/view_spec_methods.rb
@@ -1,0 +1,56 @@
+module RSpec
+  module Rails
+    # Adds methods (generally to ActionView::TestCase::TestController).
+    # Intended for use in view specs.
+    module ViewSpecMethods
+      module_function
+
+      # Adds methods `extra_params=` and `extra_params` to the indicated class.
+      # When class is `::ActionView::TestCase::TestController`, these methods
+      # are exposed in view specs on the `controller` object.
+      def add_to(klass)
+        return if klass.method_defined?(:extra_params) && klass.method_defined?(:extra_params=)
+
+        klass.module_exec do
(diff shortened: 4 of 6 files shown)


lib/rspec/rails/view_controller.rb
```diff
@@ -0,0 +1,43 @@
+module RSpec
+  module Rails
+    # Methods added to ActionView::TestCase::TestController, only in view specs
+    module ViewController
+      def self.included(_klass)
```
samphippen: Unless I'm mistaken, this will permanently include the `ExtraParamsAccessors` methods into ActionView::TestCase::TestController for the rest of the process, including other types of specs. Is my read on this right, and is that the intended behaviour? To me, this looks like it's basically lazy addition of some methods to `ActionView::TestCase::TestController` as opposed to only including some methods in to view specs. If that's the intention, then does the comment need to be changed to better reflect that's whats happening?

PS: If I'm wrong, please let me know, I'm not super familiar with this code ^_^

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,52 @@
...
+
+      def route_exists?(path_params)
+        path_keys = path_params.keys
+
+        routes.any? do |route|
+          actions_match = (route.defaults[:action] == path_params[:action])
+          controllers_match = (route.defaults[:controller] == path_params[:controller])
+
+          required_keys =
+            if route.respond_to?(:required_keys) # Rails 3.2 and above
```
samphippen: the `= if` constructs that I've seen elsewhere have the if on the same line of the equals, it's a minor quibble and not a merge blocker, but if you do go back over this, it'd be great if you could  bring these on to the same line.

samphippen:
I've left a few comments. @cupakromer if you agree/have further questions I'd love if you could take a look at this when you get a chance.

samphippen:
Open question: does this need to be squashed before merging?

lib/rspec/rails/view_controller.rb
sarahmei: I hadn't thought about that distinction. The intention is to include them only for view specs. It's an awkward fit because the user has to set them somehow, and they've already got a controller object available. I thought about surfacing a whole different object, but the extra params are not unrelated to the controller, request, etc. 

That said - what's the right way to include those methods in the controller class just for view specs?

samphippen: @cupakromer would know better than me, but my instincts are saying before/after that adds/remove methods in a similar way to how syntax.rb works: https://github.com/rspec/rspec-expectations/blob/master/lib/rspec/expectations/syntax.rb#L38-L64

JonRowe:
I'd like it if we could get this squashed :)

features/view_specs/view_spec.feature
```diff
@@ -212,3 +212,34 @@ Feature: view spec
...
+        it "displays the widget with id: 1" do
+          widget = Widget.create!(:name => "slicer")
+          controller.extra_params = { :id => widget.id }
+
+          expect(controller.request.fullpath).to eq widget_path(widget)
+        end
+      end
+      """
+    When I run `rspec spec/views`
+    Then the examples should all pass
```
cupakromer: Yay cukes! :blue_heart: I recognize that the style for these mirrors the existing spec. If you'd be so kind to update them with a newer style we working towards that would be great!

- Use backticks for code in descriptions (including scenarios) as in normal markdown, so the scenarios should use _\`request.fullpath\`_
- Scenarios should start capitalized; does not apply when starting with a code fragment
- Avoid "should" opting for more direct and present language. For example "Actions which do not require extra parameters have `request.fullpath` set"

spec/rspec/rails/view_spec_methods_spec.rb
```diff
@@ -0,0 +1,62 @@
+require "spec_helper"
```
cupakromer: We can drop this now that it's in `.rspec`. :smile_cat:

spec/rspec/rails/view_spec_methods_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module RSpec::Rails
+  describe ViewSpecMethods do
```
cupakromer: I haven't set it just yet, but I'd like to move towards zero monkey patching mode. Could we use `RSpec.describe` for this outter most `describe`? (It's the only one which needs it; inside this block `describe` is available as part of the block's DSL)

lib/rspec/rails/example/view_example_group.rb
```diff
@@ -161,6 +163,12 @@ def _include_controller_helpers
           controller.controller_path = _controller_path
           controller.request.path_parameters[:controller] = _controller_path
           controller.request.path_parameters[:action]     = _inferred_action unless _inferred_action =~ /^_/
+          controller.request.path = ViewPathBuilder.new(::Rails.application.routes).path_for(controller.request.path_parameters)
+          ViewSpecMethods.add_to(::ActionView::TestCase::TestController)
+        end
+
+        after do
+          ViewSpecMethods.remove_from(::ActionView::TestCase::TestController)
```
cupakromer: Do we need these hooks to add the `extra_params`? I ask as there is access to [`params`](https://github.com/rspec/rspec-rails/blob/1fb6f9f642e887df1b1ee344594de02d7de3b3fe/lib/rspec/rails/example/view_example_group.rb#L77-L83) which is `public` in the view spec. We don't appear to have a feature scenario demoing it :grimacing: 

Would that be an alternative API to use? I may be missing some sort of glaring issue with how merging that with the path setup could cause issues.

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,53 @@
+module RSpec
+  module Rails
+    # Builds paths for view specs using a particular route set.
+    class ViewPathBuilder
+      def initialize(route_set)
+        @routes = route_set.routes
+        self.class.send(:include, route_set.url_helpers)
```
cupakromer: Does this always re-add the route helpers to the `ViewPathBuilder` class? I'm a little tired, I think I am missing something. If this uses `extend` on the object does it accomplish the same thing without affecting `ViewPathBuilder`?

```ruby
self.extend route_set.url_helpers
```

cupakromer: After a 2nd look at the class, is this used somewhere in the class? Is this how we're getting access to `url_for`?

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,53 @@
...
+      #     # => "/post/54"
+      #
+      # @example
+      #     # path cannot be built because the params are missing a required element (:id)
+      #     view_path_builder.path_for({ :controller => 'posts', :action => 'delete' })
+      #     # => nil
+      def path_for(path_params)
+        return nil unless route_exists?(path_params)
+        url_for(path_params.merge(:only_path => true))
+      end
```
cupakromer: Could we return a helpful message as the path? Something like: "No path for { hash details } could be found."

This way if someone relies on this being set, and it fails, we give them some details instead of just a `nil`.

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,53 @@
...
+        routes.any? do |route|
+          actions_match = (route.defaults[:action] == path_params[:action])
+          controllers_match = (route.defaults[:controller] == path_params[:controller])
+
+          required_keys = if route.respond_to?(:required_keys) # Rails 3.2 and above
+                            route.required_keys
+                          else
+                            route.segment_keys - [:format]
+                          end
+          has_all_other_keys = required_keys.all? { |key| path_keys.include?(key) }
```
cupakromer: What do you think about:

```ruby
has_all_other_keys = (required_keys - path_keys).empty?
```

cupakromer:
I like how this is progressing. :smile_cat: : I like seeing the progress over the commit history. I'd really like to persevere a bunch of that if we can. I think when doing the final squash, if the ideas, the _why_ of the current design decisions, and details about attempted but ultimately removed design decisions would be awesome :blue_heart: :blue_heart:

In general I strongly avoid using `rescue` for logic flow. I know @sarahmei and I talked about this a bit on IRC before I went MIA being all :mask: :sleeping:. Coming back to this with fresh eye, I'm wondering if `rescue` may actually be the best strategy in this case. My main concern is legacy and monolith apps. I do not have any hard numbers currently (will ask around tomorrow), but I'm thinking of those apps with possibly > 100 routes.

My thinking goes:

  - `url_for` is called all over the place in Rails; so they need to do that look up a lot and I feel better _assuming_ they've optimize the logic better than I could
  - this runs for each view spec and in large apps could really slow down view specs (of course we'd need to bench mark it)
  - we call `url_for` either way, in the rare edge case where we may have missed some logic in our lookup, we'll still get the exception; which we are no longer rescuing

I'm happy to talk to some people tomorrow and get some hard numbers are larger apps and their route count. I'm also happy to test out a benchmark using `rescue` and our lookup to see which is faster.

lib/rspec/rails/example/view_example_group.rb
sarahmei: At this point, we're before any of the user's code, so they haven't set params yet. That means that when they do set params, later, we need to re-run the path creation. It felt odd to override `params#[]` to do that, so we opted to add new functionality. We can do the `params` thing instead if you'd rather.

lib/rspec/rails/view_path_builder.rb
sarahmei: Yep, that's how we get `url_for`.

sarahmei:
@IMRYO and I are working on a new branch that does this with a `rescue` instead of iterating, and also incorporates these suggestions.

sarahmei:
@cupakromer @samphippen what's your recommendation for dealing with code that's different in 1.8.7 vs 1.9? `instance_methods` has string keys in 1.8.7 and symbol keys in 1.9+. :frowning:

samphippen:
`.find {|x| x.to_sym == the_thing}.nil?` is probably the best I can come up with right now. @cupakromer thoughts?

myronmarston:
We often conditionally define one of two method implementations based on a Ruby version. Would that work here? (On my phone so hard to link you to examples right now)

myronmarston:
Actually, looking at the include check, I think that 'klass.method_defined?(name)' would work best. It works with strings or symbols on 1.8 and 1.9+ and is constant time instead of linear.

sarahmei:
Thanks @myronmarston! `method_defined?` worked on all the rubies. @cupakromer this is ready to look at again. :)

JonRowe:
Ping @cupakromer, did you get a chance to rereview this?

cupakromer:
Sorry, slipped off my todo list. I _will_ get to this today :-D

.gitignore
```diff
@@ -21,3 +21,4 @@ bundle
 .bundle
 specs.out
 spec/examples.txt
+specs.out
```
cupakromer: :heart:

lib/rspec/rails/example/view_example_group.rb
cupakromer: My only concern is adding a new API to the spec. If for some reason my view does look at `params`, say `params[:id]`, and looks at the path (I know not something that should be done, but it still could happen), the `id` value will need to get set twice. That seems a bit odd to me.

I thought about suggesting we override `render` in the spec, but then I realized that my standard debugging technique wouldn't necessarily work there:

```ruby
p controller.request.path
render
```

Instead the inspection would need to occur after `render`. Which may not be an issue unless `render` causes an exception in which case inspecting the path wouldn't be possible without more hackery.

I'm honestly a bit torn between the extra API and having to shim around `params`.

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,27 @@
+module RSpec
+  module Rails
+    # Builds paths for view specs using a particular route set.
+    class ViewPathBuilder
+      def initialize(route_set)
+        self.class.send(:include, route_set.url_helpers)
+      end
```
cupakromer: Is there a benefit to using this over storing the `route_set` and calling `route_set.url_for` below?

lib/rspec/rails/view_path_builder.rb
```diff
@@ -0,0 +1,27 @@
...
+      #     view_path_builder.path_for({ :controller => 'posts', :action => 'show', :id => '54' })
+      #     # => "/post/54"
+      #
+      # @example
+      #     # path cannot be built because the params are missing a required element (:id)
+      #     view_path_builder.path_for({ :controller => 'posts', :action => 'delete' })
+      #     # => nil
+      def path_for(path_params)
+        url_for(path_params.merge(:only_path => true)) rescue nil
+      end
```
cupakromer: :heart: on the very thorough docs!

cupakromer: what are your thoughts on something like:

```ruby
def path_for(path_params)
  url_for(path_params.merge(:only_path => true))
rescue => e
  e.message
end
```

I suggest this because we've had requests (for example #1332) to provide better info into fields which may or may not get set.

lib/rspec/rails/view_spec_methods.rb
```diff
@@ -0,0 +1,55 @@
...
+          # Use to read extra parameters that are set in the view spec.
+          #
+          # @example
+          #
+          #     # After the before in the above example:
+          #     controller.extra_params
+          #     # => { :id => 4 }
+          def extra_params
+            @extra_params ||= {}
+          end
```
cupakromer: While I see the intent for this is to simply read the extra params, I'm afraid that people will instead use this as:

```ruby
controller.extra_params[:id] = 1
```

This would appear to work, but the path will not get updated. On the surface this seems like the same issue you raised about overriding `params`. I think I may be missing something. :smile_cat: Thoughts?

spec/rspec/rails/view_spec_methods_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    describe ".add_extra_params_accessors_to" do
+      describe "when accessors are not yet defined" do
+        it "adds them as instance methods" do
+          ViewSpecMethods.add_to(VCSampleClass)
+
+          expect(VCSampleClass.instance_methods).to(include("extra_params=").or(include(:extra_params=)))
+          expect(VCSampleClass.instance_methods).to(include("extra_params").or(include(:extra_params)))
+        end
```
cupakromer: Is there a Ruby version reason to have both string and symbol checks here?

JonRowe: Yes.

IAMRYO:
What's the status on this PR?  Seems like it's ready to be merged. Any other questions that need to be answered?

spec/rspec/rails/view_spec_methods_spec.rb
JonRowe: (Older versions of ruby return them as strings whilst later as symbols)

lib/rspec/rails/view_spec_methods.rb
IAMRYO: Hey @cupakromer do you think a comment here to the effect of "Don't do that" would suffice?

IAMRYO:
@cupakromer I think we have one more comment to straighten out before this can be accepted. It would be great to finish this up.

spec/rspec/rails/view_spec_methods_spec.rb
samphippen: shouldn't this just normalize instead of using the either or check?

JonRowe: Possibly.

features/view_specs/view_spec.feature
samphippen: This is resolved, and it's a hangover comment that github didn't squash.

spec/rspec/rails/view_spec_methods_spec.rb
samphippen: Dealt with

lib/rspec/rails/view_path_builder.rb
samphippen: This is fine.

Issue #1403, opened 2015-06-18
Option not to pluralize feature name on generation


carlesjove:
When generating a `feature` with `rails g rspec:feature` its name is always pluralized. Sometimes this leads to undesired results, such as `user_manages_account` being pluralized into `accounts`. The description within the spec file will be pluralized, too. This requires renaming the file plus editing the description.

I would expect the generator *not* to pluralize the name.

However, it'd be also nice if a `singular` or `no-pluralize` option could be passed.

I could try to implement this if it's accepted.

cupakromer:
Did you manually generate the associated controller / models or use a generator?

carlesjove:
For features, I always generate them independently, so I run `g rspec:feature`.

cupakromer:
@carlesjove yes, I see what you are saying. I agree I do not think pluralizing the name is generally what you want for a feature spec. Additionally, I think it would be nice to be able to provide it a string description which it would properly handle in the file name as well:

```console
$ bin/rails g rspec:feature "User manages account"
      create  spec/features/user_manages_account_spec.rb
```

I'm sure how useful additional options would be considering the generated file is very minimal:

```ruby
require 'rails_helper'

RSpec.feature "User manages account", type: :feature do
  pending "add some scenarios (or delete) #{__FILE__}"
end
```

Though I am always open to the community stating otherwise.

PR #1426, opened 2015-07-28
#1403 : string description for feature specs

sivagollapalli:
#1403 string description for feature specs

### DIFF
--- a/lib/generators/rspec/feature/feature_generator.rb
+++ b/lib/generators/rspec/feature/feature_generator.rb
@@ -8,7 +8,7 @@ module Rspec
 
       def generate_feature_spec
         return unless options[:feature_specs]
-        file_name = table_name.split(" ").collect(&:downcase).join("_")
+        file_name = table_name.parameterize.underscore
 
         template 'feature_spec.rb', File.join('spec/features', class_path, "#{file_name}_spec.rb") # file_name?
       end
--- a/lib/generators/rspec/feature/feature_generator.rb
+++ b/lib/generators/rspec/feature/feature_generator.rb
@@ -8,8 +8,9 @@ module Rspec
 
       def generate_feature_spec
         return unless options[:feature_specs]
+        file_name = table_name.split(" ").collect(&:downcase).join("_")
 
-        template 'feature_spec.rb', File.join('spec/features', class_path, "#{table_name}_spec.rb") # file_name?
+        template 'feature_spec.rb', File.join('spec/features', class_path, "#{file_name}_spec.rb") # file_name?
       end
     end
   end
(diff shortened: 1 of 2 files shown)


cupakromer:
Thanks for submitting this :heart:

Looks like the issue is with one of the rubocop's. I think simply changing `collect` to `map` will fix it.

lib/generators/rspec/feature/feature_generator.rb
```diff
@@ -8,8 +8,9 @@ class FeatureGenerator < Base
 
       def generate_feature_spec
         return unless options[:feature_specs]
+        file_name = table_name.split(" ").collect(&:downcase).join("_")
```
cupakromer: Since this is in a generator Rails, and Active Support, have already been loaded. Perhaps change this to the following to provide a wider range of character conversion support:

```ruby
file_name = table_name.parameterize.underscore
```

Thoughts?

sivagollapalli:
@cupakromer Thanks for your suggestion. I have updated my PR with latest changes. Please check.

cupakromer:
Looking good so far. I double checked and I think `table_name` isn't what we want here. Issue #1403 is asking about not pluralizing the names. As written this PR update will not fix that.

I think what you want is simply `name` instead of `table_name`. I believe that should be what the user passes in on the command line. Additionally the generator template will need an update to not pluralize the name there as well.

However, thinking about this a little more this change would break existing functionality for generators that map to expected constants or routes. Here are some command samples with the file and spec name I would expect to get generated:

```console
$ bin/rails g rspec:feature Admin::WidgetsController

# spec/features/admin/widgets_controller_spec.rb
RSpec.feature "Admin::WidgetsController", type: :feature do
# ...
end

$ bin/rails g rspec:feature api/v1/widgets

# spec/features/api/v1/widgets_spec.rb
RSpec.feature "api/v1/widgets", type: :feature do
# ...
end

$ bin/rails g rspec:feature "Hello! This is my first RSpec::Rails feature"

# spec/features/hello_this_is_my_first_rspec_rails_feature_spec.rb
RSpec.feature "Hello! This is my first RSpec::Rails feature", type: :feature do
# ...
end
```

sivagollapalli:
@cupakromer  I am thinking to modify `table_name.parameterize.underscore` to `name.camelize.demodulize.parameterize.underscore`. But this line would satisfy following one scenario:

```
$ bin/rails g rspec:feature api/v1/widgets

# spec/features/api/v1/widgets_spec.rb
RSpec.feature "api/v1/widgets", type: :feature do
  # ...
end
```

I am stuck to how to get it pass the remaining two scenarios. Could you help me?

sivagollapalli:
@cupakromer Could you please reply?

cupakromer:
:crying_cat_face: sorry I completely dropped the ball on this one. Closing in favor of #1503

PR #1429, opened 2015-07-30
Extend helper feature tests.

driv3r:
Lately I was struggling with stubbing current user in helper tests, with some tries I found this behaviours:
1. Methods from helper are defined in ExampleGroup as well
2. Methods defined within ExampleGroup are then also visible within those methods

Stubbing current user for helper ended up with error
```ruby
allow(helper).to receive(:current_user).and_return(User.new)
# => #<#<Class:0x00000004585be8>:0x00000004509f20 .................... > does not implement: current_user
```
And of course it's because it's defined in controllers. So the question is how else can we stub helper methods from controllers in helpers?

### DIFF
--- a/features/helper_specs/helper_spec.feature
+++ b/features/helper_specs/helper_spec.feature
@@ -9,7 +9,7 @@ Feature: helper spec
   your app.
 
   To access the helper methods you're specifying, simply call them directly
-  on the `helper` object.
+  on the `helper` object, or call them directly within the example
 
   NOTE: helper methods defined in controllers are not included.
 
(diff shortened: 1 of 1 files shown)


features/helper_specs/helper_spec.feature
```diff
@@ -120,3 +120,96 @@ Feature: helper spec
...
+    And a file named "app/helpers/user_helper.rb" with:
+      """ruby
+      module UserHelper
+        def email
+          "jon.snow@example.com"
+        end
+      end
+      """
+    When I run `rspec spec/helpers/user_helper_spec.rb`
+    Then the examples should all pass
```
cupakromer: I'm a bit surprised by this behavior. I think it may is a side-effect of the current implementation, but it is not the intended usage of a helper spec. I'd rather this not be documented; I actually view this as a :bug:.

features/helper_specs/helper_spec.feature
```diff
@@ -120,3 +120,96 @@ Feature: helper spec
...
+    And a file named "app/helpers/user_helper.rb" with:
+      """ruby
+      module UserHelper
+        def email
+          current_user.nil? ? 'n/a' : current_user.email
+        end
+      end
+      """
+    When I run `rspec spec/helpers/user_helper_spec.rb`
+    Then the examples should all pass
```
cupakromer: This is also some very odd behavior. Combining this with the prior example, it really feels like there's a lot of state leak in these spec types. Stated a slightly different way, the barrier between test and object undercuts doesn't exist. This is bad. :frowning: 

This can easily lead to some very surprising and unexpected behaviors:

```ruby
require 'rails_helper'

module UserHelper
  def email
    :user_helper
  end
end

RSpec.describe UserHelper, :type => :helper do

  it "directly accesses the helper message" do
    expect(email).to eq :user_helper
  end

  it "the helper uses it's definition" do
    expect(helper.email).to eq :user_helper
  end

  context "with an inner context defining a conflicting let" do
    let(:email) { :let_memo }

    it "uses the let object" do
      expect(email).to eq :let_memo
    end

    it "the helper uses it's definition" do
      expect(helper.email).to eq :user_helper
    end

  end

end

RSpec.describe UserHelper, "with a conflicting let", :type => :helper do

  let(:email) { :let_memo }

  it "accesses the helper message" do
    expect(email).to eq :user_helper
  end

  it "the helper uses it's definition" do
    expect(helper.email).to eq :user_helper
  end

end

RSpec.describe UserHelper, "with a conflicting method", :type => :helper do

  def email
    :custom_method
  end

  it "accesses the conflicting message" do
    expect(email).to eq :custom_method
  end

  it "the helper uses it's definition" do
    expect(helper.email).to eq :user_helper
  end

end

RSpec.describe UserHelper, "conflicts are order dependent", :type => :helper do

  def email
    :custom_method
  end

  let(:email) { :let_memo }

  it "accesses the helper message" do
    expect(email).to eq :user_helper
  end

  it "the helper uses it's definition" do
    expect(helper.email).to eq :user_helper
  end

end

RSpec.describe UserHelper, "conflicts are order dependent", :type => :helper do

  let(:email) { :let_memo }

  def email
    :custom_method
  end

  it "accesses the conflicting message" do
    expect(email).to eq :custom_method
  end

  it "the helper uses it's definition" do
    expect(helper.email).to eq :user_helper
  end

end
```

features/helper_specs/helper_spec.feature
```diff
@@ -120,3 +120,96 @@ Feature: helper spec
...
+    And a file named "app/helpers/user_helper.rb" with:
+      """ruby
+      module UserHelper
+        def email
+          current_user.email
+        end
+      end
+      """
+    When I run `rspec spec/helpers/user_helper_spec.rb`
+    Then the examples should all pass
```
cupakromer: Given the surprising behavior illustrated above I think this is a spec we will need to add. I'm not sure that this level of the docs is the proper place; as I would never expect this to work. I think this type of check should be belong in the regular spec suite. If for some reason that's not possible then the internal acceptance suite would be a good place.

cupakromer:
Thank you for taking the time to put this together :blue_heart: :blue_heart: 

Unfortunately, this PR as an addition to the public relish docs isn't something as it that I think we'll be able to accept. The behaviors documented here are not the intended behavior of helper specs :cold_sweat: 

I _really_ appreciate your effort and work on this :smile_cat:.  The working examples have clearly pointed out some major flaws in the helper spec implementations. They also are a great start as future implementation checks that should be added / moved into the regular spec suite.

To answer your initial question, my current suggestion for controller based helpers are to make them into a module which your application would include. However, to server as documentation if that's not an option, I would consider starting with one of the following approaches:

```ruby
require 'rails_helper'

module UserHelper
  def email
    current_user.email
  end
end

RSpec.describe UserHelper, :type => :helper do

  specify "uses the controller helper's current user" do
    controller_helper = Module.new do
      def current_user
        User.new(email: "any email")
      end
    end
    helper.extend controller_helper

    expect(helper.email).to eq "any email"
  end

  another_controller_helper = Module.new do
    def current_user; end
  end

  specify "stubbing the helper under test feels awkward" do
    helper.extend another_controller_helper
    allow(helper).to receive(:current_user).and_return(
      instance_double(User, email: :any_email)
    )

    expect(helper.email).to eq :any_email
  end
end
```

Given the slightly awkward nature of that and accessing `let` and other helpers in the module can be a little odd I would be open to considering a new DSL method. Perhaps something like:

```ruby
RSpec.describe UserHelper, :type => :helper do

  let(:an_email) { :let_memo }

  controller_helper(:current_user) do
    instance_double(User, email: an_email)
  end

  specify "the anonymous module is setup and extended for you" do
    expect(helper.email).to eq an_email
  end

end
```

Thoughts?

JonRowe:
I'd just like to :+1: @cupakromer's comments, I'm very surprised you can access the methods directly in the test, that seems insane.

driv3r:
Thanks @cupakromer and yes those examples were mainly to show off the weird behaviour. 

`controller_helper` method would be great! Although we can probably use this anonymous module in mean time.

samphippen:
@driv3r do you have the time to finish this pull request off? We'd love to see it merged.

JonRowe:
@samphippen this PR is unrelated to the changes we'd actually like to see, as @cupakromer mentioned this behaviour is currently a bug not a feature, I'm going to close this because any changes arising from this discussion would be suitably different to the original content.

Issue #1430, opened 2015-07-30
Another "undefined method fixture_path" error


cyu:
#1372 fixes one "undefined method fixture_path" error, but created another.

The issues starts here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/configuration.rb#L50.  Note that this line is 18 lines before we're expecting to load `RSpec::Rails::FixtureSupport`.

Loading `RSpec::Rails::FeatureExampleGroup` loads `RSpec::Rails::RailsExampleGroup` here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/example/feature_example_group.rb#L6

`RSpec::Rails::RailsExampleGroup` loads `RSpec::Rails::FixtureSupport` here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/example/rails_example_group.rb#L14

And `RSpec::Rails::FixtureSupport` references `RSpec.configuration.fixture_path`, here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/fixture_support.rb#L20, which doesn't exist until here: https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/configuration.rb#L67

cupakromer:
Thanks for noticing that. Any idea why the spec didn't run into that?

PR #1436, opened 2015-08-01
Resolving #1430 by relocating

jdax:


cupakromer:
Thanks for fixing this :smiley:

PR #1437, opened 2015-08-01
Rubocop Upgrade

jamelablack:
This provides a basic update of Rubocop to 0.32.1 adding support for checking on Ruby 2.2. The following offenses have been left failing to allow for determining if they should simply be disabled:

> lib/rspec/rails/matchers/have_http_status.rb:183:21: C: Do not use trailing _s in
parallel assignment.

```
            status, _ = Rack::Utils::SYMBOL_TO_STATUS_CODE.find do |_, c|
                    ^
```

> lib/rspec/rails/tasks/rspec.rake:2:12: W: Assignment in condition - you probably meant
to use ==.

```
if default = Rake.application.instance_variable_get('@tasks')['default']
           ^
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,13 +6,16 @@ LineLength:
 
 Lint/NestedMethodDefinition:
   Enabled: false
-    
+
 # Over time we'd like to get this down, but this is what we're at now.
 MethodLength:
   Max: 50
-    
+
 # Over time we'd like to get this down, but this is what we're at now.
 Metrics/AbcSize:
+  Max: 40
+
+Metrics/PerceivedComplexity:
   Max: 40
 
 # We use spaces, so it's less of a change to stick with that.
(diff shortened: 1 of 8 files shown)


cupakromer:
Thanks for taking care of this! I think we are going to simply disable those two remaining cops across the repos.

soulcutter:
I'm not a huge fan of assignments in conditions for readability reasons. It would be great to get 2.2 copping, though.

soulcutter:
Taking a look at them, I think we should correct both issues.

```ruby
# status, _ = Rack::Utils::SYMBOL_TO_STATUS_CODE.find do |_, c|
status, = Rack::Utils::SYMBOL_TO_STATUS_CODE.find do |_, c|
```

```ruby
# if default = Rake.application.instance_variable_get('@tasks')['default']
default = Rake.application.instance_variable_get('@tasks')['default']
if default
```

Is there a reason why we'd disable those cops, @cupakromer ?

samphippen:
I'm going to close this because it's stale and conflicting. @jamelablack If you still would like to see this merged, please bring it up to date, green, and then re-open. We'll discuss the changes happily once the PR is in that state :)

PR #1439, opened 2015-08-01
Bug fix: removed spaces in regex

jamelablack:


cupakromer:
Thanks for catching this!!! :blue_heart: :smile_cat: This was accidentally introduced in f6c277f8 during the prior Rubocop updates :grimacing: :sob:

Issue #868, opened 2013-12-08
Set `cache_key` correctly for Rails 4


alindeman:
It seems like we're doing something wrong with cache_key in view specs? https://github.com/rails/rails/pull/13208#issuecomment-29978002

Investigate and fix as required. Might be able to revert https://github.com/rspec/rspec-rails/pull/864
- [ ] 2.14
- [ ] 2.99
- [ ] 3.0

/cc: @JonRowe

PR #1452, opened 2015-09-02
Halve the time it takes to run view specs by enabling the resolver cache

zetter:
Hi,

Recently at FutureLearn we have taken some time to try and speed up our tests.

We have lots of view specs in our test suite, and we noticed that they took quite a while to run. This surprised us given that one of the benefits of testing views independently is speed. We found that one of the reasons for this slowness was that rspec-rails had disabled the ActionView resolver cache, increasing the time it took to load views.

There are two commits in this pull request. The first contains a fix the original issue which caused the cache to be disabled, the second enables the cache again. Please see their respective commit messages for more detail. When running our tests against this branch, it halves the time it takes us to run our view specs.

I'd like to hear peoples thoughts on this change, particularly:
+ Could converting the format to a symbol introduce breaking changes for anyone? The FutureLearn and rspec-rails test suites are fine, but I don't know if people are testing views in other ways.
+ Are there any other reasons why a user of rspec-rails might want ActionView's resolver cache to be disabled?

### DIFF
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -114,7 +114,7 @@ module RSpec
 
             render_options = { :template => template }
             render_options[:handlers] = [handler] if handler
-            render_options[:formats] = [format] if format
+            render_options[:formats] = [format.to_sym] if format
             render_options[:locales] = [locale] if locale
 
             render_options
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -155,9 +155,6 @@ module RSpec
             view.lookup_context.prefixes << _controller_path
           end
 
-          # fixes bug with differing formats
-          view.lookup_context.view_paths.each(&:clear_cache)
-
           controller.controller_path = _controller_path
           controller.request.path_parameters[:controller] = _controller_path
           controller.request.path_parameters[:action]     = _inferred_action unless _inferred_action =~ /^_/
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -121,7 +121,7 @@ module RSpec::Rails
           view_spec.render
 
           if ::Rails::VERSION::STRING >= '3.2'
-            expect(view_spec.received.first).to eq([{:template => "widgets/new", :locales=>['en'], :formats=>['html'], :handlers=>['erb']}, {}, nil])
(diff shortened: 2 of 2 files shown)


JonRowe:
> Could converting the format to a symbol introduce breaking changes for anyone?

Vaguely, this rings a bell, I believe the reason why we didn't just do this at the time is rails treats views with strings and symbols differently

zetter:
@JonRowe thanks for taking a look at this,

> I believe the reason why we didn't just do this at the time is rails treats views with strings and symbols differently

I couldn't see any reference to this in the [original commit](https://github.com/rspec/rspec-rails/pull/864) or [subsequent discussion](https://github.com/rspec/rspec-rails/issues/868) that suggests that there would be a problem marking the value of `format` a symbol. Can anyone provide an example of existing test that would behave unexpectedly after this change?

I have looked myself and could only think of an issue when a user of rspec-rails has registered a multiple custom mime types with a string rather than a symbol, e.g:

```ruby
Mime::Type.register "audio/mp3", "mp3"
Mime::Type.register "audio/mp4", "mp4"
```

This change would cause the resolver cache could mix templates in this case. It looks to me from the [`Mime::Type.register` API](http://api.rubyonrails.org/classes/Mime/Type.html#method-c-register) that the second argument is expected be a symbol (based upon examples and the parameter name) so I don't think we should worry about this case. This change also continues allow the options to `render` to be overridden and a stringified format to be provided if needed.

soulcutter:
Seems like this originated from 96d3e75916b7b1ed872368193c15537ec7ba53ef - if the cukes are passing now perhaps it's no longer necessary? Would this break view specs for people using some outdated version of rails that has the invalidation issue that @JonRowe referenced?

I don't know… but if they're upgrading their RSpec and staying on a legacy version of rails I think that they should probably use an older rspec-rails until they update their rails since it sounds like more of a bug on rails' side.

So I think I've talked myself into :+1: :shipit: on this issue.

zetter:
@soulcutter thanks for your time spent reviewing this and merging it in.

In case anyone else has a lot of view specs and wants to take advantage of this change now (but doesn't want the other changes on `master` or to use the pre-release rspec gems) I have back-ported these commits to `3.3.3` at https://github.com/Futurelearn/rspec-rails/tree/faster-view-specs-backport

You can use this changing the `rspec-rails` gem in your `Gemfile` to use this fork, e.g:

```ruby
gem 'rspec-rails', git: 'https://github.com/Futurelearn/rspec-rails.git', branch: 'faster-view-specs-backport'
```

PR #1457, opened 2015-09-18
Replace old Ruby hash rocket with new JSON style

mkyaw:
Replace old Ruby hash rocket (:foo => :bar) with new JSON style (foo: :bar)


JonRowe:
Sorry but as RSpec (including rspec-rails) proports to support 1.8.7 we should stick with 1.8.7 compatible documentation.

mkyaw:
Thanks for letting me know. I did not know about that.

PR #1458, opened 2015-09-18
Rails exclusion filters

soulcutter:
Rails in backtraces can get quite spammy, and the configuration option
to `filter_gems_from_backtrace` is not well-known since it is not found
in the generator's `rails_helper.rb`.

This adds `filter_rails_from_backtrace!` as a convenience method for
nixing rails gems in the backtrace and adds it to `rails_helper.rb` to
raise awareness.

Credit to @smcabrera for the idea and pairing

### DIFF
--- a/features/.nav
+++ b/features/.nav
@@ -6,6 +6,7 @@
 - Changelog.md
 - RailsVersions.md (Rails versions)
 - directory_structure.feature
+- backtrace_filtering.feature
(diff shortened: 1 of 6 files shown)


soulcutter:
Ok that's weird, locally rubocop preferred my version. Will fix and repush

JonRowe:
LGTM

PR #1459, opened 2015-09-18
Fix config ordering.

aceofspades:
use_fixtures setting must be declared before RSpec::Rails::FeatureExampleGroup

Already fixed on master via 33c3c738e150c34732333778b6508f52cc8dc1bd.

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -41,6 +41,13 @@ module RSpec
       config.backtrace_exclusion_patterns << /vendor\//
       config.backtrace_exclusion_patterns << %r{ lib/rspec/rails }
 
+      # fixture support
+      config.add_setting :use_transactional_fixtures, :alias_with => :use_transactional_examples
+      config.add_setting :use_instantiated_fixtures
+      config.add_setting :global_fixtures
+      config.add_setting :fixture_path
+      config.include RSpec::Rails::FixtureSupport, :use_fixtures
+
       config.include RSpec::Rails::ControllerExampleGroup, :type => :controller
       config.include RSpec::Rails::HelperExampleGroup,     :type => :helper
       config.include RSpec::Rails::ModelExampleGroup,      :type => :model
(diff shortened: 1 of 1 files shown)


soulcutter:
Can you further describe the problem? I'm not sure I understand.

`:use_fixtures` is used as a metadata tag filter so that `RSpec::Rails::FixtureSupport` will be included when an example group is tagged with `:use_fixtures => true` . I don't think of it as a setting, but we could just be using different terminology.

I see that `RSpec::Rails::FeatureExampleGroup` [includes RailsExampleGroup](https://github.com/rspec/rspec-rails/blob/3-3-maintenance/lib/rspec/rails/example/feature_example_group.rb#L6), which in turn [includes FixtureSupport](https://github.com/rspec/rspec-rails/blob/3-3-maintenance/lib/rspec/rails/example/rails_example_group.rb#L14) - so isn't it included for `:type => :feature` specs in any ordering?

aceofspades:
@soulcutter The problem I am seeing is that [fixture_support.rb](https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/fixture_support.rb) references `RSpec.configuration.fixture_path`, which bombs without `config.add_setting :fixture_path` being hit first.

soulcutter:
I can't reproduce this problem. I tried defining an example, then requiring rspec-rails so that it would attempt to apply the include on `RSpec.world.all_example_groups` before defining the setting, but no matter what I tried I never hit this problem. Can you supply an example of when this fails?

aceofspades:
Never mind then. I thought the reasoning was clear, I guess not. I'll use a local patch or master instead.

soulcutter:
@aceofbassgreg I'm sorry if my questions frustrated you into closing this PR, that was not my intention. I want to address this issue if it's something you're encountering in practice.

I believe I follow your reasoning, and have tried to come up with scenarios that would demonstrate the problem but have been unsuccessful. Having a reproducible test case would help to prevent regressions in this behavior in the future (even though it's currently fixed for you in `master`).

I appreciate the time you've taken to make this contribution, and I hope this conversation doesn't put a damper on you contributing in the future.

aceofspades:
@soulcutter No worries, to me it visually looked wrong but I trust your judgment.

aceofbassgreg:
Hi, no frustrations from me either ;-).

PR #1462, opened 2015-09-30
Update rails helper description for maintain_test_schema!

dcrec1:


lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -23,7 +23,7 @@
 # Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 
 <% if RSpec::Rails::FeatureCheck.can_maintain_test_schema? -%>
-# Checks for pending migrations before tests are run.
+# Updates test schema before tests are run.
```
JonRowe: If you want to change this I'd suggest `# Checks for pending migration and applies them before tests are run.` As the original text is currently clearer than your suggestion even if it's not the whole story.

dcrec1: Ok, will change it.

Thanks

dcrec1:
@JonRowe it's ok now?

JonRowe:
(thanks)

Issue #1334, opened 2015-03-05
Add support for `ActiveJob` based matchers


cupakromer:
Now that we officially support `ActiveJob` specs. It would be kind to the community to add a small set of matchers to work with it. There are some helpers on the Rails side: http://api.rubyonrails.org/classes/ActiveJob/TestHelper.html

My personal preference is for being explicit over some sort of magic. With that in mind, I am suggesting to only support the block based syntax for matchers. Based on the Rails matchers I tentatively suggest the following syntax:

```ruby
expect { :noop }.not_to have_enqueued_jobs

expect {
  HelloJob.perform_later('aaron')
  HelloJob.perform_later('rafael')
}.to have_enqueued_jobs(2)    # not the best phrasing

expect {
  MyJob.perform_later(1,2,3)
}.to enqueue_job(job: MyJob, args: [1,2,3], queue: 'low')
```

myronmarston:
Looks like someone already built a gem for this:

https://github.com/gocardless/rspec-activejob

Maybe that can be rolled into rspec-rails or at least serve as inspiration?

/cc @isaacseymour

isaacseymour:
Would be great to have rspec-activejob rolled in to rspec-rails. The ones I've written so far are:
```ruby
expect { FooJob.perform_later(model_instance, kw: 'arg') }.to enqueue_a(FooJob).
  with(global_id(model_instance), deserialize_as(kw: 'arg'))
```
(`global_id` can also be used on just a class, like `instance_of`, and the `deserialize_as` matcher is kinda poorly named, but deals with all the issues around hash de/serialization in 4.2.1).

There's also a non-block version of the `enqueue_a` matcher, which uses the same code under the hood:
```ruby
FooJob.perform_later(model_instance, kw: 'arg')

expect(FooJob).to have_been_enqueued.with(global_id(model_instance), deserialize_as(kw: 'arg'))
```

These have been enough for all the uses so far, but there's still quite a few features missing:
- expectations on the queue name
- expectations on time enqueued for (i.e. when using `wait: 10.minutes` or `wait_until: 10.days.from_now`)
- expectations on multiple jobs being enqueued at the same time within a block

I'd be happy to turn the gem into a PR onto rspec-rails when I have some time, if you think that would be valuable.

bf4:
(Before I saw this issue or #1155) I adapted a simple one from the rails active job test helpers, but unfortunately haven't figured out how to test rescuing and retries. cc @seuros

[...]

PR #1464, opened 2015-10-03
Add have_enqueued_jobs matcher

morgoth:
There is still 3 more matchers to be added `have_enqueued_with`, `have_performed_jobs` and `have_performed_with`, but I can work on them in another PR, unless you want all in one.

Anyway, I would like to get feedback on implementation and naming, as I'm not familiar with rspec codebase.

I didn't figure out why I don't have to clear enqueued jobs after/before each test, but the tests are passing so somehow this queue is cleared.
You can compare it with https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_helper.rb#L22-L23

Refs #1334

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -9,9 +9,9 @@ module RSpec
       module ActiveJob
         # @private
         class HaveEnqueuedJobs < RSpec::Matchers::BuiltIn::BaseMatcher
-          def initialize(number, only = nil)
-            @number = number
-            @only   = only
+          def initialize(only)
+            @only = only
+            set_expected_number(:exactly, 1)
           end
 
           def matches?(proc)
@@ -20,15 +20,47 @@ module RSpec
             before_block_jobs_size = enqueued_jobs_size(@only)
             proc.call
             @in_block_jobs_size = enqueued_jobs_size(@only) - before_block_jobs_size
-            @number == @in_block_jobs_size
+
+            case @expectation_type
+            when :exactly then @expected_number == @in_block_jobs_size
+            when :at_most then @expected_number >= @in_block_jobs_size
+            when :at_least then @expected_number <= @in_block_jobs_size
+            end
+          end
+
+          def exactly(count)
+            set_expected_number(:exactly, count)
+            self
+          end
+
+          def at_least(count)
+            set_expected_number(:at_least, count)
+            self
+          end
+
+          def at_most(count)
+            set_expected_number(:at_most, count)
+            self
+          end
+
+          def times
+            self
           end
 
           def failure_message
-            "expected to enqueue #{@number} jobs, but enqueued #{@in_block_jobs_size}"
+            "expected to enqueue #{message_expectation_modifier} #{@expected_number} jobs, but enqueued #{@in_block_jobs_size}"
           end
 
           def failure_message_when_negated
-            "expected not to enqueue #{@number} jobs, but enqueued #{@in_block_jobs_size}"
+            "expected not to enqueue #{message_expectation_modifier} #{@expected_number} jobs, but enqueued #{@in_block_jobs_size}"
+          end
+
+          def message_expectation_modifier
+            case @expectation_type
+            when :exactly then "exactly"
+            when :at_most then "at most"
+            when :at_least then "at least"
+            end
           end
 
           def supports_block_expectations?
@@ -38,11 +70,21 @@ module RSpec
         private
 
           def enqueued_jobs_size(only)
-            if only
-              queue_adapter.enqueued_jobs.count { |job| Array(only).include?(job.fetch(:job)) }
+            if only.any?
+              queue_adapter.enqueued_jobs.count { |job| only.include?(job.fetch(:job)) }
             else
               queue_adapter.enqueued_jobs.count
             end
+          end
+
+          def set_expected_number(relativity, count)
+            @expectation_type = relativity
+            @expected_number = case count
+                               when Numeric then count
+                               when :once then 1
+                               when :twice then 2
+                               when :thrice then 3
+                               end
           end
 
           def queue_adapter
@@ -52,20 +94,29 @@ module RSpec
       end
 
       # @api public
-      # Passess if `number` of jobs were enqueued inside block
+      # Passess if `count` of jobs were enqueued inside block
       #
       # @example
       #     expect {
       #       HeavyLiftingJob.perform_later
-      #     }.to have_enqueued_jobs(1)
+      #     }.to have_enqueued_jobs
       #
       #     expect {
       #       HelloJob.perform_later
       #       HeavyLiftingJob.perform_later
-      #     }.to have_enqueued_jobs(1, only: HelloJob)
-      def have_enqueued_jobs(number = 1, options = {})
-        only = options[:only]
-        ActiveJob::HaveEnqueuedJobs.new(number, only)
+      #     }.to have_enqueued_jobs(HelloJob).exactly(:once)
+      #
+      #     expect {
+      #       HelloJob.perform_later
+      #       HelloJob.perform_later
+      #       HelloJob.perform_later
+      #     }.to have_enqueued_jobs(HelloJob).at_least(2).times
+      #
+      #     expect {
+      #       HelloJob.perform_later
+      #     }.to have_enqueued_jobs(HelloJob).at_most(:twice)
+      def have_enqueued_jobs(*jobs)
+        ActiveJob::HaveEnqueuedJobs.new(jobs)
       end
     end
   end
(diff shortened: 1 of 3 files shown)


morgoth:
Specs are failing for older version of rails because there is no ActiveJob - what would be the best way to filter specs for older versions?
Also specs are failing because of `bin/yard stats --list-undoc` - where should I add docs?

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -0,0 +1,84 @@
+require "spec_helper"
+require "rspec/rails/matchers/active_job"
+
+RSpec.describe "active_job matchers" do
```
JonRowe: You can use the skip logic and the feature detection you've used above to skip these on non supported Rails...

e.g. `RSpec.describe "active_job matchers", :skip => !RSpec::Rails::FeatureCheck.has_active_job? do`

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,71 @@
...
+      module ActiveJob
+        # @private
+        class HaveEnqueuedJobs < RSpec::Matchers::BuiltIn::BaseMatcher
+          def initialize(number, only = nil)
+            @number = number
+            @only   = only
+          end
+
+          def matches?(proc)
+            return false unless Proc === proc # TODO: better to raise error?
```
JonRowe: Yes I think it'd be better to raise an error rather than silently failing, an argument error with a message would be best.

spec/rspec/rails/matchers/active_job_spec.rb
morgoth: Looks like it won't be so easy https://travis-ci.org/rspec/rspec-rails/jobs/83654996#L274
Also, I cannot simply `return` from file, because rspec throws error `/home/wojtek/Projects/rspec-rails/spec/rspec/rails/matchers/active_job_spec.rb: /home/wojtek/Projects/rspec-rails/spec/rspec/rails/matchers/active_job_spec.rb:3: Invalid return (SyntaxError)`
Any other ideas how to pass it?

JonRowe:
Ah of course it's still parsing the file so your defined classes are blowing up, you can either move these into let definitions (we'd typically do this with anonymous classes to avoid namespacing issues anyway)

e.g. 
```Ruby
let(:heavy_lifting_job) do
  class.new(ActiveJob::Base) do
    # ...
  end
end
```

And then referencing `heavy_lifting_job` rather than a constant.

Or you can wrap the entire inner of the describe in an if statement, the former option is preferred in my opinon.

morgoth:
@JonRowe Tests are passing and I squashed commits to one. Do you think it can be merged or should I work on other matchers on the same PR?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,72 @@
...
+      module ActiveJob
+        # @private
+        class HaveEnqueuedJobs < RSpec::Matchers::BuiltIn::BaseMatcher
+          def initialize(number, only = nil)
+            @number = number
+            @only   = only
+          end
+
+          def matches?(proc)
+            raise ArgumentError, "Proc have to be passed to expectation" unless Proc === proc
```
JonRowe: A better wording for this would be "have_enqueued_jobs only supports block expectations"

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,72 @@
...
+            @in_block_jobs_size = enqueued_jobs_size(@only) - before_block_jobs_size
+            @number == @in_block_jobs_size
+          end
+
+          def failure_message
+            "expected to enqueue #{@number} jobs, but enqueued #{@in_block_jobs_size}"
+          end
+
+          def failure_message_when_negated
+            "expected not to enqueue jobs, but enqueued #{@in_block_jobs_size}"
```
JonRowe: This should mention `@number` too

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,72 @@
...
+      #     expect {
+      #       HeavyLiftingJob.perform_later
+      #     }.to have_enqueued_jobs(1)
+      #
+      #     expect {
+      #       HelloJob.perform_later
+      #       HeavyLiftingJob.perform_later
+      #     }.to have_enqueued_jobs(1, only: HelloJob)
+      def have_enqueued_jobs(number = 1, options = {})
+        only = options[:only]
```
JonRowe: This is rather a strange way to do this, I think it'd be better to have this more like the `receive` matchers, So you'd have:
```Ruby
have_enqueued_jobs(HelloJob)
have_enqueued_jobs(HelloJob).at_least(2).times
have_enqueued_jobs(HelloJob).at_most(5).times
have_enqueued_jobs(HelloJob).exactly(10).times
```
etc (`.times` would be a no op method for sugar, and we should probably accept `:once`, `:twice` etc)

JonRowe: the job type would still be optional to allow any job...

morgoth: I'm looking for some examples, but with no luck. Should I implement all this methods by hand or is there some "macro" that have it unified already?

morgoth: The matcher name is `have_enqueued_jobs` (plural) - should it be also aliased to singular form?

morgoth:
I changed matcher usage to proposed - it's much more complicated now and more error prone ;)

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,123 @@
...
+              queue_adapter.enqueued_jobs.count { |job| only.include?(job.fetch(:job)) }
+            else
+              queue_adapter.enqueued_jobs.count
+            end
+          end
+
+          def set_expected_number(relativity, count)
+            @expectation_type = relativity
+            @expected_number = case count
+                               when Numeric then count
```
JonRowe: This case should probably be handled by an `else` and `Integer(count)`

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -0,0 +1,123 @@
...
+      #     expect {
+      #       HelloJob.perform_later
+      #       HelloJob.perform_later
+      #       HelloJob.perform_later
+      #     }.to have_enqueued_jobs(HelloJob).at_least(2).times
+      #
+      #     expect {
+      #       HelloJob.perform_later
+      #     }.to have_enqueued_jobs(HelloJob).at_most(:twice)
+      def have_enqueued_jobs(*jobs)
```
JonRowe: I think this should be singular and only take one job, as it's ambiguous when specifying multiple jobs and their count.

JonRowe: It'd be fine with a "multiple" version that took no specific job though

lib/rspec/rails/matchers/active_job.rb
morgoth: I took it from https://github.com/rspec/rspec-mocks/blob/b840503a4f1d309c24f675f6f4502828cbbd1431/lib/rspec/mocks/message_expectation.rb#L576-L581 - should there be a mismatch in behaviour?

lib/rspec/rails/matchers/active_job.rb
morgoth: Rails mini test helpers allow to pass multiple job names - if we limit it to only one, we don't provide full functionality that is available in Rails.

JonRowe: Rails mini test helpers don't place a focus on readability and lack of ambiguity, we do, and we do still have the full functionality as you can have multiple expectations. `have_enqueued_job(job_a).and have_enqueued_job(job.b)`

morgoth:
@JonRowe I updated it to singular version. I don't know what kind of black magic rspec uses with `and` method, but somehow it works ;-)

myronmarston:
> @JonRowe I updated it to singular version. I don't know what kind of black magic rspec uses with and method, but somehow it works ;-)

No black magic.  `and` just returns a new compound matchers that has a reference to both of the other matchers and ANDs them together.

morgoth:
@JonRowe Thanks. I'll be unavailable in the next week, but when I come back I'll work on improving this matcher. 
At the beginning I was thinking about separate one (like in rails), where you can match on passed args https://github.com/rails/rails/blob/master/activejob/lib/active_job/test_helper.rb#L234-L245
but here maybe it will be achievable using the same matcher and new methods on it `with_args`, `with_queue` and `at`

samnang:
Thanks for the great pull request :+1: 

I'm not pretty sure, should we passing expectation inline instead of passing block:

```ruby
user_params = { ... }

user = CreateUser.call(user_params)

expect(SubscribeUserMailingList).to have_enqueued_job.with(user)

# instead of

user_params = { ... }
user = nil

expect {
  user = CreateUser.call(user_params)
}.to  have_enqueued_job(SubscribeUserMailingList).with(user)
```

We want to get rid of temporary variables when we have assert that depends on the result in the expect block.

PR #1469, opened 2015-10-08
add load hook

masarakki:
I want loaded hook to configure in my gem after default `rspec-rails` configuration.

For example, [devise.gem](https://github.com/plataformatec/devise#test-helpers) can write in `devise/railtie.rb` like this

```ruby
  ActiveSupport.on_load :rspec_rails do
    RSpec.configuration.include Devise::TestHelpers, type: :controller
  end
```

### DIFF
--- a/lib/rspec/rails.rb
+++ b/lib/rspec/rails.rb
@@ -14,3 +14,5 @@ require 'rspec/rails/vendor/capybara'
 require 'rspec/rails/configuration'
 require 'rspec/rails/active_record'
 require 'rspec/rails/feature_check'
+
+ActiveSupport.run_load_hooks :rspec_rails, RSpec::Rails


JonRowe:
Sorry, can you explain the purpose of this?

masarakki:
a lot of gem provides helpers for rspec,  for example [devise](https://github.com/plataformatec/devise#test-helpers) and [factory_girl](http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Configure_your_test_suite),
but we are requested to to write `config.include Devise::TestHelper` in `spec/rails_helper` by myself and I feel it lazy.
Now I'm creating a new gem that has helpers using rspec-rails's methods, I want to include the helpers automatically.

At first, I included like 

```ruby
module MyNewGem
  class Railtie < ::Rails::Railtie
    config.after_initialize do
       RSpec.configuration.include MyNewGem::TestHelper
    end
  end
end
```

but it doesn't work because rspec-rails's methods are included by `require 'rspec/rails'`.
In this case, `@items_and_filters` are  
```ruby 
  @items_and_filters=
    [[MyNewGem::TestHelper, {}],
     [RSpec::Rails::FixtureSupport, {:use_fixtures=>true}],
     [RSpec::Rails::FixtureSupport, {}],
     ...
```

I want to include my helpers after `require 'rspec/rails'`, but `require 'rspec/rails'` is in `spec/rails_helper.rb` and I have no time to do.
so I want hook after rspec_rails was ready.

myronmarston:
This seems useful but with the split of `spec_helper` and `rails_helper` in RSpec 3 we're encouraging users to write rails-less specs where it makes sense to do so, as you can get really fast pass/fail feedback from RSpec when you don't need to boot rails.  I worry that this would couple 3rd party gems to only be usable when rails is booted, which could get in the way of being able to also use the gem w/o rails loaded.

soulcutter:
I see the appeal of this since it would remove a step for users to gain functionality from a gem they're using.

I also feel configuration transparency has been a goal in RSpec, and this change hides what modules get included where, and doesn't give users a chance to opt out of that behavior.

I don't worry that it couples 3rd party gems to only be usable with rails, since modules could still be included manually (as they are now) wherever a person wants, but I do think this could be confusing. For example, I have used factory_girl in a rails project, with non-activerecord objects - it would seem strange to need to include FactoryGirl methods in one of those specs, but not in another because the other one has loaded rails.

Also the cost of change here is that gems would have to update to use the new hook (not a big deal), but then they also must either arbitrarily require a recent rspec-rails that runs the hook or give instructions on how to use it with an older rspec, which would be a documentation burden that projects would have to carry for a long time.

These are some of the things that came to mind looking at this. Interesting proposal.

JonRowe:
3rd party gems are able to configure themselves by configuring RSpec directly, it's just the require has to be in our spec_helper. I think thats enough really.

soulcutter:
The consensus is that the current behavior has advantages over this automatic hook, and it's not too onerous to have a line in the configuration.

Thanks for this contribution, however I am closing this without merging. Keep the ideas coming!

PR #1470, opened 2015-10-11
Bump Rails version check to 3.2.22

tancnle:


lib/rspec/rails/adapters.rb
```diff
@@ -38,8 +38,7 @@ def self.disable_testunit_autorun
         gem 'minitest' if defined?(Kernel.gem)
         require 'minitest/unit'
         Assertions = MiniTest::Assertions
-      elsif ::Rails::VERSION::STRING >= '3.2.21'
-        # TODO: Change the above check to >= '3.2.22' once it's released
+      elsif ::Rails::VERSION::STRING >= '3.2.22'
```
JonRowe: @cupakromer can you remember why is necessary? When is rails going to be higher than this but lower than 4.0.0?

soulcutter:
I share @JonRowe 's curiosity about 'why', but since this clearly satisfies a TODO, I am tempted to merge without fully understanding. What say you?

JonRowe:
I'd actually rather remove it altogether as I think it's from when Rails 4 wasn't released...

soulcutter:
That makes a lot of sense. I'd feel better about that than blindly merging away.

lib/rspec/rails/adapters.rb
cupakromer: This was necessary for support for the newer Ruby 2.2 syntax. Officially I don't believe Rails 3.2.21 was supporting that just yet. There were some code and gem level changes they needed to do. So I think this was here just as a reminder about that condition and if we wanted to match 1:1 with the Rails support we would have to update the version string.

If you look at the Travis config for Rails [3.2.1](https://github.com/rails/rails/blob/v3.2.21/.travis.yml) and [3.2.2](https://github.com/rails/rails/blob/v3.2.22/.travis.yml) you can see in 3.2.2 they added Ruby 2.1 and 2.2.

cupakromer:
Based on the Rails support and changes for Ruby 2.2 I'm going to merge this so we keep up with their side.

JonRowe:
@cupakromer did you mean to close this or merge it?

PR #1476, opened 2015-10-15
Fix warning in generate example of Getting Started

### LINKED ISSUE #1475: Fix warning from generate command in documentation
The documentation here https://relishapp.com/rspec/rspec-rails/v/3-3/docs/gettingstarted mentions the command `$ rails generate scaffold Widgets name:string` but running that locally gives the warning:

> [WARNING] The model name 'Widgets' was recognized as a plural, using the singular 'Widget' instead. Override with --force-plural or setup custom inflection rules for this noun before running the generator.

Change documentation's example to use `$ rails generate scaffold Widget name:string` as the example to avoid this warning.

marr:
Fixes #1475


JonRowe:
Thanks

PR #1479, opened 2015-10-20
Label Rspec as @private for docs, update other modules visibility

JonRowe:
This should remove the broken documentation links, which it does but it also labels everything under `RSpec` as private :/ This is for rspec/rspec.github.io#50...
screen shot 2015-10-20 at 21 52 04
screen shot 2015-10-20 at 21 52 10

### DIFF
--- a/lib/generators/rspec.rb
+++ b/lib/generators/rspec.rb
@@ -1,6 +1,7 @@
 require 'rails/generators/named_base'
 require 'rspec/rails/feature_check'
 
+# @private
 # Weirdly named generators namespace (should be `RSpec`) for compatability with
 # rails loading.
 module Rspec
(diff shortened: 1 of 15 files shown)


JonRowe:
See lsegal/yard#906

JonRowe:
Ok, so it turns out YARD wasn't responsible for this entirely, our modules are actually mislabeled on master so we can just hide lowercase `Rspec` and it's ok, I went through all the docs and updated the visibility correctly. Needs a review @myronmarston @soulcutter @cupakromer

myronmarston:
If this fixes the doc issue, LGTM.

JonRowe:
@myronmarston it does for me with `yard --server`

myronmarston:
> @myronmarston it does for me with yard --server

Have you tried generating the docs for rspec.info, though?  Would be nice to see it work for that since that's a different process.

JonRowe:
Am I ok to merge this? I'd like to cherry pick it back to 3-3 so I can test doc regeneration

myronmarston:
Merge away.

PR #1480, opened 2015-10-22
Add more methods to have_enqueued_job matcher

morgoth:
This is a follow up to https://github.com/rspec/rspec-rails/pull/1464

### DIFF
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -15,13 +15,38 @@ RSpec.describe "ActiveJob matchers", :skip => !RSpec::Rails::FeatureCheck.has_ac
 
   let(:hello_job) do
     Class.new(ActiveJob::Base) do
-      def perform; end
+      def perform(*)
+      end
     end
   end
 
   let(:logging_job) do
     Class.new(ActiveJob::Base) do
       def perform; end
+    end
+  end
+
+  let(:global_id_model) do
+    Class.new do
+      include GlobalID::Identification
+
+      attr_reader :id
+
+      def self.find(id)
+        new(id)
+      end
+
+      def self.name
+        "AnonymousClass"
+      end
+
+      def initialize(id)
+        @id = id
+      end
+
+      def to_global_id(options = {})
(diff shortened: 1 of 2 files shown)


spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -144,5 +169,41 @@ def perform; end
...
+
+    it "passes with provided queue name" do
+      expect {
+        hello_job.set(:queue => "low").perform_later
+      }.to have_enqueued_job.on_queue("low")
+    end
+
+    it "passes with provided at date" do
+      expect {
+        hello_job.set(:wait_until => Date.tomorrow.noon).perform_later
```
soulcutter: Can you extract `Date.tomorrow.noon` into a variable? I know the window is small, but it's possible for this to fail when run precisely at midnight when the first call is before the date change and the second call is after.

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -144,5 +169,41 @@ def perform; end
...
+    it "passes with provided arguments containing global id object" do
+      global_id_object = global_id_model.new(42)
+
+      expect {
+        hello_job.perform_later(global_id_object)
+      }.to have_enqueued_job.with(global_id_object)
+    end
+
+    it "generates failure message with all provided options" do
+      message = "expected to enqueue exactly 2 jobs, with [42], on queue low, at #{Date.tomorrow.noon}, but enqueued 0"
```
soulcutter: Extract the relative date into a variable here as well

soulcutter:
This looks pretty good - would you be interested in adding cukes as well for documentation?

spec/rspec/rails/matchers/active_job_spec.rb
morgoth: @soulcutter Done

morgoth:
@soulcutter Yes, I was thinking about adding cukes already. I don't have experience with it so it may take some time. Should I try to add it in this PR or separate one is fine?

soulcutter:
A separate PR is fine

soulcutter:
Thanks for these enhancements! This is really coming together :+1:

PR #1489, opened 2015-11-12
Throw descriptive error when have_enqueued_job matcher is used with non test adapter

morgoth:
Otherwsie error is thrown:
```
NoMethodError:
  undefined method `enqueued_jobs' for ActiveJob::QueueAdapters::InlineAdapter:Class
```

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -158,6 +158,9 @@ module RSpec
       #       HelloJob.set(wait_until: Date.tomorrow.noon, queue: "low").perform_later(42)
       #     }.to have_enqueued_job.with(42).on_queue("low").at(Date.tomorrow.noon)
       def have_enqueued_job(job = nil)
+        unless ::ActiveJob::QueueAdapters::TestAdapter === ::ActiveJob::Base.queue_adapter
+          raise StandardError, "To use have_enqueued_job matcher set `ActiveJob::Base.queue_adapter = :test`"
+        end
         ActiveJob::HaveEnqueuedJob.new(job)
       end
     end
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -207,5 +207,16 @@ RSpec.describe "ActiveJob matchers", :skip => !RSpec::Rails::FeatureCheck.has_ac
         }.to have_enqueued_job(hello_job).with(42).on_queue("low").at(date).exactly(2).times
       }.to raise_error(message)
     end
+
+    it "throws descriptive error when no test adapter set" do
+      queue_adapter = ActiveJob::Base.queue_adapter
+      ActiveJob::Base.queue_adapter = :inline
+
+      expect {
+        expect { heavy_lifting_job.perform_later }.to have_enqueued_job
+      }.to raise_error("To use have_enqueued_job matcher set `ActiveJob::Base.queue_adapter = :test`")
+
+      ActiveJob::Base.queue_adapter = queue_adapter
+    end
   end
 end


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -158,6 +158,9 @@ def queue_adapter
       #       HelloJob.set(wait_until: Date.tomorrow.noon, queue: "low").perform_later(42)
       #     }.to have_enqueued_job.with(42).on_queue("low").at(Date.tomorrow.noon)
       def have_enqueued_job(job = nil)
+        if ::ActiveJob::Base.queue_adapter.class.name != "ActiveJob::QueueAdapters::TestAdapter"
+          raise StandardError, "To use have_enqueued_job matcher set ActiveJob::Base.queue_adapter to :test"
```
soulcutter: Minor nitpick - would rather it say `set ActiveJob::Base.queue_adapter = :test` so that it's easy to copy and paste.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -158,6 +158,9 @@ def queue_adapter
       #       HelloJob.set(wait_until: Date.tomorrow.noon, queue: "low").perform_later(42)
       #     }.to have_enqueued_job.with(42).on_queue("low").at(Date.tomorrow.noon)
       def have_enqueued_job(job = nil)
+        if ::ActiveJob::Base.queue_adapter.class.name != "ActiveJob::QueueAdapters::TestAdapter"
```
soulcutter: I think it would make Demeter happier if this were

```ruby
! ::ActiveJob::Base.queue_adapter.instance_of? ActiveJob::QueueAdapters::TestAdapter
```

If Rails changes that constant name, we'll hear about it this way also.

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -207,5 +207,13 @@ def to_global_id(options = {})
         }.to have_enqueued_job(hello_job).with(42).on_queue("low").at(date).exactly(2).times
       }.to raise_error(message)
     end
+
+    it "throws descriptive error when no test adapter set" do
+      ActiveJob::Base.queue_adapter = :inline
```
soulcutter: If this is the last test in this file to get run, we are leaving the queue_adapter as `:inline`. I'm not sure it matters, but I'm always a bit wary of polluting the global environment

lib/rspec/rails/matchers/active_job.rb
JonRowe: I think it'd make me and @myronmarston happier if this was
`if ActiveJob::QueueAdapters::TestAdapter === ActiveJob::Base.queue_adapter`

lib/rspec/rails/matchers/active_job.rb
JonRowe: With backticks to show it's code :)

spec/rspec/rails/matchers/active_job_spec.rb
JonRowe: And as we run tests randomly you can't guarantee this, we should reset it back to it's original setting (not :test specifically).

lib/rspec/rails/matchers/active_job.rb
soulcutter: even better :+1:

morgoth:
Updated

spec/rspec/rails/matchers/active_job_spec.rb
morgoth: normally I would do this with `ensure`, but rspec syntax doesn't allow it and not sure if wrapping whole code in `begin; end` block is fine.
Anyway I changed it

JonRowe: Ruby syntax doesn't allow it, nothing to do with us ;)

morgoth: not really:
```ruby
def test
ensure
  puts "test"
end
test #=> "test"
```

PR #1496, opened 2015-11-29
Fix route helpers collision with asset helpers in view spec

sikachu:
This fixes the problem when calling a route helper method that has the same name as an asset helper method, such as `video_path` or `image_path`. This is done by re-including `Rails.application.routes.url_helpers` in the context of view specs.

Note that the implementation to include `url_helpers` has been taken from `lib/rspec/rails/example/feature_example_group.rb`.

---

The bug I found was that when I tried to do `video_path(@video)`, I would get `"/videos/#<Video:0x007ff92ad2aa48>"` returned instead of `"/videos/42"`. This is because the [asset helper](http://api.rubyonrails.org/classes/ActionView/Helpers/AssetUrlHelper.html#method-i-video_path) method was being called instead of routes helper.

By the way, I spent some time trying to write the failing unit spec instead of integration spec (using Cucumber) but couldn't get the example within the example to run. Please let me know if I should try again to rewrite it to be a unit spec, or if this is already okay.

### DIFF
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -31,6 +31,16 @@ module RSpec
 
       # DSL exposed to view specs.
       module ExampleMethods
+        extend ActiveSupport::Concern
+
+        included do
+          include ::Rails.application.routes.url_helpers
+
+          if ::Rails.application.routes.respond_to?(:mounted_helpers)
+            include ::Rails.application.routes.mounted_helpers
+          end
+        end
+
         # @overload render
         # @overload render({:partial => path_to_file})
         # @overload render({:partial => path_to_file}, {... locals ...})
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -90,6 +90,27 @@ module RSpec::Rails
       end
     end
 
+    describe "routes helpers collides with asset helpers" do
+      let(:view_spec) do
+        Class.new do
+          include ActionView::Helpers::AssetTagHelper
+          include ViewExampleGroup::ExampleMethods
+        end.new
+      end
+
+      let(:test_routes) do
+        ActionDispatch::Routing::RouteSet.new.tap do |routes|
+          routes.draw { resources :images }
+        end
+      end
+
+      it "uses routes helpers" do
+        allow(::Rails.application).to receive(:routes).and_return(test_routes)
+        expect(view_spec.image_path(double(:to_param => "42"))).
+          to eq "/images/42"
+      end
+    end
+
     describe "#render" do
       let(:view_spec) do
         Class.new do


lib/rspec/rails/example/view_example_group.rb
```diff
@@ -149,6 +149,22 @@ def _include_controller_helpers
 
         helper(*_default_helpers)
 
+        app = ::Rails.application
+        if app.respond_to?(:routes)
```
JonRowe: Is there a version of rails that doesn't respond to routes? We only do defensive programming like this if theres a version inconsistency, and even then it's often better to make it a specific to the version of Rails, and often better to conditionally define a method e.g.

```Ruby
included do
  ...
  setup_view_helpers
  ...
end

if ::Rails::VERSION < n
  def setup_view_helpers
    ...
  end
else
  def setup_view_helpers
    ...
  end
end
```

And this pretty much goes for all the if statements in this PR

features/view_specs/view_spec.feature
```diff
@@ -212,3 +212,24 @@ Feature: view spec
       """
     When I run `rspec spec/views`
     Then the examples should all pass
+
+  Scenario: path helpers should take precedence over asset helper
```
JonRowe: It would be better to write this as a unit spec as it's not applicable to our documentation really, also is this a breaking change in our behaviour? Some people may be relying on inverse behaviour already...

sikachu: > It would be better to write this as a unit spec as it's not applicable to our documentation really

I'd love some help to get this rewritten as unit test as well. I tried to do something like this in the spec and it doesn't work ...

```ruby
# in spec/rspec/rails/example/view_example_group_spec.rb
    describe "routes helpers collides with aset helpers" do
      it "prefers routes helper" do
        group = RSpec::Core::ExampleGroup.describe 'show.html.erb' do
          include ViewExampleGroup

          it "passes" do
            ### this never gets called
            puts "TESTING"
          end
        end

        group.run NullObject.new
      end
    end
```

Am I doing this right?

sikachu: > also is this a breaking change in our behaviour? Some people may be relying on inverse behaviour already...

I doubt it. This breaks the view test completely, as when you call `video_path` in the _view_ you get `/videos/42` (i.e. it correctly calls routes helper). This collision only occur in the view spec.

JonRowe: Yes but consider the case where someone is testing that an output matches an asset path

```
it "displays my video" do
  expect(view).to include video_path('my_asset')
end
```

lib/rspec/rails/example/view_example_group.rb
sikachu: To be honest, I was initially going with just doing a straight `Rails.applicaiton.routes.url_helpers`. However, then I saw [this code](https://github.com/rspec/rspec-rails/blob/b4c00c45cfeb4ba5be3aeef5fd2b4205f77c0a47/lib/rspec/rails/example/feature_example_group.rb#L13-L21) in `feature_example_group.rb` is doing this defensive coding, so I just copied and pasted it. I saw it was introduced in https://github.com/rspec/rspec-rails/commit/0fda98b77b43d785e095b6f525840f7d233d9a1d#diff-07cac43fac355c3ef9dac2caf3a08574R11 by @jnicklas and @alindeman, so maybe they will have some insight if it's necessary.

JonRowe: Try it without it, if the build passes its unnecessary.

features/view_specs/view_spec.feature
JonRowe: Take a look at how we check `params` / `template` etc is tested, but if you absolutely need to run a spec to test this then you need to check the `run_count` and check for failures, you can't rely on output it'll be swallowed by the null object.

sikachu: > Yes but consider the case where someone is testing that an output matches an asset path

In that case, that said someone would already have to use [`path_to_video`](http://api.rubyonrails.org/classes/ActionView/Helpers/AssetUrlHelper.html#method-i-path_to_video) method in their view becasue if they're calling `video_path` in their view file (while also having a route name `video`) it'll actually call the route helper instead of the asset helper. Otherwise, they are probably testing the wrong thing.

So, to recap: with `resources :videos` in your route file

* Behavior before this patch:
    * `video_path` in view -> route helper
    * `video_path` in view spec -> asset helper
* Behavior after this patch:
    * `video_path` in view -> route helper
    * `video_path` in view spec -> route helper

--- 

> if you absolutely need to run a spec to test this then you need to check the run_count and check for failures

Aha! So that's what it was for. I'll rewrite my test then.

samphippen:
@sikachu are you still interested in finishing up this PR? We'd love to see this merged.

sikachu:
@samphippen I will work on this this Friday. Sorry for the delay.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -90,6 +90,21 @@ module ::ApplicationHelper; end
       end
     end
 
+    describe "routes helpers collides with aset helpers" do
+      it 'delegates to the controller' do
+        with_isolated_config do
+          ::Rails.application.routes.draw { resources :videos }
```
sikachu: I was playing around with this, tried to move this to `before` block to clean it up and spec just failed. I can look into this to make sure it was set back to whatever it was.

sikachu:
@samphippen I updated the code and test. This is ready for another look.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -90,6 +90,27 @@ module ::ApplicationHelper; end
       end
     end
 
+    describe "routes helpers collides with asset helpers" do
+      it "uses routes helpers" do
+        with_isolated_config do
+          ::Rails.application.routes.draw { resources :images }
+          output = nil
+
+          RSpec::Core::ExampleGroup.describe "a view spec", :type => :view do
```
JonRowe: Rather than this, try to write this spec as below, where you create a class and include `ViewExampleGroup::ExampleMethods`.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -90,6 +90,27 @@ module ::ApplicationHelper; end
...
+            before do
+              controller.assets_dir = "public"
+              allow(controller).to receive(:_routes).
+                and_return(::Rails.application.routes)
+            end
+
+            specify { output = image_path(double(:to_param => "42")) }
```
JonRowe: Rather than running a spec within a spec you can then instantiate the class from above and assert this directly

sikachu:
@JonRowe I've updated PR per your feedback. Do you mind give this another look?

JonRowe:
Thanks :)

PR #1499, opened 2015-12-06
Fix helper specs with internal classes

samphippen:
A rebase of #1340. Got a few more things to fix up.

### DIFF
--- a/lib/rspec/rails/example/helper_example_group.rb
+++ b/lib/rspec/rails/example/helper_example_group.rb
@@ -12,8 +12,8 @@ module RSpec
 
       # @private
       module ClassMethods
-        def determine_default_helper_class(_ignore)
-          described_class
+        def determine_constant_from_test_name(_ignore)
+          described_class if yield(described_class)
         end
       end
 
--- a/lib/rspec/rails/example/helper_example_group.rb
+++ b/lib/rspec/rails/example/helper_example_group.rb
@@ -12,8 +12,15 @@ module RSpec
 
       # @private
       module ClassMethods
-        def determine_constant_from_test_name(_ignore)
-          described_class if yield(described_class)
+        if ::Rails::VERSION::MAJOR > 3
+          def determine_constant_from_test_name(_ignore)
+            described_class if yield(described_class)
+          end
+        else
+          def determine_default_helper_class(_ignore)
+            return unless Module === described_class && !(Class === described_class)
+            described_class
+          end
         end
       end
 
--- a/spec/rspec/rails/example/helper_example_group_spec.rb
+++ b/spec/rspec/rails/example/helper_example_group_spec.rb
@@ -2,7 +2,11 @@ require "spec_helper"
 
 module RSpec::Rails
   describe HelperExampleGroup do
-    module ::FoosHelper; end
+    module ::FoosHelper
+      class InternalClass
+      end
+    end
+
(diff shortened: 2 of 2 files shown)


JonRowe:
LGTM, whats still outstanding?

samphippen:
There's this from @cupakromer https://github.com/rspec/rspec-rails/pull/1340#issuecomment-126691006 but maybe we already fixed it? @cupakromer can you weigh in here?

lib/rspec/rails/example/helper_example_group.rb
```diff
@@ -12,8 +12,15 @@ module HelperExampleGroup
...
       module ClassMethods
-        def determine_default_helper_class(_ignore)
-          described_class
+        if ::Rails::VERSION::MAJOR > 3
+          def determine_constant_from_test_name(_ignore)
+            described_class if yield(described_class)
+          end
+        else
+          def determine_default_helper_class(_ignore)
+            return unless Module === described_class && !(Class === described_class)
```
samphippen: I want to rewrite this boolean expression to read with less complexity.

JonRowe: I'm still not groking why theres two different names for these methods

samphippen: these have to be rails hooks

cupakromer:
Ok, I remember what I was thinking before. It was more a question of does this make conceptual / semantic sense, not that I had a problem with the implementation.

Essentially, the nested group _will_ have `:type => :helper` - with all of the helper stuff mixed into it - even though it is not specing a helper. I think before this concerned me as it will bloat and slow the nested group unnecessarily. It also may be misleading for tools / plugins that look for `:type => :helper`.

I did some basic sanity checking tonight and I don't have any real problems with this. My initial concern about the metadata and what goes along with it still stands but it's not something I feel is a blocker. There is one corner case that may be confusing to people:

```ruby
# I ran this using Rails 4.2.5 and rspec-rails 3.4.0
require "rails_helper"

module FooHelper
  def foo_helper
    :foo
  end

  class InnerClass
  end
end

RSpec.describe FooHelper, type: :helper do
  def self.determine_default_helper_class(_ignore)
    return unless Module === described_class && !(Class === described_class)
    described_class
  end

  specify "#foo_helper returns :foo" do
    expect(helper.foo_helper).to be(:foo)
  end

  describe FooHelper::InnerClass do
    it "it has a helper" do
      expect(helper).to be
    end

    it "the helper does not have behavior like the parent group's" do
      expect {
        helper.foo_helper
      }.to raise_error(NoMethodError)
    end
  end
end
```

samphippen:
@cupakromer I think that is a an edge case that we're unlikely to encounter, given that this is already a fix for an edge case. @cupakromer @JonRowe is this good to merge?

cupakromer:
Yeah good to merge. My concern isn't a blocker. Do you think there's a way to leverage [`Configuration#define_derived_metadata`](http://rspec.info/documentation/3.4/rspec-core/RSpec/Core/Configuration.html#define_derived_metadata-instance_method)? Or would that potentially change metadata semantics too much to be even more confusing?

samphippen:
@cupakromer I think changing metadata semantics in a minor version is a bad idea. I think we should revisit when we start work on 4.0 proper.

PR #1501, opened 2015-12-08
Warn if a fixture method is called from a `before(:context)` block.

### LINKED ISSUE #1442: Warn on context fixture helper
Fixtures are controlled by the underlying `ActiveRecord::TestFixtures` code. Which sets up fixtures before **each** example. While the database _may_ have fixture data in it before this point, that data will be reset once the test fixture code runs.

When the `fixtures` message is sent in an example group, or global fixtures are defined, the Active Record test fixtures register the accessors in an anonymous module. That module is included in the example group's ancestors.

Which means the accessor helpers are available in any example group instance; thus they are available in `before(:context)` and `after(:context)`. However, the internal fixture setup and cache has not been performed yet. Attempting to call a fixture helper prior to the fixtures being setup, raises an undefined method on `NilClass` error.

Related: #1201 and #1208

samphippen:
Fixes #1442.

The basic approach here is to capture the addition of the fixture
methods to the example group instance and then monkeypatch them. The
monkeypatch checks to see if we're currently in a `before(:context)` hook
and if we are then it prints a warning and doesn't invoke the method.

The warning here is a little sparse at the moment, and I'd like to make
it more clear. One thing that's a little gross about this implementation
is that it uses the inspect string of the example group to determine if
we're in a `before(:context)` hook. As far as I can tell there isn't a
better way to make that determination, but maybe someone's got a clever
trick.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -28,6 +28,28 @@ module RSpec
             self.use_transactional_fixtures = RSpec.configuration.use_transactional_fixtures
           end
           self.use_instantiated_fixtures  = RSpec.configuration.use_instantiated_fixtures
+
+          def self.fixtures(*args)
+            orig_methods = private_instance_methods
+            super.tap do
+              new_methods = private_instance_methods - orig_methods
+              new_methods.each do |method_name|
+                proxy_method_warning_if_called_in_before_context_scope(method_name)
+              end
+            end
+          end
+
+          def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
+            orig_implementation = instance_method(method_name)
+            define_method(method_name) do |*args, &blk|
+              if inspect.include?("before(:context)")
+                RSpec.warn_with("Calling fixture method in before :context ")
+              else
+                orig_implementation.bind(self).call(*args, &blk)
+              end
+            end
+          end
+
           fixtures RSpec.configuration.global_fixtures if RSpec.configuration.global_fixtures
         end
       end
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,8 @@ Bug fixes:
 
 * Make it possible to write nested specs within helper specs on classes that are
   internal to helper classes. (Sam Phippen, Peter Swan, #1499).
+* Warn if a fixture method is called from a `before(:context)` block, instead of
+  crashing with a `undefined method for nil:NilClass`. (Sam Phippen, #1501)
 
 ### 3.4.0 / 2015-11-11
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.3.3...v3.4.0)
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -14,6 +14,7 @@ module ExampleAppHooks
 
     def final_tasks
       copy_file 'spec/verify_active_record_spec.rb'
+      copy_file 'spec/verify_fixture_warning_spec.rb'
       run('bin/rake db:migrate')
       if ::Rails::VERSION::STRING.to_f < 4.1
         run('bin/rake db:migrate RAILS_ENV=test')
--- a/example_app_generator/spec/verify_fixture_warning_spec.rb
+++ b/example_app_generator/spec/verify_fixture_warning_spec.rb
@@ -1,0 +1,64 @@
+require 'rails_helper'
+
+RSpec.describe "Fixture warnings" do
+  def generate_fixture_example_group(hook_type)
+    RSpec.describe do
+      include RSpec::Rails::RailsExampleGroup
+      fixtures :things
+
+      before(hook_type) do
+        things :a
+      end
+
+      it "" do
+
+      end
+    end
+  end
+
+  it "Warns when a fixture call is made in a before :context call" do
+    expect(RSpec).to receive(:warn_with).with(match(/Calling fixture method in before :context/))
+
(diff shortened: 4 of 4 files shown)


lib/rspec/rails/fixture_support.rb
```diff
@@ -29,6 +29,27 @@ module FixtureSupport
...
+              new_methods.each do |method_name|
+                proxy_method_warning_if_called_in_before_context_scope(method_name)
+              end
+            }
+          end
+
+          def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
+            orig_implementation = instance_method(method_name)
+            define_method(method_name) do |*args, &blk|
+              if self.inspect.include?("before(:context)")
```
JonRowe: Is this really the only way to detect this?

samphippen: I could not think of a better one after staring at the source... maybe @myronmarston has thoughts?

lib/rspec/rails/fixture_support.rb
```diff
@@ -29,6 +29,27 @@ module FixtureSupport
...
+                RSpec.warn_with("Calling fixture method in before :context ")
```
samphippen: This message needs to be better. @cupakromer @JonRowe can you post some suggestions in here?

JonRowe:
Overall this LGTM (I'd rather we had a better way of knowing we were in a hook but if thats the only way then c'est la vie) but the rubocop has a few words to say :)

lib/rspec/rails/fixture_support.rb
JonRowe: `Using fixtures from before(#{type}) hook`? Also what happens with `before(:group)` here? And `before(:all)` / `before(:suite)`?

samphippen: do you think we should interpolate the fixture method that was called?

lib/rspec/rails/fixture_support.rb
myronmarston: Yeah, this is pretty gross and brittle.  Also, I think we should also raise errors in `after(:context)` hooks, right?

Maybe we can make a change to rspec-core so this can be:

``` ruby
if RSpec::Core::Hooks::ContextHook === self
  # ...
end
```

lib/rspec/rails/fixture_support.rb
JonRowe: We return the line number / file right? If so no, otherwise yes.

lib/rspec/rails/fixture_support.rb
JonRowe: :+1:

myronmarston: Actually, if we're going to switch on type like that...could we just `include` a module with an appropriate method definition in a class like `RSpec::Core::Hooks::ContextHook`?  Seems much more straightforward.

samphippen: OK. It seems like the resolution here is to patch something into RSpec core. I'll open an issue over there.

samphippen:
blocked on https://github.com/rspec/rspec-core/issues/2130

lib/rspec/rails/fixture_support.rb
```diff
@@ -29,6 +29,27 @@ module FixtureSupport
...
           fixtures RSpec.configuration.global_fixtures if RSpec.configuration.global_fixtures
+
+          def self.fixtures(*args)
+            orig_methods = private_instance_methods
+            super.tap do
+              new_methods = private_instance_methods - orig_methods
+              new_methods.each do |method_name|
+                proxy_method_warning_if_called_in_before_context_scope(method_name)
+              end
+            end
```
cupakromer: Could we make this a name module (I'm ok with dynamically creating it) so that it's clear from the ancestor chain we've inserted something?

cupakromer:
We need to take into account possible `global_fixtures`. I think this doesn't cover that case.

samphippen:
Can you push a failing test that demonstrates that ?

lib/rspec/rails/fixture_support.rb
samphippen: :+1:

samphippen:
> Also note that :all can be provided as an option to allow everything.

This definitely works for the :all case... we're just capturing all the added methods

JonRowe:
This just needs @cupakromer's final change made right?

samphippen:
@JonRowe I think I understand enough to deal with this myself... I'll get to it a little later today if @cupakromer can't. He's a busy dude and so I'll try to not create any more work for him if I don't need to ^_^

JonRowe:
Nah I mean his final suggested change ;)

samphippen:
@JonRowe totally. I'm planning on getting to that sometime today.

JonRowe:
:+1: :)

samphippen:
@cupakromer @JonRowe done the module and the global_fixtures. Can I get a review one more time :) ?

samphippen:
@JonRowe @cupakromer I need an opinion here:

Using the module has broken on rails 3.0.20 only because it doesn't use a module, but instead redefines methods on the object proper. We could:

1. go back to the proxying approach
2. add conditional logic to use a module on rails versions > 3.0, and proxy only on 3.0

I personally prefer 1 because it will be less code, but makes it slightly harder to debug.

Thoughts please.

samphippen:
@JonRowe @cupakromer this is finished. Can one of you give it one last once over.

JonRowe:
I'm happy with this given the reasoning above.

PR #1502, opened 2015-12-09
Development improvements

deivid-rodriguez:
Hi!

I was trying to get the tests passing locally and noticed some issues. Some commits you might not be interested or might just be plain wrong, so feel free to cherry pick what you like!

Regards.


Gemfile-rails-dependencies
```diff
@@ -17,9 +17,9 @@ when /stable$/
 when nil, false, ""
   if RUBY_VERSION < '1.9.3'
     # Rails 4+ requires 1.9.3+, so on earlier versions default to the last 3.x release.
-    gem "rails", "~> 3.2.17"
+    gem "rails", "~> 3.2"
   else
-    gem "rails", "~> 4.0.4"
+    gem "rails", "~> 4.0"
```
JonRowe: This would allow Rails 4.1, which isn't the intent.

Gemfile-rails-dependencies
```diff
@@ -17,9 +17,9 @@ when /stable$/
 when nil, false, ""
   if RUBY_VERSION < '1.9.3'
     # Rails 4+ requires 1.9.3+, so on earlier versions default to the last 3.x release.
-    gem "rails", "~> 3.2.17"
+    gem "rails", "~> 3.2"
```
JonRowe: While in practise this has the same effect due to Rails policies, it isn't the intent of the original statement, a 3.3 release might break support for this.

Gemfile
```diff
@@ -21,7 +21,7 @@ gem 'sqlite3', '~> 1.3.6'
 
 # Capybara versions that support RSpec 3 only support RUBY_VERSION >= 1.9.3
 if RUBY_VERSION >= '1.9.3'
-  gem 'capybara', '~> 2.2.0', :require => false
+  gem 'capybara', '~> 2.2', :require => false
```
JonRowe: This has a different meaning in semantic versioning that I'm not sure we want, a major release would break this.

Rakefile
```diff
@@ -23,7 +23,7 @@ namespace :generate do
   desc "generate a fresh app with rspec installed"
   task :app do |_task|
     unless File.directory?('./tmp/example_app')
-      bindir = File.expand_path("bin")
+      bindir = File.expand_path("bin", './tmp/example_app')
```
JonRowe: Can you elaborate why this is necessary?

Gemfile-rails-dependencies
cupakromer: This is for the default Rails version someone will use when setting up the repo. I agree we probably don't want this to be `~> 4.0`, but we probably should bump this to either `~> 4.1.0` or `~> 4.2.0`. My vote is for the latter.

Rakefile
cupakromer: This changes what `bindir` is pointing to.

```ruby
bindir = File.expand_path("bin")
# => RSPEC_RAILS_REPO_DIR/bin
bindir = File.expand_path("bin", './tmp/example_app')
# => RSPEC_RAILS_REPO_DIR/tmp/example_app/bin
```

We want the original. This is specifically for handling installs which use `--binstubs`.

Rakefile
```diff
@@ -145,27 +145,26 @@ namespace :no_active_record do
     desc "generate a fresh app with no active record"
     task :app do
       unless File.directory?(example_app_dir)
-        bindir = File.expand_path("bin")
+        bindir = File.expand_path("bin", example_app_dir)
```
cupakromer: Same issue with `bindir` not being the same location.

Rakefile
```diff
@@ -145,27 +145,26 @@ namespace :no_active_record do
...
-            application_file = File.read("config/application.rb")
-            sh "rm config/application.rb"
-            File.open("config/application.rb","w") do |f|
-              f.write application_file.gsub(
-                "config.assets.enabled = true",
-                "config.assets.enabled = false"
-              )
-            end
+        in_example_app(:app_dir => example_app_dir) do
+          sh "./travis_retry_bundle_install.sh 2>&1"
```
cupakromer: Nice catch! :blue_heart:

rspec-rails.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.description = "rspec-rails is a testing framework for Rails 3.x and 4.x."
 
   s.files            = `git ls-files -- lib/*`.split("\n")
-  s.files           += %w[README.md License.md Changelog.md Capybara.md .yardopts .document]
+  s.files           += %w[README.md LICENSE.md Changelog.md Capybara.md .yardopts .document]
```
cupakromer: Thanks! :smile_cat:

example_app_generator/generate_app.rb
```diff
@@ -1,7 +1,7 @@
 rspec_rails_repo_path = File.expand_path("../../", __FILE__)
 rspec_dependencies_gemfile = File.join(rspec_rails_repo_path, 'Gemfile-rspec-dependencies')
 rails_dependencies_gemfile = File.join(rspec_rails_repo_path, 'Gemfile-rails-dependencies')
-bundle_install_path = File.join(rspec_rails_repo_path, '..', 'bundle')
+bundle_install_path = File.join(rspec_rails_repo_path, 'vendor', 'bundle')
```
cupakromer: Many of the core team setup their environment with a shared bundle path across all repos. This path is the directory containing the `rspec-rails` repo. Additionally, the Travis CI setup does this as well.

I do understand that this is likely not ideal for any system outside of the above subset. Perhaps we can come up with a better solution to keep all environments happy.

script/functions.sh
```diff
@@ -11,12 +11,6 @@ export JRUBY_OPTS=${JRUBY_OPTS:-"--server -Xcompile.invokedynamic=false"}
 SPECS_HAVE_RUN_FILE=specs.out
 MAINTENANCE_BRANCH=`cat maintenance-branch`
 
-# Don't allow rubygems to pollute what's loaded. Also, things boot faster
-# without the extra load time of rubygems. Only works on MRI Ruby 1.9+
-if is_mri_192_plus; then
-  export RUBYOPT="--disable=gem"
-fi
-
```
cupakromer: This needs to stay as it is a generated file from `rspec-dev`

script/predicate_functions.sh
```diff
@@ -23,18 +23,6 @@ function is_mri_192 {
...
-    if ruby -e "exit(RUBY_VERSION.to_f > 1.9)"; then
-      return 0
-    else
-      return 1
-    fi
-  else
-    return 1
-  fi
-}
-
```
cupakromer: Same here.

spec/generators/rspec/install/install_generator_spec.rb
```diff
@@ -27,7 +27,7 @@ def require_rspec_rails
     match(/^require 'rspec\/rails'$/m)
   end
 
-  def use_transactional_fixtures
+  def have_transactional_fixtures
```
cupakromer: How do you feel about `set_transactional_fixtures` or `enable_transactional_fixtures`?

spec/spec_helper.rb
```diff
@@ -1,9 +1,5 @@
 require 'rails/all'
 
-class << RSpec
-  attr_writer :configuration
-end
-
```
cupakromer: Nice clean up of old cruft. Looks like this got included as a private API in rspec/rspec-core#1808

Gemfile-rails-dependencies
deivid-rodriguez: This are all development dependencies. I consider a good thing that contributors try out newest versions and find out about breakages.

Gemfile
deivid-rodriguez: Same thing here, why is it bad?

Rakefile
deivid-rodriguez: Ok.

example_app_generator/generate_app.rb
deivid-rodriguez: Yeah, I find this setup weird. My `code` dir, where this setup installs gems, contains code I'm working on, not gem installations.

Anyways, removing this.

script/functions.sh
deivid-rodriguez: Are you interested in the patch for `rspec-dev`?

spec/generators/rspec/install/install_generator_spec.rb
deivid-rodriguez: Maybe `enabled_transactional_fixtures?`? This is detecting the feature, not enabling it, right?

script/functions.sh
JonRowe: No it's used to disable ruby gems as it says...

deivid-rodriguez:
Thanks for your comments, I'll update the patch.

script/functions.sh
deivid-rodriguez: Ok. Notice that this prevents installation of latest `nokogiri`, at least on my system. Hopefully they merge [this patch](https://github.com/sparklemotion/nokogiri/pull/1393) which fixes the issue (and makes setting this useless).

script/predicate_functions.sh
deivid-rodriguez: Not an expert here, but maybe a git submodule could be used to prevent having this kind of external files committed to version control in several places?

JonRowe: Yeah but that comes with it's own world of pain and version skew :)

deivid-rodriguez: I see. Have never used those, so I'll the ones who have already "suffered" them... :)

Gemfile-rails-dependencies
samphippen: This should be 4.2

spec/generators/rspec/install/install_generator_spec.rb
samphippen: this feels like unnecessary churn to me, I don't think any of those names are particularly stronger than what we have right now.

Gemfile-rails-dependencies
samphippen: I don't want to change this.

Gemfile
samphippen: I think the way we have it at the moment is fine.

script/functions.sh
samphippen: This removal shouldn't be merged, as others have commented.

spec/generators/rspec/install/install_generator_spec.rb
deivid-rodriguez: This is just to prevent a redefined method warning. I guess this conflicts with the original `use_transactional_fixtures`.

Gemfile-rails-dependencies
deivid-rodriguez: :+1:

samphippen:
@deivid-rodriguez thanks for the patch. There's quite a lot here. Can you remove all the things that we've talked about so that we can look at it again.

Gemfile
deivid-rodriguez: Sure, but why do you prefer testing against old versions?

script/functions.sh
deivid-rodriguez: As I said, ok. Just waiting for the other feedback to be clear before rebasing and force pushing.

Gemfile
cupakromer: I believe we need to be sure we don't break compatibility with the lowest version we currently support (out side of :bug: fixes / patches).

samphippen: Exactly, we need to test against older versions because we support older versions.

deivid-rodriguez: I see, but notice that you are _only_ testing against old versions.

samphippen: We understand that. It's not practical for us to slice down every version in our build matrix (we already slice about 30 ruby and rails combinations). Semantic versioning from capybara says they shouldn't break stuff going forward, so the testing impetus really falls to them here.

deivid-rodriguez: That makes sense. :+1:

deivid-rodriguez:
You (and Travis) can do a second review pass.

Gemfile-rails-dependencies
```diff
@@ -19,7 +19,7 @@ when nil, false, ""
     # Rails 4+ requires 1.9.3+, so on earlier versions default to the last 3.x release.
     gem "rails", "~> 3.2.17"
   else
-    gem "rails", "~> 4.0.4"
+    gem "rails", "~> 4.2"
```
cupakromer: `4.2.0` please :heart:

Rakefile
```diff
@@ -151,21 +151,19 @@ namespace :no_active_record do
...
-            application_file = File.read("config/application.rb")
-            sh "rm config/application.rb"
-            File.open("config/application.rb","w") do |f|
-              f.write application_file.gsub(
-                "config.assets.enabled = true",
-                "config.assets.enabled = false"
-              )
-            end
+        in_example_app(:app_dir => example_app_dir) do
+          sh "bundle install --gemfile ./Gemfile --path ../../../bundle"
```
cupakromer: Can we add back your addition of the following line just after `do`:

```
sh "./travis_retry_bundle_install.sh 2>&1"
```

deivid-rodriguez:
Here you go!

spec/generators/rspec/install/install_generator_spec.rb
samphippen: I think we should rename this to `transactional_fixtures_enabled?`

deivid-rodriguez: I'll rename that tomorrow!

deivid-rodriguez: Hei! I was going to rename this, but I noticed that the point of my first name was to make the two places where it is used read nice. For example,

```
expect(rails_helper).to have_transactional_fixtures
```

vs

```
expect(rails_helper).to transactional_fixtures_enabled?
```

Do you still want me to change this? Thanks!

samphippen: What about `have_transactional_fixtures_enabled?` ?

deivid-rodriguez: Ok!

deivid-rodriguez:
Done, this should now be good to go!

samphippen:
This LGTM. Merging.

deivid-rodriguez:
Thanks!

script/functions.sh
deivid-rodriguez: Good to see that you're giving this a second thought: https://github.com/rspec/rspec-dev/commit/e0639d522d952f883b6c4deca929231630afa775

PR #1503, opened 2015-12-10
Singularize classes for feature names

dysnomian:
This PR addresses issue #1403 by singularizing the class and table names in the feature generators.

### DIFF
--- a/lib/generators/rspec/feature/feature_generator.rb
+++ b/lib/generators/rspec/feature/feature_generator.rb
@@ -4,12 +4,25 @@ module Rspec
   module Generators
     # @private
     class FeatureGenerator < Base
-      class_option :feature_specs, :type => :boolean, :default => true, :desc => "Generate feature specs"
+      class_option :feature_specs, :type => :boolean, :default => true,  :desc => "Generate feature specs"
+      class_option :singularize,   :type => :boolean, :default => false, :desc => "Singularize the generated feature"
 
       def generate_feature_spec
         return unless options[:feature_specs]
 
-        template 'feature_spec.rb', File.join('spec/features', class_path, "#{class_name.downcase}_spec.rb") # file_name?
+        template template_name, File.join('spec/features', class_path, filename)
+      end
+
+      def template_name
+        options[:singularize] ? 'feature_singular_spec.rb' : 'feature_spec.rb'
+      end
+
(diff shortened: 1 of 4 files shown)


JonRowe:
Do we not have tests for the generators?

samphippen:
@dysnomian thanks for the pull request! Could you make it so that the singularization only happens on an option, rather than the default? I'm also *very* surprised that none of our tests failed here. @cupakromer my best guess is that it's because we use `rails g rspec:feature 'foos'` everywhere in our generator tests. Is that correct?

cupakromer:
Yes, it looks like we simply have been passing plural options in our specs - so nothing failed. Clearly a gap in our specs testing that we "pluralized" stuff (>ლ). Also the change to the filename isn't covered by our specs either :grimacing: 

While this PR doesn't exactly singularize anything, it doesn't pluralize everything. :+1: If we can add a spec case for this that would be great.

In regards to the naming, I'm a little concerned that is a divergent change. The original use of `table_name` resulted in:

```console
$ bin/rails g rspec:feature warehouse::widget
      create  spec/features/warehouse/warehouse_widgets_spec.rb
```

With this change we now will get:

```console
$ bin/rails g rspec:feature warehouse::widget                
      create  spec/features/warehouse/warehouse::widget_spec.rb
```

dysnomian:
@samphippen, @cupakromer: Thanks for the feedback! Next steps:
1. only singularize on a `--singularize` option
2. update tests so that they pass and fail appropriately on:
 * text
* spec filename
3. use `class_name` instead of `table_name` (that `::` was a surprising case I hadn't considered! :flushed: Good catch.)

Sound right?

samphippen:
@dysnomian that sounds good to me.

samphippen:
@dysnomian just wanted to check this was still something you're interested in working on? Let me know if you need any further help.

dysnomian:
@samphippen Yeah, still interested!

dysnomian:
Changes made. Any comments?

samphippen:
This looks good. I'm going to locally add a changelog, and merge it down.

samphippen:
@dysnomian it is my intention to merge this via #1508 when that PR passes CI.

samphippen:
This was merged via #1508

PR #1510, opened 2015-12-24
Prevent bad version of nokogiri and fix build

JonRowe:


JonRowe:
Merged despite error because it needs all three simultaneously. See rspec/rspec-rails#1510 rspec/rspec-support#258 and rspec/rspec-core#2140

PR #1512, opened 2015-12-29
Do not leak TestUnit specific methods after Rails 4

Ferdy89:
TestUnit requires the build_message method to be defined for its assertions to work properly. After Rails 4, Minitest is used instead of TestUnit so the method should not be included anymore. This is problematic because when using RSpec configured with
```ruby
config.expect_with :test_unit
```
the TestUnit assertions try to call build_message on the MinitestAssertionAdapter which will crash.

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -201,7 +201,7 @@ module RSpec
             select do |m|
               m.to_s =~ /^(assert|flunk|refute)/
             end
-          methods + [:build_message]
+          methods + test_unit_specific_methods
         end
 
         def define_assertion_delegators
@@ -209,6 +209,18 @@ module RSpec
             define_method(m.to_sym) do |*args, &block|
               assertion_delegator.send(m.to_sym, *args, &block)
             end
+          end
+        end
+
+        # Starting on Rails 4, Minitest is the default testing framework so no
+        # need to add TestUnit specific methods.
+        if ::Rails::VERSION::STRING >= '4.0.0'
+          def test_unit_specific_methods
+            []
+          end
+        else
+          def test_unit_specific_methods
+            [:build_message]
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/adapters.rb
```diff
@@ -201,7 +203,13 @@ def assertion_method_names
             select do |m|
               m.to_s =~ /^(assert|flunk|refute)/
             end
-          methods + [:build_message]
+          # Starting on Rails 4, Minitest is the default testing framework so no
+          # need to add TestUnit specific methods.
+          if ::Rails::VERSION::STRING >= '4.0.0'
```
soulcutter: As a point of style, it's nice to write method behavior that depends on constants by putting the conditional statement outside of the method definition so that it doesn't need to get calculated every time the method is called.

```ruby
if ::Rails::VERSION::STRING >= '4.0.0'
  def assertion_method_names
    […]
  end
else
  def assertion_method_names
    […]
  end
end
```

In this case, `define_assertion_delegators` only gets called when the concern is included, which is just twice AFAICT, so not a big deal, but wanted to mention it since it's a common pattern.

soulcutter:
928eddb8697927e07e2b971dfa46ef937a8dcd24 doesn't seem strictly necessary. RSpec 4.0 is removing 1.8.7 support, so adding changes specifically to address 1.8.7 compatibility in tests that were already working in all rubies is likely to get reverted or backed out soon. In other words, that commit is probably creating some additional work in the near-future.

lib/rspec/rails/adapters.rb
Ferdy89: Thanks for mentioning, I've updated the code accordingly

Ferdy89:
@soulcutter It's not strictly necessary, although it improves the quality of the current tests. I removed it, although I still use the same principle on the commit left. Otherwise the build breaks for Ruby <= 1.8.7.

In the end, since the bugfix is really only worth for Rails 4 and above, Ruby 1.8.7 does not matter. You tell me what to do, I'm fine either way. Thanks for reviewing!

Ferdy89:
@soulcutter any words on the actual bugfix? Does it make sense?

soulcutter:
Yeah, this looks good to me. Sorry for letting it dangle here for a bit, thanks for gently reminding me to revisit it.

PR #1514, opened 2015-12-30
Rails 5 support

sespindola:
I've loosened the rails version requirement as a first step towards supporting 5.x.

### DIFF
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -20,7 +20,7 @@ when nil, false, ""
     # Rails 4+ requires 1.9.3+, so on earlier versions default to the last 3.x release.
     gem "rails", "~> 3.2.17"
   else
-    gem "rails", "~> 4.2.0"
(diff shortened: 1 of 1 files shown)


sespindola:
Most builds seems to be failing with nokogiri.
Maybe you need to add "sudo apt-get install -y libxml2-dev" to the before_install section of the travis.yml file.

jasnow:
This is the Gemfile rule for nokogiri:
````
gem 'nokogiri', ['~> 1.5', '!= 1.6.6.3', '!= 1.6.6.4', "!= 1.6.7.1", "!= 1.6.7"]
```
and Travis log says:
```
An error occurred while installing nokogiri (1.6.7.1), and Bundler cannot continue.
```

JonRowe:
This is being handled by @samphippen elsewhere and we have the build running, currently you could add exceptions for beta1 but this doesn't apply for that.

PR #1523, opened 2016-01-08
Use RSpec::Support::ComparableVersion for capybara version check

yujinakayama:
... so that we can remove `RSpec::Support::VersionChecker` in rspec-support.


JonRowe:
Well it builds so :+1:

PR #1540, opened 2016-01-27
Prevent Rails 3.2.22.1 on Ruby 1.8.7

JonRowe:
Assuming this works it'll fix our builds elsewhere and improves upon @samphippen's quick travis fix :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,8 +104,6 @@ matrix:
     - rvm: 2.3.0
       env: RAILS_VERSION='~> 3.1.12'
   allow_failures:
-    - rvm: 1.8.7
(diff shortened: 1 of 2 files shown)


JonRowe:
@samphippen ready for review

samphippen:
@JonRowe this LGTM, do you think it wants a changelog entry?

JonRowe:
Hmm, I'm on the fence it's mostly internal but it will prevent 1.8.7 from installing that version of Rails so maybe?

samphippen:
it's marginally user facing, so adding a changelog entry can't hurt imo.

JonRowe:
Done

PR #1541, opened 2016-01-28
Allow to specify a custom automatic mapping `directory -> type`

jalkoby:
Hi everyone!
I've faced a strange (for me) behaviour of `infer_spec_type_from_file_location!`. I supposed that it will grep the first sub folder in spec, convert into singular and set that value into type. But after didn't get it in my project and looked into the source code, I found a static set of supported type. Well, it's fine solution too, but it would be useful to extend this set. So my pool request allows to do next:

```ruby
  config.infer_spec_type_from_file_location!(sidekiq: :workers, export: :exports)
  # or a shorter version
  config.infer_spec_type_from_file_location!([:worker, :export]) # => spec/workers -> :worker, spec/exports -> :export
```

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -94,8 +94,14 @@ module RSpec
           rendering_views
         end
 
-        def infer_spec_type_from_file_location!
-          DIRECTORY_MAPPINGS.each do |type, dir_parts|
+        def infer_spec_type_from_file_location!(list = {})
+          custom = list.inject({}) do |r, (k, v)|
+            dir = v ? v.to_s : "#{ k }s"
+            r[k] = ['spec', dir]
+            r
+          end
+
+          DIRECTORY_MAPPINGS.merge(custom).each do |type, dir_parts|
             escaped_path = Regexp.compile(dir_parts.join('[\\\/]') + '[\\\/]')
             define_derived_metadata(:file_path => escaped_path) do |metadata|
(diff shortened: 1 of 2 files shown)


yujinakayama:
You can use [`RSpec::Configuration#define_derived_metadata`](https://github.com/rspec/rspec-core/blob/v3.4.2/lib/rspec/core/configuration.rb#L1578-L1598) for the purpose.

jalkoby:
Well, it solves my problem, but lacks the usability. It would be good for the documentation to mention which types are supported by `config.infer_spec_type_from_file_location!` and add more information about "define_derived_metadata"( + supported options), because for now it's more like a hack in the internals of RSpec

myronmarston:
> Well, it solves my problem, but lacks the usability.

It's not as terse, but `define_derived_metadata` is _much_ more flexible -- it provides a way to automatically set any metadata, not just spec type.

> It would be good for the documentation to mention which types are supported by config.infer_spec_type_from_file_location! and add more information about "define_derived_metadata"( + supported options),

You're right.  Want to open a PR improving our documentation?

JonRowe:
It's worth mentioning that `config.infer_spec_type_from_file_location!` is to support "legacy" spec suites from pre RSpec 3, we recommend people tag their specs manually for clarity...

jalkoby:
@myronmarston worth to do it, as it's a frequent thing in my testing process

PR #1542, opened 2016-01-30
Backport: Fix empty template rendering for Rails 4.2.5.1

zimuliu:
Backport from https://github.com/rspec/rspec-rails/pull/1535

Refactor the template rendering overrides in controller specs to be more
compatible with the Action View public api. Previously `EmptyTemplatePathSetDecorator`
overrode the `find_all` method to wrap a path set. However `ActionView::PathSet`
is not public api so instead create a subclass of `ActionView::FileSystemResolver`
to return an empty template and treat `view_paths` as an array of string-like objects.

### DIFF
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -104,6 +104,10 @@ file "app/views/things/custom_action.html.erb",
      "This is a template for a custom action.",
      :force => true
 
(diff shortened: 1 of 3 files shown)


JonRowe:
We don't maintain older versions of RSpec. This has been released already please run `bundle update rspec-rails` to gain this fix.

PR #1543, opened 2016-01-31
Improve the documentation about automatic metadata

jalkoby:
Continuing the conversation https://github.com/rspec/rspec-rails/pull/1541

### DIFF
--- a/features/directory_structure.feature
+++ b/features/directory_structure.feature
@@ -28,7 +28,7 @@ Feature: Directory Structure
 
   For example, say the spec for the `ThingsController` is located in
   `spec/legacy/things_controller_spec.rb`. Simply tag the spec's
-  `RSpec.describe` block with the `type: :controller` metdata:
+  `RSpec.describe` block with the `type: :controller` metadata:
 
   ```ruby
   ​# spec/legacy/things_controller_spec.rb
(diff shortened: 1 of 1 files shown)


features/directory_structure.feature
```diff
@@ -67,6 +67,19 @@ Feature: Directory Structure
   configured `infer_spec_type_from_file_location!`, RSpec will automatically
   include the correct support functions for each type.
 
+  **Note:** `infer_spec_type_from_file_location!` affects only **the fixed list of folders**: controllers, helpers, jobs, mailers, models, requests, integration, api, routing, views, features.
+
+  To enable automatic metadata for other folder do the next:
```
JonRowe: This reads weirdly, how about `To enable automatic metadata for other folders, use `RSpec::Configuration#define_derived_metadata`, for an example see below:

features/directory_structure.feature
```diff
@@ -67,6 +67,19 @@ Feature: Directory Structure
...
+
+  ```ruby
+  RSpec.configuration do |config|
+    # mark all serializer specs with type :serializer
+    config.define_derived_metadata(:file_path => %r{/spec/serializers/}) do |metadata|
```
JonRowe: Cucumber doesn't like this for some reason, try running this feature file locally and then correcting the syntax until it works (I think it may be the use of %r{})

jalkoby:
@JonRowe I've updated the wording and change the regexp

features/directory_structure.feature
```diff
@@ -67,6 +67,19 @@ Feature: Directory Structure
   configured `infer_spec_type_from_file_location!`, RSpec will automatically
   include the correct support functions for each type.
 
+  **Note:** `infer_spec_type_from_file_location!` affects only **the fixed list of folders**: controllers, helpers, jobs, mailers, models, requests, integration, api, routing, views, features.
```
chrisarcand: I see this line as rather needless as the line directly above it says that: "If you follow the above listed canonical directory structure..." And the canonical structure mentioned is the very first thing this document goes over in detail.

Perhaps we can remove this line and just change the one below to "If you want to set metadata for a custom directory that doesn't follow fit the canonical structure above, you can do the following:" ? (Note I changed folder -> directory as well, since the rest of the document uses directory, not folder. Consistency FTW!)

JonRowe: :+1:

jalkoby:
The wording was changed. Also I found the reason of spec failing - cucumber needs a zero-length space before "#" comment symbol

JonRowe:
Thanks :heart:

chrisarcand:
:+1: :+1:

PR #1544, opened 2016-01-31
Fix performance regression with empty template rendering

### LINKED ISSUE #1537: Performance Regression in 3.4.1
While running specs on a large application I maintain, I noticed a significant slowdown in running controller specs where the action would ordinarily render a template, when using 3.4.1 as compared to 3.4.0.

I have provided an [example application which demonstrates this](https://github.com/dougal/rspec_performance_regression_demo_minimal).

### Benchmarking

The results I am seeing (using the timing in RSpec's own output) are:

#### 3.4.0

Run 1: 2.51s
Run 2: 2.55s
Run 3: 2.54s

Average: 2.53s

#### 3.4.1

Run 1: 6.99s
Run 2: 7.04s
Run 3: 7.06s

Average: 7.03s

### Conjecture

I suspect this is related to the changes in PR #1535, specifically the commit 42b44a6a6d. Having a quick look at the commit, there are at least two extra `map` operations in there compared to previously. I am not familiar with the code, and have not yet had the time to run a profiler with the specs in my example application to see if the extra iteration or resulting allocations are indeed the cause. I hope to do so tomorrow.

pixeltrix:
Closes #1537.

The change in 80637fc introduced a performance regression because the template lookup cache was being lost every time an example was run.

This commit fixes the problem by caching the EmptyTemplateResolver instances between example runs in a hash indexed by the path string.

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -85,16 +85,18 @@ module RSpec
         end
       end
 
+      # @private
+      RESOLVER_CACHE = Hash.new do |hash, path|
+        hash[path] = EmptyTemplateResolver.new(path)
+      end
+
       included do
         before do
           unless render_views?
             @_original_path_set = controller.class.view_paths
+            path_set = @_original_path_set.map { |resolver| RESOLVER_CACHE[resolver.to_s] }
 
-            empty_template_path_set = @_original_path_set.map do |resolver|
-              EmptyTemplateResolver.new(resolver.to_s)
-            end
-
-            controller.class.view_paths = empty_template_path_set
+            controller.class.view_paths = path_set
             controller.extend(EmptyTemplates)
           end
         end


JonRowe:
Can we add a benchmark for this? Before / After? That's what we'd normally do for this sort of thing... :)

pixeltrix:
@JonRowe commit a benchmark test to the repo?

JonRowe:
I meant to `./benchmarks` but that might actually be a bit tricky, I'd be happy with a note referencing why this was necessary.

pixeltrix:
NOTE: all tests carried out using the sample application in #1537

rspec/rspec-rails@e52281c (v3.4.0)

```
$ rspec spec/controllers/badgers_controller_spec.rb
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 2.07 seconds (files took 2.98 seconds to load)
1000 examples, 0 failures
```

rspec/rspec-rails@cf97d15 (v3.4.1)

```
$ rspec spec/controllers/badgers_controller_spec.rb
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 6.09 seconds (files took 4.25 seconds to load)
1000 examples, 0 failures
```

As you can see the tests take three times as long to run which is due to the extra file stats carried out whilst searching for the template. In 3.4.0 that template search is cached whereas 3.4.1 it isn't because we're creating a new resolver instance for every example run.

Caching the resolver instances gives the following improvement:

pixeltrix/rspec-rails@52522a9

```
$ rspec spec/controllers/badgers_controller_spec.rb
........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

Finished in 1.77 seconds (files took 2.97 seconds to load)
1000 examples, 0 failures
```

This actually now gives a slightly improved performance over 3.4.0.

JonRowe:
Thanks :)

yujinakayama:
Will we cut a release for this?

JonRowe:
Sure, why not, done.

yujinakayama:
:+1:

PR #1545, opened 2016-02-04
migrate if migrations are pending for all versions of rails

poudelprakash:
Guys, Instead of raising exception when tests are pending for rails < 4.1, wouldn't it be great to run migrations if pending.

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -22,17 +22,10 @@ require 'rspec/rails'
 #
 # Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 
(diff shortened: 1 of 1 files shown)


poudelprakash:
I don't know, what drawbacks it may have so I hope someone more experienced would enlighten me if this would be useless thing to do.

JonRowe:
On rails versions that support it, we do run the migrations, that's what this does:

```Ruby
ActiveRecord::Migration.maintain_test_schema!
```

On other versions all we can do is check for migrations.

poudelprakash:
@JonRowe on other versions too we can run migrations, I have used this tweak in my project and it's working smoothly, why do you say "on other versions all we can do is check for migrations"

JonRowe:
Using public APIs

PR #1550, opened 2016-02-08
Fixing index view tests for a scaffold with multiple float attributes

Alwahsh:
If a model has multiple float attributes, the auto generated tests are failing because of duplicate value(All attributes have the default value and tests check that the count is equal to 2). This fixes it. The issue still exists for double attributes but I'd like to know what you think about this before I do any further work.


JonRowe:
Thanks, it'd be good to have some sort of test for this

Alwahsh:
I've added 2 simple tests in addition to a scaffold that would fail without this change in the example_app_generator. Not sure if that's what you were looking forward to :smile:

JonRowe:
Thanks

PR #1557, opened 2016-02-23
Fix empty view rendering

### LINKED ISSUE #1556: append_view_path monkey patch breaks API
https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/view_rendering.rb#L77 is being called instead of Rails' implementation https://github.com/rails/rails/blob/7f18ea14c893cb5c9f04d4fda9661126758332b5/actionview/lib/action_view/view_paths.rb#L51

The difference is that rspec-rails' version expects a string while Rails' version also accepts an array.

Fixing this could be done with something like:
```ruby
decorators = Array(new_path).map { |path| _path_decorator(path) }
lookup_context.view_paths.push(*decorators)
```
#1551 might resolve this by accident if merged.

JonRowe:
This fixes two regressions that were introduced by #1535.

Closes #1556
Closes #1551 

/cc @samphippen @pixeltrix

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -71,17 +71,17 @@ module RSpec
       # @private
       module EmptyTemplates
         def prepend_view_path(new_path)
-          lookup_context.view_paths.unshift(*_path_decorator(new_path))
+          lookup_context.view_paths.unshift(*_path_decorator(*new_path))
         end
 
         def append_view_path(new_path)
-          lookup_context.view_paths.push(*_path_decorator(new_path))
+          lookup_context.view_paths.push(*_path_decorator(*new_path))
         end
 
       private
 
-        def _path_decorator(path)
-          EmptyTemplateResolver.new(path)
+        def _path_decorator(*paths)
+          paths.map { |path| EmptyTemplateResolver.new(path) }
         end
       end
 
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -44,6 +44,13 @@ module RSpec
       #
       # @private
       class EmptyTemplateResolver < ::ActionView::FileSystemResolver
+
+        def initialize(path, pattern=nil)
+          unless path.is_a?(::ActionView::Resolver)
+            super
+          end
+        end
+
       private
 
         def find_templates(*args)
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/view_rendering.rb
```diff
@@ -44,6 +44,13 @@ def render_views?
       #
       # @private
       class EmptyTemplateResolver < ::ActionView::FileSystemResolver
+
+        def initialize(path, pattern=nil)
+          unless path.is_a?(::ActionView::Resolver)
+            super
+          end
+        end
+
```
pixeltrix: Doesn't this just end up adding an uninitialised `EmptyTemplateResolver` to the view paths array?

JonRowe: Yes but that's the point, the parts of this that are used are here and don't depend on initialisation.

pixeltrix: This bit is to fix #1551 isn't it? If so adding your custom resolver instance to the view path with result in an instance of `EmptyTemplateResolver` with a `nil` path instance variable instead. I thought the idea would be to put the custom resolver into the path without wrapping it.

JonRowe: We don't want a custom resolver in the path, the aim is to disable view rendering, my understanding is that would do the opposite.

pixeltrix: Yes, but it'd generate template missing errors because `find_templates` would be called on the `EmptyTemplateResolver` instance and not the custom resolver instance.

JonRowe: Well we need to find a way to pass the details of the resolver down into that implementation without loosing this subclass, alternatively we could monkey patch the instance of the resolver with the required methods but I'd rather not do that, that seems dangerous

bquorning: The problem is that when `path` is a `Resolver` instance, it knows how to look up templates. And the `super` call in this `EmptyTemplateResolver#find_templates` doesn’t call up to that resolver.

bquorning: Please see #1568 for an changed implementation that works for my spec suite.

JonRowe:
Closed in prerferance to #1580

Issue #1525, opened 2016-01-12
View tests don't include any helpers


bronson:
With RSpec 3.4 / Rails 4.2, I needed to mock controller-defined helpers.

Now, with RSpec master / Rails 5, it appears I need to mock *all* helpers, including built-ins (model_url, etc).

Is this intentional?  #1076 makes it clear that leaving out Controller Helpers is intentional.  Not including any helpers at all, however, might a little extreme?

FWIW, here are the changes I needed to make to my View spec:
```diff
 describe 'search/print/_property.html.haml', type: :view do
   it 'displays the page' do
     property = build(:property)
+
+    # rails 5/rspec master doesn't include Helper classes and built-ins (Rails 4.2/rspec 3.4 did)
+    view.class_eval('include ApplicationHelper')
+    view.class_eval('include SearchHelper')
+    # even the built-in helpers!
+    allow(view).to receive(:property_path).and_return("/property/#{property.property_id}")
+
+    # rpsec intentially wants us to stub all controller-defined helpers
     # https://github.com/rspec/rspec-rails/issues/1076#issuecomment-59842447
     allow(view).to receive(:search_params).and_return(Search::Params.new({}))
     allow(view).to receive(:client_state).and_return(Search::Params.new({}))
+
     render 'search/print/property', property: property
     assert_select ":match('href', ?)", "/property/#{property.property_id}"
   end
```

JonRowe:
Rails 5 support is in a pre beta state, but cc @samphippen

samphippen:
this seems like it should work... I'm gonna have to investigate.

samphippen:
It's worth noting view tests are not really a thing anyone writes in rails without RSpec, so there's probably some fail here.

pixeltrix:
I can't reproduce this with the latest RSpec master and Rails 5.0.0.beta2 - the url helpers are both available in the template context and the spec context. Recommend closing unless some new information comes to light.

samphippen:
@bronson if you can give us a rails app to clone, which demonstrates the issue, we'll happily fix. In the meantime, I'm going to close.

bronson:
OK.  It's still happening in my app on beta2.  If nobody else is seeing this, it's probably something obscure with my app.

If anyone else runs into it, I'll take the time to isolate it.  Otherwise, happy to leave it closed.

pschambacher:
FYI running into this problem as well with route/url helpers. Investigating where the problem comes from at the moment.

[...]

PR #1559, opened 2016-02-24
Rework integration of rails-controller-testing

pschambacher:
/cc @samphippen

# Problem

When adding `rails-controller-testing` to your Gemfile, you're actually loading a bit more than you wish. In our case we ended up with some helpers missing from our view specs for instance.
I assume that all the various unnecessary includes performed in [rails-controller-testing.rb](https://github.com/rails/rails-controller-testing/blob/master/lib/rails-controller-testing.rb) are to blame.

# Solution

The only interesting bit about rails-controller-testing is actually these requires

```ruby
require 'rails/controller/testing/test_process'
require 'rails/controller/testing/integration'
require 'rails/controller/testing/template_assertions'
```

So I would recommend to use

```ruby
gem 'rails-controller-testing`, require: false
```

Now I changed the code here so instead of using `defined?(::Rails::Controller::Testing)` and assume that the gem was loaded before, I'm using a rescue block to catch `LoadError`. If the gem isn't present, the block will silently failed, otherwise it will perform the proper inclusion required by RSpec.

# Follow-up

I started my work here but I think something should be changed in rails-controller-testing so that an automatic load of the gem doesn't assume that everybody's using minitest...

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -20,7 +20,7 @@ in_root do
   gsub_file "Gemfile", /.*debugger.*/, ''
 
   if Rails::VERSION::STRING >= '5.0.0'
-    append_to_file('Gemfile', "gem 'rails-controller-testing', :git => 'https://github.com/rails/rails-controller-testing'\n")
+    append_to_file('Gemfile', "gem 'rails-controller-testing', :git => 'https://github.com/rails/rails-controller-testing', require: false\n")
   end
 
   # Nokogiri version is pinned in rspec-rails' Gemfile since it tend to cause installation problems
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -49,6 +49,25 @@ module RSpec
       config.include RSpec::Rails::FeatureExampleGroup,    :type => :feature
     end
 
+    # rubocop:disable HandleExceptions
+    # Tries to load and include rails-controller-testing
+    #
+    # @api private
+    def self.add_rails_controller_testing(config)
+      require 'rails/controller/testing/test_process'
+      require 'rails/controller/testing/template_assertions'
+      require 'rails/controller/testing/integration'
+
+      [:controller, :view, :request].each do |type|
+        config.include ::Rails::Controller::Testing::TestProcess, :type => type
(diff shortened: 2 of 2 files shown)


pschambacher:
@samphippen anything I can do to help on this PR? Red Travis is due so a worker not starting correctly.

samphippen:
@pschambacher I'll kick the build. I'm wondering if we should upstream this. @sgrif and @pixeltrix do either of you have thoughts?

samphippen:
This is green. @rspec/rspec is this something we want to merge? I'm still unsure about whether we want to merge this or whether it's something to upstream.

JonRowe:
I think this needs to be fixed upstream...

pixeltrix:
@JonRowe I'm not entirely sure what has to be fixed here? The gem has extracted `assigns` and `assert_template` - why it would affect view helpers is unclear and the discussion on #1525 is somewhat confused to what the cause is. If someone can push up a demo Rails app that exhibits the problem I'm happy to take a look at it.

samphippen:
So I think our integration with the rails controller testing gem is fine. @pschambacher if you can provide a rails app which shows a clear reproduction case, we can re-open and discuss, but I'm going to close this for now.

wintersolutions:
I'm having the same symptoms as @pschambacher described. I.e. custom helper methods are missing in view specs.

I added the following to my Gemfile:

`gem 'rails-controller-testing', require: false, group: :test`

And the following to spec/support/config/rails_controller_matching.rb (load support dir with rspec):

```
require 'rails/controller/testing/test_process'
require 'rails/controller/testing/integration'
require 'rails/controller/testing/template_assertions'

RSpec.configure do |config|
  config.include Rails::Controller::Testing::TestProcess, type: :controller
  config.include Rails::Controller::Testing::Integration, type: :controller
  config.include Rails::Controller::Testing::TemplateAssertions, type: :controller
end
```

After that the custom helpers are there again.

I tried to create a simple Rails application in which this behaviour can be reproduced, but everything works in this simple app i.e. error must be caused by some configuration and/or complex test cases.

samphippen:
@wintersolutions my best guess here is that one or more extra gems in your rails stack is doing something that breaks our setup. It's not really feasible for us to debug a fully fledged rails app, and without a solid repro case we can't really advance this issue. Paging @sgrif and @pixeltrix to see if you've got any ideas?

PR #1565, opened 2016-02-28
Move #738 regression check to spec

JonRowe:
This is confusing and not useful as documentation, all #738 was intended
to do was to prevent our let definitions from being defined earlier than
3rd party inclusions (and hence being overridden) which was confusing
behaviour, so move it to a spec.

/cc @samphippen @myronmarston

### DIFF
--- a/features/controller_specs/controller_spec.feature
+++ b/features/controller_specs/controller_spec.feature
@@ -34,29 +34,6 @@ Feature: controller spec
     When I run `rspec spec`
     Then the example should pass
 
-  Scenario: controller is extended with a helper module
-    Given a file named "spec/controllers/widgets_controller_spec.rb" with:
-      """ruby
-      require "rails_helper"
-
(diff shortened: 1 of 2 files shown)


myronmarston:
Did you confirm that this spec fails if you revert the fix from #738?  Not absolutely necessary, but it's always nice to check that this spec will catch a regression here.  I've seen enough subtle differences between the spec environment and the cucumber environment that it's hard to tell w/o actually trying it.

JonRowe:
I didn't, but I did check how the behaviour was affected by including in various places vs the let definition and I'm confident this spec represents the behaviour we expect. We could remove this check altogether if you like because the fix was actually reverted after an update to rspec-core...

myronmarston:
LGTM except for the travis failure.

JonRowe:
Merging despite build failure, as that happens to be unrelated (master also fails)

PR #1567, opened 2016-03-03
Allow to use have_enqueued_job matcher without passing block.

morgoth:
It allows to test on job arguments that are dynamically created inside called block.

Example:

```ruby
class CreateUser
  def self.call
    user = User.create!
    DeliverWelcomeEmailJob.perform_later(user)
    user
  end
end

user = CreateUser.call
expect().to have_enqueued_job(DeliverWelcomeEmailJob).with(user)
```

This is not fully working yet, as rspec doesn't allow to call `expect()` without passing block or any argument:
```
     ArgumentError:
       You must pass either an argument or a block to `expect`.
```

Do you have an idea how can I bypass this check?

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -18,14 +18,16 @@ module RSpec
             set_expected_number(:exactly, 1)
           end
 
-          def matches?(proc)
-            raise ArgumentError, "have_enqueued_jobs only supports block expectations" unless Proc === proc
+          def matches?(proc = nil)
+            jobs = if Proc === proc
+                     original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
+                     proc.call
+                     queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
+                   else
+                     queue_adapter.enqueued_jobs
+                   end
 
-            original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
-            proc.call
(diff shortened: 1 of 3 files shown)


JonRowe:
> Do you have an idea how can I bypass this check?

Please don't bypass this check.

benoittgt:
Very strange idea. Maybe it's a joke from Bart Simpson ? 😄

morgoth:
@benoittgt well, no. I want to solve an issue with limitations of this matcher. As I can see it's already "workarounded" in https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/matchers/redirect_to.rb#L13 by providing some not relevant argument.

Anyway, only other solution that comes to my mind is to introduce new matcher, something like`expect(HelloJob).to be_enqueued.{with|on_queue|at}`.
If you think that this is a way to go, I can work on it

morgoth:
Another weird workaround would be to write `expect(:code_above).to have_enqueued_job...`

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -157,6 +159,9 @@ def queue_adapter
       #     expect {
       #       HelloJob.set(wait_until: Date.tomorrow.noon, queue: "low").perform_later(42)
       #     }.to have_enqueued_job.with(42).on_queue("low").at(Date.tomorrow.noon)
+      #
+      #     HelloJob.perform_later(user_id: 2)
+      #     expect().to have_enqueued_job(HelloJob).with(user_id: 2)
```
JonRowe: This isn't syntax we want to support, it makes no grammatical sense.

PR #1568, opened 2016-03-08
Delegate properly to custom Resolver instances

bquorning:
Let `EmptyTemplateResolver` creating instances of either `EmptyTemplateResolver::ResolverDecorator` or `EmptyTemplateResolver::FileSystemResolver`, depending on whether the given path is an `ActionView::Resolver` instance or not.

The new `EmptyTemplateResolver::ResolverDecorator` class simply delegates all method calls to its given resolver, except for `#find_templates` which it overrides to return templates that render with `EmptyTemplateHandler`.

Please note that the merge target is PR #1557, not master.

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -40,16 +40,12 @@ module RSpec
       end
 
       # @private
-      class EmptyTemplateResolverFactory
-        def initialize(path)
-          @path = path
-        end
-
-        def resolver
-          if @path.is_a?(::ActionView::Resolver)
-            EmptyTemplateResolverDecorator.new(@path)
+      class EmptyTemplateResolver
+        def self.build(path)
+          if path.is_a?(::ActionView::Resolver)
+            EmptyTemplateResolverDecorator.new(path)
           else
-            EmptyTemplateFileSystemResolver.new(@path)
+            EmptyTemplateFileSystemResolver.new(path)
           end
         end
       end
@@ -125,13 +121,13 @@ module RSpec
       private
 
         def _path_decorator(*paths)
-          paths.map { |path| EmptyTemplateResolverFactory.new(path).resolver }
+          paths.map { |path| EmptyTemplateResolver.build(path) }
         end
       end
 
       # @private
       RESOLVER_CACHE = Hash.new do |hash, path|
-        hash[path] = EmptyTemplateResolverFactory.new(path).resolver
+        hash[path] = EmptyTemplateResolver.build(path)
       end
 
       included do
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -39,6 +39,7 @@ module RSpec
         self.class.render_views? || !controller.class.respond_to?(:view_paths)
       end
 
+      # @private
       class EmptyTemplateResolverFactory
         def initialize(path)
           @path = path
@@ -53,6 +54,10 @@ module RSpec
         end
       end
 
+      # Delegates find_templates to the submitted resolver and then returns templates
+      # with modified source
+      #
+      # @private
       class EmptyTemplateResolverDecorator
         def initialize(resolver)
           @resolver = resolver
@@ -78,8 +83,8 @@ module RSpec
         end
       end
 
-      # Delegates find_all to the submitted path set and then returns templates
-      # with modified source
+      # Delegates find_templates to the submitted path set and then returns
+      # templates with modified source
       #
       # @private
       class EmptyTemplateFileSystemResolver < ::ActionView::FileSystemResolver
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -43,30 +43,13 @@ module RSpec
       class EmptyTemplateResolver
         def self.build(path)
           if path.is_a?(::ActionView::Resolver)
-            EmptyTemplateResolverDecorator.new(path)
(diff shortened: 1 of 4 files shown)


samphippen:
LGTM, but I'd like another review.

lib/rspec/rails/view_rendering.rb
```diff
@@ -39,18 +39,50 @@ def render_views?
         self.class.render_views? || !controller.class.respond_to?(:view_paths)
       end
 
-      # Delegates find_all to the submitted path set and then returns templates
-      # with modified source
-      #
-      # @private
-      class EmptyTemplateResolver < ::ActionView::FileSystemResolver
+      class EmptyTemplateResolverFactory
```
JonRowe: When we build thinks like this we normally make them "instance less" to avoid unnecessary allocations, please refactor to a static instance, or better yet a module method. Can I also suggest a better name is desired? Something like `DecoratePath` or `BuildEmptyTemplateResolver`

lib/rspec/rails/view_rendering.rb
```diff
@@ -39,18 +39,50 @@ def render_views?
...
+        def resolver
+          if @path.is_a?(::ActionView::Resolver)
+            EmptyTemplateResolverDecorator.new(@path)
+          else
+            EmptyTemplateFileSystemResolver.new(@path)
           end
         end
+      end
 
+      class EmptyTemplateResolverDecorator
```
bquorning:
```
# Delegates find_templates to the submitted resolver and then returns templates
# with modified source
#
# @private
```

lib/rspec/rails/view_rendering.rb
```diff
@@ -39,18 +39,50 @@ def render_views?
...
+              template.identifier,
+              EmptyTemplateHandler,
+              :virtual_path => template.virtual_path,
+              :format => template.formats
+            )
+          end
+        end
+      end
+
+      # Delegates find_all to the submitted path set and then returns templates
```
bquorning: Should be `find_templates`.

lib/rspec/rails/view_rendering.rb
```diff
@@ -39,18 +39,50 @@ def render_views?
...
+
+        def method_missing(name, *args, &block)
+          @resolver.send(name, *args, &block)
+        end
+
+      private
+
+        def find_templates(*args)
+          templates = @resolver.find_templates(*args)
+          templates.map do |template|
```
JonRowe: This map could probably now be delegated to outside as it does the same as below.

lib/rspec/rails/view_rendering.rb
```diff
@@ -88,13 +120,13 @@ def append_view_path(new_path)
       private
 
         def _path_decorator(*paths)
-          paths.map { |path| EmptyTemplateResolver.new(path) }
+          paths.map { |path| EmptyTemplateResolverFactory.new(path).resolver }
```
JonRowe: See above about allocations.

lib/rspec/rails/view_rendering.rb
bquorning: Yeah, but I don’t really know where to move it. `EmptyTemplateHandler`?

lib/rspec/rails/view_rendering.rb
bquorning: Agree on both the allocation and naming concerns. How does `EmptyTemplateResolver.build` sound?

bquorning: Actually, we could move `EmptyTemplateFileSystemResolver` and `EmptyTemplateResolverDecorator` into this `EmptyTemplateResolver` module (class), since they’re already supposed to be private and are only used by this factory class.

bquorning:
There, I pushed some changes:

 1. Changed the factory from an instance method to a class method: `EmptyTemplateResolver.build`
 2. Renamed and moved the `ResolverDecorator` and `FileSystemResolver` classes into `EmptyTemplateResolver::` namespace.
 3. This allowed for extracting the duplicated mapping into an `EmptyTemplateResolver.nullify_template_rendering` method.

samphippen:
@JonRowe LGTM

samphippen:
Why are there no travis results for this branch?

samphippen:
(did not mean to close)

bquorning:
> Why are there no travis results for this branch?

Perhaps because the merge target is not master branch?

JonRowe:
Can you provide an example of the code this is intended to fix? I'm still hesitant around this as I haven't seen a real world scenario that's broken by the existing code...

bquorning:
@JonRowe I created a gist at https://gist.github.com/bquorning/6bf534e4b145b1738bb2. It’s a bit contrived, since I didn’t want to create tables and fetch custom templates from the database.

ento:
re: example:
I have a group of namespaced controllers that share the same templates and a module of shared actions. I believe this PR will fix my anonymous controller test for the module, which uses [a custom resolver](http://blog.plataformatec.com.br/2011/04/default-views-in-rails-3-0-with-custom-resolvers/) to get to those templates.

bquorning:
By the way, should I rather make a PR against master, so that we can see some Travis runs?

bquorning:
Added 1 commit that allows Travis to run. Please don’t merge the PR with this commit included.

bquorning:
Build fails because of the new version of Rake that has [Removed to support Ruby 1.8.x](https://github.com/ruby/rake/blob/master/History.rdoc#1100--2016-03-09)

bquorning:
Superseded by #1580.

PR #1570, opened 2016-03-10
Don't use Rake v11 with older rubies

bquorning:
It doesn't work with Ruby 1.8.7 and 1.9.2.

The latest master build has 8 jobs failing because of the new release of Rake. And it’s hard to work on pull requests when you can’t get a green build :-)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,12 @@ end
 
 gem 'sqlite3', '~> 1.3.6'
 
+if RUBY_VERSION >= '1.9.3'
+  gem 'rake', '>= 10.0.0'
+else
+  gem 'rake', '~> 10.0' # rake 11 requires Ruby 1.9.3 or later
+end
+
 # Capybara versions that support RSpec 3 only support RUBY_VERSION >= 1.9.3
 if RUBY_VERSION >= '1.9.3'
   gem 'capybara', '~> 2.2.0', :require => false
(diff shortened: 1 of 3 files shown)


JonRowe:
See how we've done this on rspec/rspec-core#2200

bquorning:
Thank you @JonRowe. I’ve updated the code.

Eric-Guo:
Seems the root cause is due to rake 11 [`last_comment` rename to `last_description`](https://github.com/ruby/rake/issues/116), so why can not change master of rspec-core similar to [PR 2202](https://github.com/rspec/rspec-core/pull/2202)?

bquorning:
This gem doesn’t use `last_comment`, nor `last_description`.

Eric-Guo:
@bquorning Sorry miss leading, I just found the exactly [PR of rspec-core](https://github.com/rspec/rspec-core/pull/2197), I means since the rspec-core master will support rake 11, so no need here to constraint to using rake 10 only.

bquorning:
Rake v11 doesn’t work with Ruby 1.8.7 and 1.9.2. This PR allows newer versions of Ruby to use the new Rake, while these two older Rubies are limited to the older, compatible version of Rake.

Eric-Guo:
@bquorning you are right, I miss leading again :smile:

bquorning:
No problem :-)

bquorning:
Ready to review/merge.

PR #1578, opened 2016-03-16
Support `with` block in `have_enqueued_job` matcher

### LINKED ISSUE #1576: have_enqueued_job can't be used with data that has not yet been created
This is somewhat a continuation of #1495.  I am using rspec-rails 3.4.2.

The crux of the issue is that data created in the `expect` block can't be used for validation in parameters to `with`.

E.g. if I have a controller like:

```ruby
UsersController < ApplicationController
  def create
    @user = User.create user_params
    UserMailer.welcome(@user).deliver_later
  end
end
```

And I want to write a test that checks that the mailer was enqueued with the correct parameters, I might write a test like:

```ruby
it 'sends welcome email' do
  expect {
    post :create, params
  }.to have_enqueued_job(ActionMailer::DeliveryJob).with(
    'UserMailer',
    'welcome',
    'deliver_now',
    Account.order(:created_at).last,
  )
end
```

But that doesn't work because the parameters passed to `with` are evaluated before the `expect` block is executed.

In #1495, @JonRowe suggests doing something like:

```ruby
my_model = nil
expect {
  my_model = MyServiceObject.call(...)
}.to have_enqueued_job(MyJob) { |arg| expect(arg).to eq GlobalID.create(my_model) }
```

But that doesn't actually work.  The block after `have_enqueued_job` never actually gets executed.  AFAICT there is no supported way of doing something like this so I thought it might be worth another look.

I have worked around this with a monkey patch like so:

```ruby
module RSpec
  module Rails
    module Matchers
      module ActiveJob
        class HaveEnqueuedJob
          private

          def serialized_attributes
            args_to_serialize = []
            @args.each do |arg|
              arg = arg.call if Proc === arg
              args_to_serialize << arg
            end

            {}.tap do |attributes|
              attributes[:args]  = ::ActiveJob::Arguments.serialize(args_to_serialize) if args_to_serialize.any?
              attributes[:at]    = @at.to_f if @at
              attributes[:queue] = @queue if @queue
              attributes[:job]   = @job if @job
            end
          end
        end
      end
    end
  end
end
```

Then I can change my test to pass lambdas to `with` that get evaluated after the `expect` block is executed:

```ruby
it 'sends welcome email' do
  expect {
    post :create, params
  }.to have_enqueued_job(ActionMailer::DeliveryJob).with(
    'UserMailer',
    'welcome',
    'deliver_now',
    -> {Account.order(:created_at).last},
  )
end
```

I wonder if there is an argument to be made to legitimize the use of lambdas like this given that GlobalID does not support serializing procs/lambdas?  Any other thoughts on supporting this type of scenario?

jschroeder9000:
Fixes #1576.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -15,6 +15,7 @@ module RSpec
             @args = []
             @queue = nil
             @at = nil
+            @with_block = nil
             set_expected_number(:exactly, 1)
           end
 
@@ -26,7 +27,12 @@ module RSpec
             in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
 
             @matching_jobs_count = in_block_jobs.count do |job|
-              serialized_attributes.all? { |key, value| value == job[key] }
+              if serialized_attributes.all? { |key, value| value == job[key] }
+                @with_block.call(*job[:args]) if @with_block.present?
+                true
+              else
+                false
+              end
             end
 
             case @expectation_type
@@ -36,8 +42,9 @@ module RSpec
             end
           end
 
-          def with(*args)
+          def with(*args, &block)
             @args = args
+            @with_block = block
             self
           end
 
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -15,7 +15,7 @@ module RSpec
             @args = []
             @queue = nil
             @at = nil
-            @with_block = nil
+            @block = Proc.new {}
             set_expected_number(:exactly, 1)
           end
 
@@ -28,7 +28,7 @@ module RSpec
 
             @matching_jobs_count = in_block_jobs.count do |job|
               if serialized_attributes.all? { |key, value| value == job[key] }
-                @with_block.call(*job[:args]) if @with_block.present?
+                @block.call(*job[:args])
                 true
               else
                 false
@@ -44,7 +44,7 @@ module RSpec
 
           def with(*args, &block)
             @args = args
-            @with_block = block
+            @block = block if block.present?
             self
           end
 
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -218,5 +218,38 @@ RSpec.describe "ActiveJob matchers", :skip => !RSpec::Rails::FeatureCheck.has_ac
 
       ActiveJob::Base.queue_adapter = queue_adapter
     end
+
+    it "fails with with block with incorrect data" do
+      expect {
+        expect {
+          hello_job.perform_later("asdf")
+        }.to have_enqueued_job(hello_job).with { |arg|
+          expect(arg).to eq("zxcv")
+        }
+      }.to raise_error { |e|
+        expect(e.message).to match(/expected: "zxcv"/)
+        expect(e.message).to match(/got: "asdf"/)
+      }
+    end
+
+    it "passes multiple arguments to with block" do
+      expect {
+        hello_job.perform_later("asdf", "zxcv")
+      }.to have_enqueued_job(hello_job).with { |first_arg, second_arg|
+        expect(first_arg).to eq("asdf")
+        expect(second_arg).to eq("zxcv")
+      }
+    end
+
+    it "only calls with block if other conditions are met" do
+      noon = Date.tomorrow.noon
+      midnight = Date.tomorrow.midnight
+      expect {
+        hello_job.set(:wait_until => noon).perform_later("asdf")
+        hello_job.set(:wait_until => midnight).perform_later("zxcv")
+      }.to have_enqueued_job(hello_job).at(noon).with { |arg|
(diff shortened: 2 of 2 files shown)


jschroeder9000:
Here's a first crack at fixing #1576.  Hopefully I'm on the right track.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -26,7 +27,12 @@ def matches?(proc)
             in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
 
             @matching_jobs_count = in_block_jobs.count do |job|
-              serialized_attributes.all? { |key, value| value == job[key] }
+              if serialized_attributes.all? { |key, value| value == job[key] }
```
JonRowe: Rather than this, set `@with_block` to an empty lambda and call it regardless.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -15,6 +15,7 @@ def initialize(job)
             @args = []
             @queue = nil
             @at = nil
+            @with_block = nil
```
JonRowe: You'd be better off calling this `@block` and setting it to a passed in block or an empty lambda.

JonRowe:
Great start

jschroeder9000:
Pushed revisions.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -26,7 +27,12 @@ def matches?(proc)
             in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
 
             @matching_jobs_count = in_block_jobs.count do |job|
-              serialized_attributes.all? { |key, value| value == job[key] }
+              if serialized_attributes.all? { |key, value| value == job[key] }
+                @block.call(*job[:args])
```
JonRowe: You should call the block regardless

jschroeder9000: My thinking was that it could be useful to try to disambiguate where the block is intended to be applied to.  For instance if multiple jobs get enqueued, you may only want to check the arguments of one of them.  E.g. calling the block always would break this spec:

https://github.com/jschroeder9000/rspec-rails/blob/8383c95f244168eaaf9deb1553b96d5a3e06ec81/spec/rspec/rails/matchers/active_job_spec.rb#L244

But if that spec were called by passing paramaters to `with` instead of a block, it would pass.  I think they should be consistent.

Giving it some more thought, though, that does seem like a flawed approach because there still would have to be some other unique attributes about the jobs to differentiate them and even that may not always be the case.

In the simplest and most general case, I think it makes sense both of these specs should behave identically:

```ruby
it 'passes with arguments' do
  expect {
    hello_job.perform_later('asdf')
    hello_job.perform_later('zxcv')
  }.to have_enqueued_job(hello_job).with('asdf')
end

it 'passes with a block' do
  expect {
    hello_job.perform_later('asdf')
    hello_job.perform_later('zxcv')
  }.to have_enqueued_job(hello_job).with { |arg|
    expect(arg).to eq('asdf')
  }
end
```

The behavior of passing arguments is that the spec would pass, but if I just always call the block without doing anything else then it would fail because _all_ enqueued jobs would have to meet the expectations of the block.  But that is inconsistent with the behavior of passing arguments.

Maybe a better approach is to call the block in a begin block and if it doesn't raise an ExpectationNotMetError then it passes.  Something like:

```ruby
@matching_jobs_count = in_block_jobs.count do |job|
  if serialized_attributes.all? { |key, value| value == job[key] }
    block_passes = true
    begin
      @block.call(*job[:args])
    rescue RSpec::Expectations::ExpectationNotMetError
      block_passes = false
    end
    block_passes
  else
    false
  end
end
```

That would keep the behavior of both cases consistent.  The behavior of passing arguments is already to tell you that the count of matched enqueued jobs did not match what is expected and it does not tell you what differed from the jobs that were enqueued, so no information is lost by suppressing the ExpectationNotMetError when passing a block versus passing arguments.

What do you think?

morgoth:
@jschroeder9000 I'm trying to solve the same issue in https://github.com/rspec/rspec-rails/pull/1567 - What do you think about 2 solutions presented there?

samphippen:
I think this implementation is fine, and @jschroeder9000's explanation makes sense for it. @JonRowe do you think we can just merge this?

morgoth:
Correct me if I'm wrong, but this solution still doesn't fix a problem with case when you want to match on dynamic values created inside job, like:

```ruby
expect {
  user = CreateUser.call
}.to have_enqueued_job(DeliverWelcomeEmailJob).with { |arg|
  expect(arg).to eq(user)
}
```

hughevans:
Maybe I’m a noob, but I expected this to work with a `do end` block too. Fails silently when you do that.

JonRowe:
@hughevans `do end` blocks have a different priority to `{ ... }` blocks, which causes that, we do pass those blocks back to match on the individual matchers, but its tricky to figure out where to apply it, it would be possible to modify this matcher to warn in that scenario if someone has the time.

pirj:
Pull request is indeed welcome, @jakedouglas :+1:

PR #1580, opened 2016-03-20
Delegate properly to custom Resolver instances

bquorning:
Let `EmptyTemplateResolver` create instances of either `EmptyTemplateResolver::ResolverDecorator` or `EmptyTemplateResolver::FileSystemResolver`, depending on whether the given path argument is an `ActionView::Resolver` instance or not.

The new `EmptyTemplateResolver::ResolverDecorator` class simply delegates all method calls to its given resolver, except for `#find_all` and `#find_all_anywhere` which it overrides to return templates that render with `EmptyTemplateHandler`.

Supersedes #1568 and hopefully also #1557.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,11 @@ else
   gem 'rake', '~> 10.0' # rake 11 requires Ruby 1.9.3 or later
 end
 
+# Version 3 of mime-types 3 requires Ruby 2.0
+if RUBY_VERSION < '2.0.0'
+  gem 'mime-types', '< 3'
+end
+
 # Capybara versions that support RSpec 3 only support RUBY_VERSION >= 1.9.3
 if RUBY_VERSION >= '1.9.3'
   gem 'capybara', '~> 2.2.0', :require => false
--- a/example_app_generator/spec/verify_custom_renderers_spec.rb
+++ b/example_app_generator/spec/verify_custom_renderers_spec.rb
@@ -1,0 +1,159 @@
+require 'rails_helper'
+
+RSpec.describe "template rendering", :type => :controller do
+  context "without render_views" do
+    context "with the standard renderers" do
+      controller do
+        def index
+          render :template => 'foo', :layout => false
+        end
+      end
+
+      it "renders the 'foo' template" do
+        get :index
+
+        expect(response).to render_template(:foo)
+      end
+
+      it "renders an empty string" do
+        get :index
+
+        expect(response.body).to eq("")
+      end
+    end
+
+    context "with a String path prepended to the view path" do
+      controller do
+        def index
+          prepend_view_path('app/views/some_templates')
+
+          render :template => 'bar', :layout => false
+        end
+      end
+
+      it "renders the 'bar' template" do
+        get :index
+
+        expect(response).to render_template(:bar)
+      end
+
+      it "renders an empty string" do
+        get :index
+
+        expect(response.body).to eq("")
+      end
+    end
+
+    context "with a custom renderer prepended to the view path" do
+      controller do
+        def index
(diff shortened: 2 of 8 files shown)


bquorning:
Example of code that this PR will fix: https://gist.github.com/bquorning/6bf534e4b145b1738bb2. It’s a bit contrived, since I didn’t want to create tables and fetch custom templates from the database.

bquorning:
cc @JonRowe and @pixeltrix

JonRowe:
Needs specs demonstrating what this actually solves vs #1557, in fact ideally we need smoke specs for both "rendering" views and "not rendering" views via this code path.

bquorning:
See my comment https://github.com/rspec/rspec-rails/pull/1557#discussion_r55332474 for why #1557 doesn’t work.

I agree that it would be nice to have specs for this (as we discussed in #1551), but they are pretty hard to write. Any suggestions are welcome.

JonRowe:
> Any suggestions are welcome.

Take the scenario you posted that supposedly doesn't work and put it into one of the smoke test apps.

bquorning:
I added a smoke spec that passes for this branch, and fails on both master and #1557. Suggestions for cleaning up the spec are much welcome.

Gemfile
```diff
@@ -25,6 +25,11 @@ else
   gem 'rake', '~> 10.0' # rake 11 requires Ruby 1.9.3 or later
 end
 
+# Version 3 of mime-types 3 requires Ruby 2.0
+if RUBY_VERSION < '2.0.0'
+  gem 'mime-types', '< 3'
+end
```
bquorning: https://travis-ci.org/rspec/rspec-rails/builds/118239747 failed to bundle because [v2.6.4 of `mail` is out](https://rubygems.org/gems/mail/versions/2.6.4), which allows bundling with `mime-types < 4` instead of, as earlier, `mime-types < 3`.

bquorning:
@JonRowe, @pixeltrix This PR has a green build and is ready for review.

example_app_generator/spec/verify_custom_renderers_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+        render :template => 'baz'
+      end
+    end
+
+    it "renders the 'baz' template" do
+      get :index
+
+      expect(response).to render_template(:baz)
+    end
```
JonRowe: Great start, we just need to check that the views are empty (as `render_views` hasn't been used) and ideally have a second set of tests that checks that the views can be rendered when specified. Appreciate your work to get this towards the line!

bquorning:
@JonRowe You were right, there was a bug in the view rendering code (inside `ResolverDecorator`, see the last commit).

How do the added specs look?

(I’ll need to squash the commits before the PR is mergeable)

lib/rspec/rails/view_rendering.rb
```diff
@@ -57,6 +60,44 @@ def find_templates(*args)
...
+        #
+        # @private
+        class ResolverDecorator
+          def initialize(resolver)
+            @resolver = resolver
+          end
+
+          def method_missing(name, *args, &block)
+            result = @resolver.send(name, *args, &block)
+            nullify_templates(result)
```
bquorning: I realize that inspecting return values’ type isn't particularly good design. The alternative is to override `#find_all` and `#find_all_anywhere` depending on Rails version – as far as I recall `#find_all` changed arity from 5 to 6 at some point.

bquorning:
@JonRowe We have specs, a green build and squashed commits. Wanna review?

JonRowe:
:+1: Thanks

PR #1585, opened 2016-03-22
WIP: Add --skip-listen flag when generating Rails apps

bquorning:
Some specs fail when Rails is using the 'listen' gem, added in https://github.com/rails/rails/commit/de6ad5665d2679944a9ee9407826ba88395a1003. The `--skip-listen` flag option was added in https://github.com/rails/rails/commit/94dbc48887bf39c241ee2ce1741ee680d773f202.


.travis.yml
```diff
@@ -109,8 +109,3 @@ matrix:
     - rvm: 2.3.0
       env: RAILS_VERSION=master
   fast_finish: true
-
-branches:
-  only:
-    - master
-    - /^\d+-\d+-maintenance$/
```
JonRowe: This should be left in, Travis triggers builds in two ways, by pushes to branches, and by pushing to PRs, without this we build twice on those PRs

bquorning: I know. I’m just trying to get a green build. I just added a "WIP" to the PR title.

JonRowe: You don't need to remove this to get a green build, as the PR will trigger that

bquorning: Only when merge target is `master`.

JonRowe: Ah I missed that detail, however you'd be better off targeting against master ;)

PR #1587, opened 2016-03-25
Add support for file fixture available in Rails 5

morgoth:


samphippen:
@morgoth can you rebase this against master and see if it starts passing?

morgoth:
@samphippen It fails rubocop check - now method has too many lines. Not sure what to do about it?

morgoth:
@samphippen I fixed rails5 failures, so now it only fails by rubocop check.
Should I disable this check for this method, change global rubocop config or do you have better idea?

JonRowe:
@morgoth you can disable that check for that method.

morgoth:
The build is green now :)

spec/rspec/rails/configuration_spec.rb
```diff
@@ -252,4 +252,10 @@ def in_inferring_type_from_location_environment
       expect(group.new).to be_a(RSpec::Rails::MailerExampleGroup)
     end
   end
+
+  if ::Rails::VERSION::STRING > '5'
+    it "#file_fixture_path defaults to 'spec/fixtures/files'" do
```
samphippen: @morgoth this it statement doesn't scan cleanly.

Can you change it to `it "has a default #file_fixture_path of 'spec/fixtures/files'"`?

features/step_definitions/additional_cli_steps.rb
```diff
@@ -15,3 +16,9 @@
     pending "ActiveJob is not available"
   end
 end
+
+Given /file fixture is available/ do
```
samphippen: Could you change this to `/file fixtures are available/`

samphippen:
Few grammar comments, if you could patch those up, I'll merge this in.

PR #1573, opened 2016-03-11
Bring RSpec current with rails 5 beta 3

samphippen:
Biggest issue here is a regression introduced in the rails controller
testing gem which forces us to do a require of
`action_controller/metal/rendering` in order to not give a NameError
when booting our test app.

PR #1589, opened 2016-03-28
Add --skip-listen flag when generating Rails apps

bquorning:
Some specs fail when Rails is using the 'listen' gem, added in https://github.com/rails/rails/commit/de6ad5665d2679944a9ee9407826ba88395a1003.

The `--skip-listen` flag option was added in https://github.com/rails/rails/commit/94dbc48887bf39c241ee2ce1741ee680d773f202.

With this PR, we’ll get a green build for Rails master. Trying to get a green build for v5.0.0.beta3 (see #1573) isn’t worth the effort, in my opinion.


bquorning:
cc/ @samphippen

samphippen:
LGTM.

PR #1593, opened 2016-03-29
Add Mailer suffix for rails 5

koenpunt:
This is the same fix as in the mailer template generator


samphippen:
@koenpunt what was failing before this commit? What doesn't fail after this commit?

koenpunt:
It's an extension on what you did here: https://github.com/rspec/rspec-rails/commit/7a13bf15a191add538c5293c18c7a8dd3fe96e12
So without this change, the mailer preview class wouldn't work because the mailer class without the `Mailer` suffix doesn't exist.

I don't believe there's a test for this file, but maybe there should be?

samphippen:
@koenpunt if you have the time, it'd be great if we could get a spec for this.

samphippen:
@koenpunt this was merged as part of #1573

koenpunt:
:+1:

PR #1598, opened 2016-04-14
Setup and teardown supports blocks

### LINKED ISSUE #1597: support setup/teardown with blocks
In many places rails call setup/teardown with blocks and not symbols. 
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/testing/setup_and_teardown.rb#L29

One example is recent fix for not rolling back transactions on tests:
https://github.com/rails/rails/pull/24500/files#diff-18d566ee525de376b1a3741881545631R172

This fix doesn't works with rspec rails as we're ignoring blocks passed to setup/teardown.
https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/adapters.rb#L160

mfazekas:
See ActiveSupport::Testing
Fixes #1597

### DIFF
--- a/spec/rspec/rails/setup_and_teardown_adapter_spec.rb
+++ b/spec/rspec/rails/setup_and_teardown_adapter_spec.rb
@@ -10,9 +10,11 @@ describe RSpec::Rails::SetupAndTeardownAdapter do
       end
       expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "foo" }
       expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "bar" }
+      expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "baz" }
 
       klass.setup :foo
       klass.setup :bar
+      klass.setup { "baz" }
     end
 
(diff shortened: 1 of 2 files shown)


spec/rspec/rails/setup_and_teardown_adapter_spec.rb
```diff
@@ -10,9 +10,11 @@ def self.bar; "bar"; end
       end
       expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "foo" }
       expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "bar" }
+      expect(klass).to receive(:before).ordered { |&block| expect(block.call).to eq "baz" }
 
       klass.setup :foo
       klass.setup :bar
+      klass.setup { "baz" }
```
JonRowe: Can you add a test for tear down too, will help prevent regressions

mfazekas: added simple tests for teardown too

JonRowe:
Thanks @mfazekas :)

PR #1601, opened 2016-04-15
Fix #1600: Update gemspec to handle dependencies that only exist in beta state

### LINKED ISSUE #1600: rails-5-stable has rspec-expectations (= 3.2.0.pre) as a dependency, and this version does not exist!
Hi all,

Noting the [above entry in the gemspec](https://github.com/rspec/rspec-rails/blob/rails-5-support/rspec-rails.gemspec#L33..L39) I took a look at rspec-expectations both @ rubygems.org and the repo on github, but the tagged release of `3.2.0.pre` does not exist anymore.  The error I'm getting is,

```
Bundler could not find compatible versions for gem "rspec-expectations":
  In Gemfile:
    rspec-rails (>= 0) ruby depends on
      rspec-expectations (= 3.2.0.pre) ruby

    rspec-expectations (>= 0) ruby
Could not find gem 'rspec-expectations (= 3.2.0.pre) ruby', which is required by gem 'rspec-rails (>= 0) ruby', in any
of the sources.
```

Looking at attempting to patch this.

bsodmike:
This accounts for dependencies that are still in their `~> .beta` state and fixes https://github.com/rspec/rspec-rails/issues/1600

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -31,14 +31,23 @@ Gem::Specification.new do |s|
     version_string << '!= 3.2.22.1'
   end
 
+  # Remove 'pre' designation from version tag.
+  rspec_rails_version = RSpec::Rails::Version::STRING \
+    .split('.')[0..1] \
+    .concat(['0']) \
+    .join('.')
+
   s.add_runtime_dependency %q<activesupport>, version_string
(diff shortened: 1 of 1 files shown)


JonRowe:
This isn't something we want to merge, the pre version is deliberate as it points to the version in our master codebases.

PR #1603, opened 2016-04-19
Adds keywords arguments in requests for controller specs

diogobenica:
Making a request without keywords arguments in specs will cause Rails to print this warning:

```
DEPRECATION WARNING: ActionController::TestCase HTTP request methods will accept only
keyword arguments in future Rails versions.

examples:

get :show, params: { id: 1 }, session: { user_id: 1 }
process :update, method: :post, params: { id: 1 }
```

### DIFF
--- a/lib/generators/rspec/scaffold/templates/controller_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/controller_spec.rb
@@ -41,7 +41,7 @@ RSpec.describe <%= controller_class_name %>Controller, <%= type_metatag(:control
   describe "GET #index" do
     it "assigns all <%= table_name.pluralize %> as @<%= table_name.pluralize %>" do
       <%= file_name %> = <%= class_name %>.create! valid_attributes
-      get :index, {}, valid_session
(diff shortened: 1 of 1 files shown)


JonRowe:
So all of these should be conditionally defined, Ruby 1.8.7 (which we support) doesn't support this syntax, whilst 1.9.3 is using the new style hash syntax to back port it :)

diogobenica:
@JonRowe I'm not sure if Rails 3.2 accept this, but leaving `:params => { :id => 1 }` is an alternative to conditionals?

JonRowe:
That won't work as keyword aeguments

samphippen:
@diogobenica I agree with @JonRowe about the conditionals. Could you carry this forward with conditional definitions?

diogobenica:
@JonRowe @samphippen Sorry for taking too long. I don't know a better way to do this, without being repetitive. But now it seems all tests are passing, in all versions.

JonRowe:
Perfect, thanks!

PR #1623, opened 2016-05-18
Make the have_rendered but was redirect error message clearer

### LINKED ISSUE #1590: undefined method `+' for nil:NilClass in have_rendered
I have the following spec:

```
    it "should complain about a missing template" do
      get :show, template: 'cdcdcdcdc'
      expect(response).to render_template(:file => "#{Rails.root}/public/404.html")
    end
```
We're in the process of upgrading from Rail 3 to Rails 4. Currently we're on Rails 4.0.13
This spec passed under Rails 3, but is failing under Rails 4 with the following error:
`undefined method '+' for nil:NilClass` at the following line: `rspec-rails-3.4.2/lib/rspec/rails/matchers/have_rendered.rb:39:in 'failure_message'`

JonRowe:
In some circumstances (#1590), the error message for the `have_rendered` matcher produces unexpected results, and can cause a `NoMethodError`. This is a simple fix for that which attempts to keep the same style of message, it's not going to be perfect but it should at least not error.

Fixes #1590

### DIFF
--- a/lib/rspec/rails/matchers/have_rendered.rb
+++ b/lib/rspec/rails/matchers/have_rendered.rb
@@ -35,8 +35,8 @@ module RSpec
           # @api private
           def failure_message
             if @redirect_is
-              rescued_exception.message[/.* but /] +
-                "was a redirect to <#{@redirect_is}>"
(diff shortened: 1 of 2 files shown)


samphippen:
👍

PR #1635, opened 2016-06-10
Reorganise matchers

### LINKED ISSUE #1488: have_enqueued_job not available by default in 3.4.0
When I try to use new matcher, I'm getting `You must pass an argument rather than a block to use the provided matcher (have enqueued job), or the matcher must implement supports_block_expectations?`
error.

After debugging it turns out that it's not autoloaded, I have to `include RSpec::Rails::Matchers`.

Any idea why is that?

JonRowe:
An attempt to fix #1488 by bringing the general matchers into "always included" but not the few very specific rails matchers.

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -47,6 +47,7 @@ module RSpec
       config.include RSpec::Rails::RoutingExampleGroup,    :type => :routing
       config.include RSpec::Rails::ViewExampleGroup,       :type => :view
       config.include RSpec::Rails::FeatureExampleGroup,    :type => :feature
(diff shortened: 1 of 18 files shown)


samphippen:
LGTM.

JonRowe:
Bah, realised this is a semver violation, I'll rework it.

samphippen:
@JonRowe isn't everything that you merged marked api privatE?

JonRowe:
Yes apart from `RSpec::Rails::Matchers` which I'd removed some classes from.

Senjai:
@JonRowe Thank you for poking at this. Much appreciated man.

PR #1636, opened 2016-06-11
Skip mocks callback when AR improperly defined.

### LINKED ISSUE #1611: Uninitialized constant ActiveRecord::Base
On a rails 4.2.6 app that doesn't require 'active_record/railtie' the block `RSpec::Mocks.configuration.when_declaring_verifying_double` defined from the `rspec/rails/active_record.rb` file raises and `Unitialized constant ActiveRecord::Base`

I would like to collaborate on solving this problem but I'm not sure what is the proper solution:

```ruby
defined?(ActiveRecord::Base) # instead just defined?(ActiveRecord)
```

And a suggestion in how this can be tested if a test is needed will be really appreaciated

JonRowe:
In the case ActiveRecord has been improperly defined skip the call back for defining methods on mocks.

Fixes #1611

### DIFF
--- a/lib/rspec/rails/active_record.rb
+++ b/lib/rspec/rails/active_record.rb
@@ -7,7 +7,7 @@ module RSpec
       def self.initialize_activerecord_configuration(config)
         config.before :suite do
           # This allows dynamic columns etc to be used on ActiveRecord models when creating instance_doubles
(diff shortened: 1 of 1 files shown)


samphippen:
LGTM.

PR #1638, opened 2016-06-14
include general matchers in all contexts when rspec-rails is present

JonRowe:
Because the matcher methods are public, and the module is public, we can't extract matchers easily without duplication. Given the original problem was simply that certain matchers (active job) weren't available outside Rails specific contexts, it seems safe just to included the matchers module generally without worrying about it. 

/cc @samphippen


JonRowe:
Be good to get this in for 3.5 @samphippen it makes the jobs matchers available everywhere

samphippen:
👍

PR #1640, opened 2016-06-15
Fix typo in changelog

aried3r:
Would you also be open to have `Changelog.md` be renamed to `CHANGELOG.md`?

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -50,7 +50,7 @@ Bug fixes:
 
 Enhancements:
 
-* Add a `---singularize` option for the feature spec generator (Felicity McCabe,
+* Add a `--singularize` option for the feature spec generator (Felicity McCabe,
   #1503)
 * Prevent leaking TestUnit methods in Rails 4+ (Fernando Seror Garcia, #1512)
 * Add support for Rails 5 (Sam Phippen, #1492)


myronmarston:
Thanks!

> Would you also be open to have `Changelog.md` be renamed to `CHANGELOG.md`?

If someone provided a compelling reason for doing so, we'd be open to it, but bear in mind that:

* It's called `Changelog.md` in each of the rspec repos and we want them to be consistent.
* Our tooling looks for `Changelog.md`, not `CHANGELOG.md` (see [here](https://github.com/rspec/rspec-rails/blob/acca723ca53d64e5ca699734fe296d4e34a3a2bf/Rakefile#L119) and [here](https://github.com/rspec/rspec-dev/blob/7b5cc25dc629164c5b6509870ef868d83d27f9e2/Rakefile#L375) for examples, and there are others) so changing the name would break these things on case-sensitive operating systems--meaning we'd have to update all of them.
* It appears that Github treats file URLs in a case-sensitive manner (i.e. https://github.com/rspec/rspec-rspec/blob/master/CHANGELOG.md 404s) so changing the cases would create broken links for anyone with a link to the existing file.

So I'd say that the reason would have to be _extremely_ compelling to cause us to go through that pain.

aried3r:
> If someone provided a compelling reason for doing so

I can't provide that, hah! My point would be "I didn't find it very easily because I'm looking for `CHANGELOG.md` and it wasn't there.", so... not very objective, indeed the exact opposite, but I could imagine it being the case for a few people.

While your points are all good, I still want to try to address them:

> It's called Changelog.md in each of the rspec repos and we want them to be consistent.

Well, renaming isn't that much of a problem.

> Our tooling looks for Changelog.md, not CHANGELOG.md

Also not a problem. It's basically `s/Changelog.md/CHANGELOG.md/`.

> It appears that Github treats file URLs in a case-sensitive manner (i.e. https://github.com/rspec/rspec-rspec/blob/master/CHANGELOG.md 404s) so changing the cases would create broken links for anyone with a link to the existing file.

Well, that's true. And there's two approaches, the first being: Don't care. Either update the links or point to a specific revision. That's not very nice to the others, but in the end, it's just links, we don't necessarily break a lot of important stuff.
The second would be to have `Changelog.md` link to `CHANGELOG.markdown`. Which is... well, confusing, for people who actually browse the repo and see two change logs.

That being said, I don't feel like this change is very important or necessary at all, but I'd love it if it were to happen. :)

myronmarston:
Closing. I don't see enough benefit for this and no one else has asked. We have lots of external links to the change log (e.g. From blog posts).

PR #1647, opened 2016-06-29
Fix the indentation of scaffold controller spec

y4ashida:
The controller specs generated by scaffold command have inconsistent indentation. Like this:
```rb
  describe "GET #show" do
    it "assigns the requested user as @user" do
      user = User.create! valid_attributes
          get :show, params: {id: user.to_param}, session: valid_session
          expect(assigns(:user)).to eq(user)
    end
  end
```


JonRowe:
Seems reasonable to me, thanks :)

PR #1651, opened 2016-07-04
Set RAILS_ENV=test early in rake execution

bitherder:
This hack mirrors a hack in Rails for test::unit.
(See https://github.com/rails/rails/blob/99873ca1ea98d73c73f8be60dfce0a54224f4ea8/railties/lib/rails/test_unit/railtie.rb)

In my particular situation (general issue for XING), this change
prevents a problem with the use of the dotenv gem where environment
variable values for development leak into the test environment.
(See https://github.com/bkeepers/dotenv/issues/219 and https://github.com/bkeepers/dotenv/pull/241)

### DIFF
--- a/lib/rspec-rails.rb
+++ b/lib/rspec-rails.rb
@@ -2,6 +2,11 @@
 require 'rspec/core'
 require 'rspec/rails/feature_check'
 
+# Prevent problems with double loading between `development` and `test` environments
+if defined?(Rake.application) && Rake.application.top_level_tasks.grep(/^(default$|spec(:|$))/).any?
+  ENV['RAILS_ENV'] ||= 'test'
+end
+
 # Namespace for all core RSpec projects.
 module RSpec
   # Namespace for rspec-rails code.


JonRowe:
In general we recommend people do this in their `spec_helper.rb` and then load that as soon as possible, giving you more control. Can I ask why this doesn't work for you?

bitherder:
I'd have to investigate it putting code in the `spec_helper.rb` would avoid the problem with the dotenv gem.

We have 20+ rails projects using rspec and we are converting them all to use the dotenv gem. We are adding a patch to all the projects to deal with this problem, but it would be better (for us) if there was a centralized way of dealing this issue so we can pull the per-project code out.

JonRowe:
> We have 20+ rails projects using rspec and we are converting them all to use the dotenv gem. We are adding a patch to all the projects to deal with this problem, but it would be better (for us) if there was a centralized way of dealing this issue so we can pull the per-project code out.

The thing is the default RSpec Rails install doesn't need this patch, we do:

```Ruby
ENV['RAILS_ENV'] ||= 'test'
# Prevent database truncation if the environment is production
abort("The Rails environment is running in production mode!") if Rails.env.production?
require 'spec_helper'
require 'rspec/rails'
# ... more defaults
```

Which means this RAILS_ENV is already test before this is ever reached, so given this code's added complexity I'm declined to close this.

PR #1660, opened 2016-07-07
Don't swallow exceptions for flow control.  fixes #1656

betesh:
See discussion on #1656

### DIFF
--- a/lib/rspec/rails/example/request_example_group.rb
+++ b/lib/rspec/rails/example/request_example_group.rb
@@ -11,10 +11,8 @@ module RSpec
       include RSpec::Rails::Matchers::RenderTemplate
       include ActionController::TemplateAssertions
 
-      begin
+      if ActionPack::VERSION::MAJOR >= 5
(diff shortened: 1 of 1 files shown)


myronmarston:
LGTM.

I wonder if there are any other places where we use an exception approach that could switch to a version check approach?

betesh:
Can we get this released as 3.5.1?  That would allow me to use it with rspec 3.5.0

myronmarston:
> I believe @myronmarston is planning a release of all the gems shortly, would you mind waiting a day or so? We like to bundle together a few fixes at a time :)

I released rspec-core 3.5.1 last night, and didn't have any more releases planned. No need to wait for this.

JonRowe:
My mistake, I assumed we were bumping them all, I can handle the release if you'd like?

myronmarston:
There was nothing new to release in the others, and I usually release patch versions freely when needed.

> I can handle the release if you'd like?

Go for it!

betesh:
Thank you!

PR #1661, opened 2016-07-07
Rake notes compatibility in Rails 5.1.0

johnmeehan:
As of Rails 5.1.0 gems can register a directories to work with `rake notes`

in this case we can tell Rails it to now look in the `/spec` folder when searching for annotations like `# TODO:` `# FIXME` etc.

This was added in this PR:
rails/rails#25692

Resulting in:

```bash

$ rake notes
app/spec/models/article_spec.rb:
  * [2] [TODO] add more specs
  * [3] [FIXME] Spec A is broken
  * [4] [OPTIMIZE] improve this test

```

Thats one less daily annoyance. 

John

### DIFF
--- a/lib/rspec-rails.rb
+++ b/lib/rspec-rails.rb
@@ -8,12 +8,10 @@ module RSpec
   module Rails
     # Railtie to hook into Rails.
     class Railtie < ::Rails::Railtie
-
       # As of Rails 5.1.0 you can register directories to work with `rake notes`
-      if Gem::Requirement.new(">= 5.1.0").satisfied_by?(Gem::Version.new(::Rails.version))
+      if ::Rails::VERSION::STRING > '5.1'
         SourceAnnotationExtractor::Annotation.register_directories("spec")
       end
-      
       # Rails-3.0.1 requires config.app_generators instead of 3.0.0's config.generators
       generators = config.respond_to?(:app_generators) ? config.app_generators : config.generators
       generators.integration_tool :rspec
@@ -73,7 +71,6 @@ module RSpec
         # always return `true`.
         config.respond_to?(:action_mailer) && ::Rails::VERSION::STRING > '4.1'
       end
-
     end
   end
 end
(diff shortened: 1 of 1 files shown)


lib/rspec-rails.rb
```diff
@@ -8,6 +8,12 @@ module RSpec
   module Rails
     # Railtie to hook into Rails.
     class Railtie < ::Rails::Railtie
+
+      # As of Rails 5.1.0 you can register directories to work with `rake notes`
+      if Gem::Requirement.new(">= 5.1.0").satisfied_by?(Gem::Version.new(::Rails.version))
```
JonRowe: Could you switch this away from using rubygems please? It's generally a good idea to not assume it's present, a check on the rails version string is sufficient normally (e.g. ::Rails.version >= "5.1.0").

lib/rspec-rails.rb
```diff
@@ -8,6 +8,12 @@ module RSpec
...
+        SourceAnnotationExtractor::Annotation.register_directories("spec")
+      end
+
```
JonRowe: The build is breaking due to the whitespace inserted here, our convention is to not allow trailing whitespace in files, would you mind cleaning it up?

lib/rspec-rails.rb
```diff
@@ -67,6 +73,7 @@ def supports_action_mailer_previews?(config)
         # always return `true`.
         config.respond_to?(:action_mailer) && ::Rails::VERSION::STRING > '4.1'
       end
+
```
JonRowe: Would you remind removing this? Adding extra spacing to existing code tends to cause git churn which we would like to avoid.

lib/rspec-rails.rb
JonRowe: You can see our preferred form a few lines down in fact, so making this `::Rails::VERSION::STRING >= '5.1'` would be preferred :)

JonRowe:
Thanks :heart:

PR #1676, opened 2016-07-22
Expand documentation for anonymous controllers

### LINKED ISSUE #1670: Anonymous controller cannot define a custom route for custom controller
**TL;DR** Renaming the controller under test from `ApplicationController` to `CustomController` breaks the example.

Given a fresh Rails 4.2.6 application, [the anonymous controller code sample provided by Relish for RSpec 3.5](https://www.relishapp.com/rspec/rspec-rails/docs/controller-specs/anonymous-controller):

```ruby
require "rails_helper"

RSpec.describe ApplicationController, :type => :controller do
  controller do
    def custom
      render :text => "custom called"
    end
  end

  specify "manually draw the route to request a custom action" do
    routes.draw { get "custom" => "anonymous#custom" }

    get :custom
    expect(response.body).to eq "custom called"
  end
end
```

...this passes. However, if you do the same example, but from a controller that is not `ApplicationController`, such as `CustomController` defined in `custom_controller.rb`:

```ruby
class CustomController < ActionController::Base
  # Prevent CSRF attacks by raising an exception.
  # For APIs, you may want to use :null_session instead.
  protect_from_forgery with: :exception
end
```

Then we'd expect the same example to pass:

```ruby
require "rails_helper"

RSpec.describe CustomController, :type => :controller do
  controller do
    def custom
      render :text => "custom called"
    end
  end

  specify "manually draw the route to request a custom action" do
    routes.draw { get "custom" => "anonymous#custom" }

    get :custom
    expect(response.body).to eq "custom called"
  end
end
```

But it fails with:

```
Failures:

  1) CustomController manually draw the route to request a custom action
     Failure/Error: get :custom
     
     ActionController::UrlGenerationError:
       No route matches {:action=>"custom", :controller=>"custom"}
     # ./spec/controllers/custom_controller_spec.rb:13:in `block (2 levels) in <top (required)>'
```

Am I doing something obviously wrong? Or is this broken?

JonRowe:
anonymous controllers that dont inherit from ActionController::Base need special custom routes, fixes #1670

### DIFF
--- a/features/controller_specs/anonymous_controller.feature
+++ b/features/controller_specs/anonymous_controller.feature
@@ -470,6 +470,31 @@ Feature: anonymous controller
     When I run `rspec spec`
     Then the examples should all pass
 
+  Scenario: Draw custom routes for anonymous controllers which don't inherit from application controller
+    Given a file named "spec/controllers/other_controller_spec.rb" with:
+      """ruby
+      require "rails_helper"
+      class OtherController < ActionController::Base
+      end
+
+      RSpec.describe OtherController, :type => :controller do
+        controller do
+          def custom
+            render :text => "custom called"
+          end
+        end
+
+        specify "manually draw the route to request a custom action" do
+          routes.draw { get "custom" => "other#custom" }
+
+          get :custom
+          expect(response.body).to eq "custom called"
+        end
+      end
+      """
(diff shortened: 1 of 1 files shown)


samphippen:
LGTM

PR #1684, opened 2016-08-02
Fixed deserialization of passed arguments to with block in AJ matcher

morgoth:
Without it, one get arguments like `{"_aj_globalid"=>"gid://rspec-suite/GlobalIdModel/42"}` which is internal way of serialization for AJ arguments.

BTW, I didn't find a way to avoid adding `GlobalIdModel` in global namespace (it's needed for deserialization) - let me know if you have some idea how to avoid it.


JonRowe:
Thanks!

Issue #1672, opened 2016-07-20
skip generating specs for new/edit routes/controllers/views for API only app


audy:
RSpec currently generates specs for new and edit routes, controllers and views for a new Rails (5.0.0) app generated with `--api`, when running `rails g scaffold ...`. These tests fail and are unnecessary.

Can these be skipped? If not, would this be a useful feature to add to RSpec (automatically detecting an API-only app and not generating the unnecessary specs)?

## unnecessary specs

- [ ] view specs
- [ ] certain controller specs (new, edit)

JonRowe:
It would certainly make sense for the generator to be aware of `--api` and produce different specs accordingly, if you'd like to tackle that feel free!

AustinMCrane:
if there is anyway i could help with this let me know, this is an inconvenience and would be helpful by default when generating an api only app.

mrageh:
I'd like to go ahead and tackle this issue @JonRowe for api only apps are we not supposed to generate view specs?

audy:
@mrageh I started making a list of unnecessary specs in the original issue body.

AustinMCrane:
Some generated specs could be more helpful for testing against JSON responses or status codes as well, for instance you can replace the redirect checks with http status code checks and JSON formats

danrabinowitz:
Is this issue resolved by https://github.com/rspec/rspec-rails/pull/1685 ?

JonRowe:
I believe so

PR #1685, opened 2016-08-05
Update scaffold generator to care about --api flag.

k3rni:
* don't generate view specs
* don't generate route specs for new and edit
* generate a slightly different controller spec, without redirects but
  with JSON expectations

Addresses #1672


k3rni:
The CI failures seem to be of two kinds:

1. https://travis-ci.org/rspec/rspec-rails/jobs/150062031 - weird unrelated failure ?
2. https://travis-ci.org/rspec/rspec-rails/jobs/150062007 - Rubocop complaints. BTW it seems that Rubocop is only enabled for rubies below 2.2, is that intentional?

How should I proceed? Do you have suggestions on

JonRowe:
I've bumped the broken build, you'll need to fix the rubocop failure though.

> BTW it seems that Rubocop is only enabled for rubies below 2.2, is that intentional?

It's a side effect of the version we're running.

k3rni:
Per the linked issue, the easiest way to satisfy Rubocop here would be to just drop these methods. Is it okay, if that solution were to be accepted, to do that in this PR?

JonRowe:
If you're sure they're not used, the build should fail if they are...

samphippen:
@k3rni you'll need to update this for #1689 which removes assigns and assert templates from our specs. We definitely don't want those anymore. If you could do that, I can probably merge this down.

k3rni:
Rebasing to current master and updating.

k3rni:
I chose to add test for `response.location` returning the newly created object's url to `POST#create`. This is to match testing `redirect_to` in regular controllers.

lib/generators/rspec/scaffold/templates/api_controller_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+<% end -%>
+        expect(response).to have_http_status(:ok)
+        expect(response.content_type).to eq('application/json')
+      end
+    end
+
+    context "with invalid params" do
+      it "renders a JSON response with errors for the <%= ns_file_name %>" do
+<% if RUBY_VERSION < '1.9.3' -%>
+        <%= file_name %> = <%= class_name %>.create! valid_attributes
```
samphippen: @k3rni 

you missed moving this file name assignment up to outside of the 1.9.3 conditional, so this does not work on ruby interpreters greater than or equal to 1.9.3, can you add a spec for this case and make it pass? Thanks

samphippen:
@k3rni sorry about the slow review here, it's been a crazy week at work. I've left you a review note, but apart from that, I think this looks great and I'll merge it when you're done 👍  Thank you so much ✨

k3rni:
@samphippen so basically just move the assignment (line 140 above) up outside the conditional, and verify in tests, got it.

JonRowe:
Thanks :heart:

PR #1697, opened 2016-09-12
switch git from git to https

JonRowe:
as per rspec/rspec-core#2327


samphippen:
LGTM on 💚

mikegee:
This still looks legit 2 years later. The build issue related to ammeter pulling in a version of rails that's too new doesn't make a lot of sense. Perhaps a retry after rebasing would get bundler to behave there?

JonRowe:
A rebase made this green, good digging @mikegee

PR #1702, opened 2016-09-27
3 5 maintenance

Spencer8989:


JonRowe:
@Spencer8989 `3-5-maintenance` is the branch maintaining the `3.5.x` series of gems and is not intended to be merged into `master` which is currently supporting the development of the as yet unreleased `3.6.x` series of gems.

PR #1705, opened 2016-09-28
Some updates for the contributing file.

samphippen:


CONTRIBUTING.md
```diff
@@ -23,3 +23,28 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
 If you need help getting started, check out the [DEVELOPMENT](DEVELOPMENT.md) file for steps that will get you up and running.
 
 Thanks for helping us make RSpec better!
+
+## `Small` issues
+
+Small issues are the ones that we be believe are best suited for new
```
JonRowe: Personally I feel these lines would read better as `These issues ...` as its a bit repetitious at the moment.

CONTRIBUTING.md
```diff
@@ -23,3 +23,28 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+contributors to get started on. They represent a meaningful contribution to the
+project that can be made that should not be too hard to pull off.
```
JonRowe: This reads a bit funny to me, how about `They represent a potential meaningful contribution to the project that should not be too hard to pull off.`

CONTRIBUTING.md
```diff
@@ -23,3 +23,28 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+## `Triage` issues
+
+Triage issues are ones that have been labelled by the maintainers that we
+believe do not currently have enough information for them to be reproduced by
```
JonRowe: Suggestion: `currently have enough information to be reproduced by`

CONTRIBUTING.md
```diff
@@ -23,3 +23,28 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+us. You can help us move triage issues into `ready to go` or `has reproduction
+case` (for feature requests or bugs) as a seriously meaningful contribution.
+
+## `Has reproduction case` issues
+
+Issues that have reproduction cases have a repository that we can clone that
+enable us to quickly determine the issue is valid and then start debugging
+within RSpec. These issues are good ones to tackle to help us actively fix bugs.
```
JonRowe: Do we do this on other repos?

myronmarston:
We might consider creating an Issue template.  For example, here's Elixir's:

https://github.com/elixir-lang/elixir/blob/master/ISSUE_TEMPLATE.md

You can see how this looks when you create a new issue:

https://github.com/elixir-lang/elixir/issues/new

Thoughts?

CONTRIBUTING.md
samphippen: :+1:

CONTRIBUTING.md
samphippen: I've been doing this on RR for a while, not sure about the other repos.

samphippen:
@myronmarston would you propose that instead of these changes, or as an additional step?

myronmarston:
In addition to.  The changes here are fine, but I think it would make for fewer issues where we have to ask the user for more details if we give them a template to fill out that provides the details we need.

samphippen:
@JonRowe @myronmarston I've made some edits to contributing.md and added an issue template. Can I get another review pass?

CONTRIBUTING.md
```diff
@@ -23,3 +23,27 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
 If you need help getting started, check out the [DEVELOPMENT](DEVELOPMENT.md) file for steps that will get you up and running.
 
 Thanks for helping us make RSpec better!
+
+## `Small` issues
+
+These issues are the ones that we be believe are best suited for new
+contributors to get started on. They represent a potential meaningful
+contribution to the project that should not be too hard to pull off.
+## `Triage` issues
```
myronmarston: There should be a line break between the paragraph and the next section header.

CONTRIBUTING.md
```diff
@@ -23,3 +23,27 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+These issues are ones that have been labelled by the maintainers that we
+believe do not currently have enough information to be reproduced the RSpec
+team. You can help us move triage issues into `ready to go` or `has reproduction
+case` (for feature requests or bugs) as a seriously meaningful contribution.
```
myronmarston: The way this sentence is worded, it sounds like people can help us by changing the issue label. I think it would be better to state what is needed to get a triaged issue into one of these states.

CONTRIBUTING.md
myronmarston: The interactions between rspec and rails in rspec-rails and subtle enough that having a clonable repo is pretty important here.  For the other repos, it's much less so; often times the description from the user is sufficient, or a snippet from them.  Sometimes we can't reproduce and do ask for a repo.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,60 @@
...
+We prefer that the RSpec Rails issue tracker be mainly used for bug reports.
+Feature requests or general requests for help should go to our google group:
+rspec@googlegroups.com. We use that as our primary location for higher level
+RSpec discussions.
+
+If you are filing a bug report, there's a few steps it'd be super if you could
+follow. If you can't do all of these, don't worry, but if you can it'll help us
+resolve your bug faster!
+
+-->
```
myronmarston: Is the idea of the HTML comments that it won't render on the github UI if they are left in place?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Description of issue
```
myronmarston: I think instead of the description of the issue, we should ask for observed behavior and expected behavior.  Often times we get issues where users describe the problem, but the behavior is what we would expect.  Then we have to go back and forth with the user to get them to explain what about the behavior was unexpected.  Asking for them to explain observed vs expected behavior up front will hep, I think.

CONTRIBUTING.md
```diff
@@ -23,3 +23,29 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
 If you need help getting started, check out the [DEVELOPMENT](DEVELOPMENT.md) file for steps that will get you up and running.
 
 Thanks for helping us make RSpec better!
+
+## `Small` issues
```
ashleygwilliams: i would use literally anything other than `small`. 2 reasons:

- downplays the effort of new contributors
- doesn't properly signal that these issues should be reserved for new contributors

recommendations:  `your-first-PR`, `starter-issue`

i would also recommend a `PRs welcome` tag that flags that PRs are welcome, and then put another tag alongside it that flags that it would be good for a new contributor

another consideration: when you talk about "first" issues, do you mean first contribution to the rspec collection of software, OR first contribution ever? this will inform the name of the tag you use as well as the amount of scaffolding that you'll need to provide in the issue.

CONTRIBUTING.md
```diff
@@ -23,3 +23,29 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+These issues are the ones that we be believe are best suited for new
+contributors to get started on. They represent a potential meaningful
+contribution to the project that should not be too hard to pull off.
```
ashleygwilliams: i would considered adding/reframing this text based on how you answer the above issues/questions

CONTRIBUTING.md
```diff
@@ -23,3 +23,29 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+## `Triage` issues
```
ashleygwilliams: maybe this should be "need reproduction"- `triage` is a label i've seen on many repos as `triaged` which flags that it has *already* been triaged, and as a result may have the opposite effect.

CONTRIBUTING.md
```diff
@@ -23,3 +23,29 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+These issues are ones that have been labelled by the maintainers that we
+believe do not currently have enough information to be reproduced the RSpec
+team. While not directly counted by the GitHub contribution graph, we consider
+helping us triage issues to the point where they can be reproduced as an
+extremely meaningful contribution.
```
ashleygwilliams: you might add classic "how to triage" advice, e.g. what versions of X, Y, Z.

ashleygwilliams: ah, based on the subsequent section it sounds like you want a repo with a repro case in it for this. you should say that!

CONTRIBUTING.md
```diff
@@ -23,3 +23,29 @@ If you'd like to help make RSpec better, here are some ways you can contribute:
...
+
+## `Has reproduction case` issues
+
+Issues that have reproduction cases have a repository that we can clone that
+enable us to quickly determine the issue is valid and then start debugging
+within RSpec. These issues are good ones to tackle to help us actively fix bugs.
+
+## Maintenance branches
```
ashleygwilliams: this seems very different than the previous sections, you might consider putting this is a different section. it is a not a issue label so perhaps "dev env" or something like this is a better move. either way i'd get it out of this section.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,67 @@
...
+rspec@googlegroups.com. We use that as our primary location for higher level
+RSpec discussions.
+
+If you are filing a bug report, there's a few steps it'd be super if you could
+follow. If you can't do all of these, don't worry, but if you can it'll help us
+resolve your bug faster!
+
+-->
+
+## Observed behaviour
```
ashleygwilliams: before this question, you should ask "what were you trying to do"-

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,67 @@
...
+-->
+
+## Expected behaviour
+
+<!--
+Please provide a description of what you expect to be happening, and how that
+differs from the current behaviour.
+-->
+
+## Can you provide an example app?
```
ashleygwilliams: you want a link, so you should say link...

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,67 @@
...
+2. commit that app, so that we have the rails skeleton in a separate commit
+3. Make all the changes necessary (adding RSpec, models, specs, controllers,
+   etc) to reproduce the issue. There should be a failing test or tests that you
+   expect to pass. We'll run `bundle exec rspec`, but if you can provide a
+   readme with more detailed instructions that'd be amazing :)
+4. Commit
+5. Provide a link to a github repo, a description of the app and what you're expecting here
+-->
+
+## What Ruby, Rails and RSpec versions are you using?
```
ashleygwilliams: honestly i'd put this as the very first question and i'd consider making it 3 separate questions. people don't read and will only answer this partially which will be a bummer

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,67 @@
...
+bundle exec ruby --version
+bundle exec rails --version
+bundle exec rspec --version
+```
+
+and paste the output below
+
+-->
+
+## Did this problem exist before you upgraded to Rails 5?
```
ashleygwilliams: you could put checkboxes or radio boxes for yes/no here

ashleygwilliams: if you want a freeform answer you might widen this to "did you notice this bug after a specific update or event? please explain. e.g. updated to rails 5"

ISSUE_TEMPLATE.md
benoittgt: Isn't mention on line 45?

ISSUE_TEMPLATE.md
benoittgt: Yes

PR #1680, opened 2016-07-27
Remove skip statements from scaffold controller tests during smoke

JonRowe:

PR #1710, opened 2016-09-30
Use keyword args for HTTP methods in controller specs only on Rails 5

### LINKED ISSUE #1669: Scaffold generator generates invalid controller specs on versions prior to Rails 5
In #1603, we made scaffold generator to generate controller specs with HTTP request method (`#get`, `#post`, etc.) with keyword arguments:

``` diff
- post '/books', { book: { title: 'Foo' } }
+ post '/books', params: { book: { title: 'Foo' } }
```

... regardless of the current project's Rails version.

However the keyword arguments are actually supported only on Rails 5 (https://github.com/rails/rails/pull/18323). The CI in #1603 was passing because [they were skipped](https://github.com/rspec/rspec-rails/blob/v3.5.1/lib/generators/rspec/scaffold/templates/controller_spec.rb#L27-L33).

yujinakayama:
This closes #1669.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,11 @@ Enhancements:
 
 * Add compatibility for Rails 5.1. (Sam Phippen, Yuichiro Kaneko, #1790)
 
+Bug Fixes:
+
+* Fix scaffold generator so that it does not generate broken controller specs
+  on Rails 3.x and 4.x. (Yuji Nakayama, #1710)
+
 ### 3.6.0.beta2 / 2016-12-12
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.6.0.beta1...v3.6.0.beta2)
 
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -134,4 +134,9 @@ gsub_file 'spec/spec_helper.rb',
           'config.warnings = false'
 gsub_file '.rspec', '--warnings', ''
 
+# Remove skips so we can test controller specs work
+gsub_file 'spec/controllers/gadgets_controller_spec.rb',
+          'skip("Add a hash of attributes valid for your model")',
+          '{}'
+
 final_tasks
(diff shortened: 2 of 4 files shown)


JonRowe:
These aren't keyword arguments tho, they're keyword argument on Rails 5, but new style hash access on Rails 4, I'm still unsure why it breaks and this doesn't fix the fact that specs don't pick it up...

yujinakayama:
Should we replace the `skip` in the spec with empty hash?

```diff
  let(:valid_attributes) {
-   skip("Add a hash of attributes valid for your model")
+   {}
  }
```

JonRowe:
I tried that in #1680, (basically we don't want to remove them from peoples scaffolded specs, but we do want them to run in our specs) but it passes and I'm not sure it's working correctly

yujinakayama:
It seems that the scaffold specs are not executed in cukes since they're always removed before each scenario.

https://github.com/rspec/rspec-rails/blob/2175c72f42ebd40c946ccc45fe3a2ad0caa8df6f/features/support/env.rb#L37-L44

mrageh:
@yujinakayama and @JonRowe I think we should merge this PR, because in Rails 5 HTTP request methods use keyword args to process requests. Whereas any version below Rails 5 uses does not use keyword args. That means our scaffold generator will produce invalid controller specs.

For more context please refer to this [PR](https://github.com/rails/rails/pull/18323/).

JonRowe:
@mrageh as per my earlier comments this shouldn't be merged, Rails 5 swapped option hashes for keyword args, the api is supposed to be compatible without any changes to the user

JonRowe:
@mrageh However if you'd like to work on getting our generated specs being tested as part of the build I'd welcome the help in a seperate PR.

yujinakayama:
@JonRowe

> the api is supposed to be compatible without any changes to the user

No, `get '/', params: {}` does not work on Rails 4.2. Try the reproducible example in https://github.com/rspec/rspec-rails/issues/1669#issuecomment-239619487.

mrageh:
@JonRowe I may have missed something, but I'm fairly certain the Rails 5 api change is not backwards compatible. Can you send me some docs/github discussion anything that says otherwise?

yujinakayama:
This should be rebased once https://github.com/rspec/rspec-rails/pull/1680 is merged.

PR #1713, opened 2016-10-02
Force railties to be the same version as rails for maintenance branches

JonRowe:
This is an attempt to fix the build but forcing railties to be the same version as rails


JonRowe:
Hurrah! A wild green build appears, anyone care to review @rspec/owners ?

samphippen:
LGTM

PR #1719, opened 2016-10-04
split spec bootstapping into separate files

maxcal:
This splits the spec bootstrapping process into  spec_helper.rb, rails_helper.rb and generator_helper.rb to avoid loading the entire rails stack when not needed.

I mostly picked the low hanging fruit - further optimizations could likely be gained by only loading the needed parts of the rails stack with `config.include_context` and loading files explicitly from spec/support.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,2 @@
 --warnings
 --color
---require spec_helper
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --warnings
 --color
+--require spec_helper
--- a/spec/generators/rspec/controller/controller_generator_spec.rb
+++ b/spec/generators/rspec/controller/controller_generator_spec.rb
@@ -2,7 +2,6 @@ require 'generator_helper'
 # Generators are not automatically loaded by Rails
 require 'generators/rspec/controller/controller_generator'
 
-
 RSpec.describe Rspec::Generators::ControllerGenerator, :type => :generator do
   setup_default_destination
 
--- a/spec/generators/rspec/controller/controller_generator_spec.rb
+++ b/spec/generators/rspec/controller/controller_generator_spec.rb
@@ -1,6 +1,7 @@
+require 'generator_helper'
 # Generators are not automatically loaded by Rails
 require 'generators/rspec/controller/controller_generator'
-require 'support/generators'
+
 
 RSpec.describe Rspec::Generators::ControllerGenerator, :type => :generator do
   setup_default_destination
--- a/spec/rspec/rails/assertion_delegator_spec.rb
+++ b/spec/rspec/rails/assertion_delegator_spec.rb
@@ -1,4 +1,3 @@
-require 'spec_helper'
 
 describe RSpec::Rails::AssertionDelegator do
   it "provides a module that delegates assertion methods to an isolated class" do
--- a/spec/rspec/rails/assertion_delegator_spec.rb
+++ b/spec/rspec/rails/assertion_delegator_spec.rb
@@ -1,4 +1,4 @@
-require "spec_helper"
+require 'spec_helper'
 
 describe RSpec::Rails::AssertionDelegator do
   it "provides a module that delegates assertion methods to an isolated class" do
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,4 +1,4 @@
-
+# This file should only contain the bare minimum required to configure the spec suite
 require 'rspec/support/spec'
 
 class RSpec::Core::ExampleGroup
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,17 +1,5 @@
-require 'rails/all'
-
-module RSpecRails
-  class Application < ::Rails::Application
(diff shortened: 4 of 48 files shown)


.rspec
```diff
@@ -1,3 +1,2 @@
 --warnings
 --color
---require spec_helper
```
JonRowe: Please add this back, it's our project standard to require this here and not in spec files.

spec/generators/rspec/controller/controller_generator_spec.rb
```diff
@@ -1,6 +1,7 @@
+require 'generator_helper'
 # Generators are not automatically loaded by Rails
 require 'generators/rspec/controller/controller_generator'
-require 'support/generators'
+
```
JonRowe: Please remove the excess whitespace introduced here and in similar places

spec/rspec/rails/assertion_delegator_spec.rb
```diff
@@ -1,4 +1,4 @@
-require "spec_helper"
+require 'spec_helper'
```
JonRowe: This can be removed as `spec_helper` should be required by `.rspec` otherwise please don't change our quotes we don't standardise them and would prefer less git churn.

spec/spec_helper.rb
```diff
@@ -1,17 +1,5 @@
...
-
-module RSpecRails
-  class Application < ::Rails::Application
-    self.config.secret_key_base = 'ASecretString' if config.respond_to? :secret_key_base
-  end
-end
-I18n.enforce_available_locales = true if I18n.respond_to?(:enforce_available_locales)
-
-require 'rspec/support/spec'
 require 'rspec/rails'
```
JonRowe: Theres no point splitting the helpers if you leave this here.

.rspec
maxcal: Fair enough - the point was really to be explicit about which bootstrapper is used for a given spec.

JonRowe: Our convention is that `spec_helper` contains the bare minimum required to configure the spec suite and everything else assumes its presence

maxcal:
Changes done

JonRowe:
Does this run locally for you? Seems to be failing significantly on CI

maxcal:
I get two failures from `sanity_check_spec.rb` but they where failing even before I made any changes.

JonRowe:
If you check the build at least one spec is failing due to not having the correct things loaded

maxcal:
Which build?

JonRowe:
Travis, see "Details" on the failure on this PR

maxcal:
This one seems to be failing due to a missing constant - so should have `require 'rails_helper'`. Funny that I did not catch that one locally.

https://travis-ci.org/rspec/rspec-rails/jobs/165104825
```
[Running spec/rspec/rails/example/feature_example_group_spec.rb
An error occurred while loading ./spec/rspec/rails/example/feature_example_group_spec.rb.
Failure/Error:
    describe FeatureExampleGroup do
      it_behaves_like "an rspec-rails example group mixin", :feature,
        './spec/features/', '.\\spec\\features\\'
  
      it "includes Rails route helpers" do
        with_isolated_stderr do
          Rails.application.routes.draw do
            get "/foo", :as => :foo, :to => "foo#bar"
          end
        end
NameError:
  uninitialized constant RSpec::Rails::FeatureExampleGroup](url)
```

A few of the other builds are failing due to dependency incompatibilities - which is not related to this PR since it does not touch the gemspec.

https://travis-ci.org/rspec/rspec-rails/jobs/165104820
```
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    rails was resolved to 3.2.22.5, which depends on
      actionpack (= 3.2.22.5)
    ammeter (~> 1.1.2) was resolved to 1.1.2, which depends on
      railties (>= 3.0) was resolved to 5.0.0.1, which depends on
        actionpack (= 5.0.0.1)
    rspec-rails was resolved to 3.6.0.pre, which depends on
      actionpack (>= 3.0)
```

maxcal:
Is the current master passing on Travis?

JonRowe:
Bar the dependency issues which are being resolved in #1713 (hopefully)

maxcal:
I'm going to try to run the builds locally and track down any remaining issues and will author a new PR

JonRowe:
Hm, would prefer you kept it to this PR... we regularly triage builds until they pass in PRs, better to keep the discussion in one place.

maxcal:
Ok

JonRowe:
If you fix the non dependancy related failures I'll probably merge this

JonRowe:
You might need to run each spec file individually to find which ones are broken, we do this during the Travis build to ensure there aren't load order dependencies which we've missed, which I think is the case here.

maxcal:
Ok will do when I have time.

JonRowe:
No rush :) If you rebase off master you should find the dependency build failures go away too

mikegee:
This PR still looks pretty good to a casual observer like me. @maxcal do you think you could rebase it on the current master and see how that goes?

JonRowe:
Closing as stale for now, if theres any interest in this let me know and I'll reopen

PR #1722, opened 2016-10-11
Improve failure message for AJ matcher

morgoth:
Printing unmatched jobs should help in finding failure reason.

If you have some suggestions how to improve this message, let me know.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -67,7 +67,14 @@ module RSpec
           end
 
           def failure_message
-            "expected to enqueue #{base_message}"
+            "expected to enqueue #{base_message}".tap do |msg|
+              if @unmatching_jobs.any?
+                msg << "\nQueued jobs:"
+                @unmatching_jobs.each do |job|
+                  msg << "\n  #{base_job_message(job)}"
+                end
+              end
+            end
           end
 
           def failure_message_when_negated
@@ -89,7 +96,7 @@ module RSpec
         private
 
           def check(jobs)
-            @matching_jobs_count = jobs.count do |job|
+            @matching_jobs, @unmatching_jobs = jobs.partition do |job|
               if serialized_attributes.all? { |key, value| value == job[key] }
                 args = ::ActiveJob::Arguments.deserialize(job[:args])
                 @block.call(*args)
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -89,7 +96,7 @@ def supports_block_expectations?
         private
 
           def check(jobs)
-            @matching_jobs_count = jobs.count do |job|
+            @matching_jobs, @unmatching_jobs = jobs.partition do |job|
```
JonRowe: Would this return jobs from another class?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -67,7 +67,14 @@ def thrice
           end
 
           def failure_message
-            "expected to enqueue #{base_message}"
+            "expected to enqueue #{base_message}".tap do |msg|
+              if @unmatching_jobs.any?
+                msg << "\nUnmatched enqueued jobs:"
```
JonRowe: I think this would read better as `"Queued jobs:"`

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -67,7 +67,14 @@ def thrice
...
+                @unmatching_jobs.each do |job|
+                  msg << "\n* #{base_job_message(job)}"
```
JonRowe: We don't use a '*' elsewhere when indenting, maybe for consistency just go with 2 spaces?

lib/rspec/rails/matchers/active_job.rb
morgoth: Yes, it will return all jobs that were queued during given action.
I think most of the time number of jobs should be low, so we don't need to scope only to the same class jobs.
This way it's clear for end user what's going on, and it may happen that mistake is also in job class name (like copy pasting) ;-)

lib/rspec/rails/matchers/active_job.rb
morgoth: done

lib/rspec/rails/matchers/active_job.rb
morgoth: and done

JonRowe:
Thanks :)

PR #1723, opened 2016-10-15
add example for single spec

Oneill38:
Add example for running a single spec, in the README.md. I think it fits in well there, and is something I didn't realize you could do, until doing a fair bit of digging.


samphippen:
LGTM. Will merge on green.

PR #1726, opened 2016-10-18
Clarify spec_helper vs rails_helper

sdhull:
When greenfielding a new app & coming from older versions of rspec, the out-of-box experience was confusing for me. Since `.rspec` required `spec_helper` I figured I'd be able to simply write specs & run them.  So I was confounded when my first attempt yielded undefined constant `Rails`.

Searching for answers showed I wasn't the only one. Might as well add a short explanation here, as it's not exactly self-explanatory and even the comments in the files don't really spell it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,13 @@ details.
 **NOTE:** Generators run in RSpec 3.x will now require `rails_helper` instead
 of `spec_helper`.
 
+### `rails_helper` vs `spec_helper`
+As described [in this stackoverflow answer](http://stackoverflow.com/a/24145760/495637), `spec_helper.rb` is for specs which don't depend on Rails (such as specs for classes in the lib directory). `rails_helper.rb` is for specs which do depend on Rails (in a Rails project, most or all of them).
+
+If you want your non-Rails-dependent specs to enforce that they're non-Rails-dependent, and to run as fast as possible when you run them by themselves, you could require `spec_helper.rb` rather than `rails_helper.rb` in those.
+
+The generated `.rspec` file requires `spec_helper`.
+
 ### Generators
 
 Once installed, RSpec will generate spec files instead of Test::Unit test files
--- a/README.md
+++ b/README.md
@@ -98,12 +98,9 @@ details.
 **NOTE:** Generators run in RSpec 3.x will now require `rails_helper` instead
 of `spec_helper`.
 
-### `rails_helper` vs `spec_helper`
-As described [in this stackoverflow answer](http://stackoverflow.com/a/24145760/495637), `spec_helper.rb` is for specs which don't depend on Rails (such as specs for classes in the lib directory). `rails_helper.rb` is for specs which do depend on Rails (in a Rails project, most or all of them).
-
-If you want your non-Rails-dependent specs to enforce that they're non-Rails-dependent, and to run as fast as possible when you run them by themselves, you could require `spec_helper.rb` rather than `rails_helper.rb` in those.
-
-The generated `.rspec` file requires `spec_helper`.
+#### `rails_helper` vs `spec_helper`
+`spec_helper.rb` is for configuration and extensions you use in every spec, however it does not load Rails. The generated `.rspec` file requires `spec_helper` by default.
+`rails_helper.rb` is for specs which do depend on Rails (in a Rails project, most or all of them). You should explicitly require `rails_helper` at the top of specs that require Rails, or you may adjust your `.rspec` file to require `rails_helper` by default.
 
 ### Generators
 


README.md
```diff
@@ -98,6 +98,13 @@ details.
 **NOTE:** Generators run in RSpec 3.x will now require `rails_helper` instead
 of `spec_helper`.
 
+### `rails_helper` vs `spec_helper`
+As described [in this stackoverflow answer](http://stackoverflow.com/a/24145760/495637), `spec_helper.rb` is for specs which don't depend on Rails (such as specs for classes in the lib directory). `rails_helper.rb` is for specs which do depend on Rails (in a Rails project, most or all of them).
+
+If you want your non-Rails-dependent specs to enforce that they're non-Rails-dependent, and to run as fast as possible when you run them by themselves, you could require `spec_helper.rb` rather than `rails_helper.rb` in those.
```
JonRowe: We don't recommend this, it's much simpler to require `spec_helper.rb` in the `.rspec` file, and requiring `spec_helper.rb` doesn't enforce that specs are independant from Rails.

README.md
```diff
@@ -98,6 +98,13 @@ details.
 **NOTE:** Generators run in RSpec 3.x will now require `rails_helper` instead
 of `spec_helper`.
 
+### `rails_helper` vs `spec_helper`
+As described [in this stackoverflow answer](http://stackoverflow.com/a/24145760/495637), `spec_helper.rb` is for specs which don't depend on Rails (such as specs for classes in the lib directory). `rails_helper.rb` is for specs which do depend on Rails (in a Rails project, most or all of them).
```
JonRowe: Please don't reference stack overflow links, if our documentation is in sufficient we should improve it internally.

JonRowe: This is slightly misleading, `spec_helper.rb` is for configuring RSpec and loading the "bare minimum" of support for your suite whilst `rails_helper.rb` is for loading Rails and making it available to your specs.

README.md
```diff
@@ -98,6 +98,13 @@ details.
...
+
+If you want your non-Rails-dependent specs to enforce that they're non-Rails-dependent, and to run as fast as possible when you run them by themselves, you could require `spec_helper.rb` rather than `rails_helper.rb` in those.
+
+The generated `.rspec` file requires `spec_helper`.
```
JonRowe: This would be better worded as "We recommend for consistency that your `.rspec` file requires `spec_helper`,  this is the default when generated.

README.md
```diff
@@ -98,6 +98,13 @@ details.
 **NOTE:** Generators run in RSpec 3.x will now require `rails_helper` instead
 of `spec_helper`.
 
+### `rails_helper` vs `spec_helper`
```
JonRowe: Is this remains in this part of the documentation, it should really be included with the above as it's a kind of upgrade note.

README.md
JonRowe: spec_helper is for all specs

README.md
sdhull: In my view, this sentence answers the inevitable question "so why have a separate spec_helper file at all?" The answer is this--if you have files in `lib` that are self-contained and not dependent on Rails, you could only skip requiring `rails_helper` to ensure they are independent of Rails and also such specs would run faster (by not requiring Rails et all).

This sentence doesn't mention the `.rspec` file at all.

sdhull: That is to say, it's redundant to both explicitly `require "spec_helper"` and have `.rspec` require spec_helper?

It's also redundant to have `.rspec` require spec_helper while explicitly requiring `rails_helper` since rails_helper also explicitly requires `spec_helper`.

The specifics of _how_ you require the appropriate files where are fine to leave up to the developer (ie explicitly vs in `.rspec` file)--what I aim to explain here is _why_ you might only require `spec_helper` in some cases. And if there are no cases where a Rails developer might only require `spec_helper`, then I propose that `spec_helper` be merged with `rails_helper` since merging the files would obviate the need to explain why there are two files in the first place.

JonRowe: > It's also redundant to have .rspec require spec_helper while explicitly requiring rails_helper since rails_helper also explicitly requires spec_helper.

Yes it is, but thats a trade off to allow the file to be "copy pasted" and work.

> The specifics of how you require the appropriate files where are fine to leave up to the developer (ie explicitly vs in .rspec file)--what I aim to explain here is why you might only require spec_helper in some cases. 

You should **always** require spec_helper is my point.

So instead consider explaining why you would require rails_helper as well.

Maybe:

> The seperate `rails_helper` exists to boot rails for specs that require it, any specs that don't need Rails (normal ruby specs) should skip requiring this file.

sdhull: > exists to boot rails for specs that require it

This still strikes me as... unnecessarily difficult. This is `rspec-rails`, right? As in _rspec_... for _Rails_ projects. I mean isn't it true that in 99% of Rails projects, 99% of specs will require Rails? Why not make `.rspec` require `rails_helper` by default?

It seems strange to me that _by default_, the only specs that will run are specs that don't require Rails. If we go to the trouble of generating an `.rspec` file that requires a helper, why not make it require `rails_helper`, so that typical Rails specs will run?

Then if someone wants to do something unusual, they can adjust `.rspec` to only include `spec_helper`

sdhull:
OK I updated the wording to better reflect how you framed these two files.

> exists to boot rails for specs that require it

This still strikes me as... unnecessarily difficult. This is `rspec-rails`, right? As in _rspec_... for _Rails_ projects. I mean isn't it true that in 99% of Rails projects, 99% of specs will require Rails? Why not make `.rspec` require `rails_helper` by default?

It seems strange to me that _by default_, the only specs that will run are specs that don't require Rails. If we go to the trouble of generating an `.rspec` file that requires a helper, why not make it require `rails_helper`, so that typical Rails specs will run?

Then if someone wants to do something unusual, they can adjust `.rspec` to only include `spec_helper`

myronmarston:
> I mean isn't it true that in 99% of Rails projects, 99% of specs will require Rails?

For some users, perhaps, but that's not the way that we would recommend you use RSpec or Rails, and we believe many of your users feel the same.  If 99% of your code is so tightly coupled to rails that rails has to be loaded for it to work, you're going to have all sorts of future difficulties: e.g. upgrades in the future will be more difficult, the code will be harder to reason about (you can't consider the code just on its own but also have to consider the full rails context it depends upon), etc.

If you want to _always_ have rails loaded, then you might as well merge `spec_helper` and `rails_helper` and put a require for the one file in `.rspec`.  But that's not the way we as the RSpec team recommend users do things.  It's nearly impossible to have a good, tight TDD feedback loop with such a setup.

sdhull:
@myronmarston this all makes perfect sense for other rspec gems, but not `rspec-rails`. What kinds of specs do you envision writing? Model specs? You need activerecord. Controller specs? You need actioncontroller. Concern specs? You need activesupport. Feature specs? You need the entire Rails stack.

I'd love to see a few mature Rails apps with specs that largely don't depend on Rails. I simply haven't seen such a thing, so it's difficult to even imagine.

That said, you can simply consider this a dissenting opinion. I still think the README ought to address this `spec_helper` vs `rails_helper` question if you are intent on keeping default setup the same.

myronmarston:
If I wrote rails app these days (which I don't) I would probably implement most of the business logic apart from persistence in standard ruby classes using no part of rails.

For model specs, I'd probably use a special spec helper that loaded just active record and not all of rails, like the one @coreyhaines uses:

https://gist.github.com/coreyhaines/2068977

I'm not sure I'd write controller specs, opting for full-stack request specs instead (which would load and use all of rails).

Of course, take this with a grain of salt since I'm not a rails dev and haven't been since ~2010 or so.

That said, my understanding is this this kind of setup is the consensus recommended setup for rspec-rails by rspec dev team as a whole, and not just me, and many of those folks do rails.  Maybe they can chip in here and explain how their setup works.

> I still think the README ought to address this spec_helper vs rails_helper question if you are intent on keeping default setup the same.

Agreed that the README should explain the intended difference better.

sdhull:
IMO, with the improvements to the speed of `require` with modern versions of ruby combined with `spring` preloader, doing backflips to require fewer files is usually not worth the trouble. Requiring only AR is cool at first, until you do basically anything of substance and you realize activesupport is required for everything.

Nowadays most of my specs are feature specs. But when something is sufficiently complex I'll write specs at lower levels as well. But at the end of the day, I want _all_ my specs to be executed in a way as close to my production setup as possible. Selectively loading files is a great way to have some "gotcha!" moments because things can behave differently when everything is loaded (see: activesupport and monkey patching); as I'm sure you're aware.

Anyway--are you guys happy with the updated wording I came up with or you want more edits?

JonRowe:
> model specs

I use a custom configure and load active record, way quicker than booting Rails.

> controller specs

Now deprecated by the Rails team, it actually is possible to not boot rails to write unit tests but not recommended.

> helper specs

Should really be plain old ruby specs.

> request / feature specs

Now this is where you *require* Rails.

> IMO, with the improvements to the speed of require with modern versions of ruby combined with spring preloader, doing backflips to require fewer files is usually not worth the trouble. 

Spring is a band aid on the problem, Rails boots slowly, nothing to do with require, it's the time Rails takes to boot. We don't recommend spring because it causes more problems than it's worth. 

> Requiring only AR is cool at first, until you do basically anything of substance and you realize activesupport is required for everything.

Fortunately requiring active support is easy, and doesn't requiring booting rails, which is what we're trying to avoid. Also more and more ruby libraries are being written without active support, as it allows them to be better citizens in the ruby world.

JonRowe:
Closed in favour of #1731, thanks for your work here @sdhull :)

Issue #1727, opened 2016-10-21
Mailer spec generator adds an extra "Mailer" if it's already present


mltsy:
Rails 5 apparently now allows you to use `rails g mailer user` to generate UserMailer, hence this commit: https://github.com/rspec/rspec-rails/commit/7a13bf15a191add538c5293c18c7a8dd3fe96e12

However, if the name given includes `Mailer` Rails doesn't append it again, but RSpec does.

`rails g mailer CommentMailer`

generates the correct files/classes, but comment_mailer_spec.rb contains:

`describe CommentMailerMailer`

(which obviously raises an exception when the specs run)

mltsy:
Thanks!  Nice work!

PR #1729, opened 2016-10-24
removed rails 5.0 check for mailer spec generator

hanumakanthvvn:
Removed unnecessary code in `lib/generators/rspec/mailer/templates/mailer_spec.rb`

Now if we give `rails g mailer CommentMailer`
it generates the correct files/classes with file `comment_mailer_spec.rb` containing as
describe `CommentMailer` instead of `CommentMailerMailer`

Fixed: #1727

### DIFF
--- a/lib/generators/rspec/mailer/templates/mailer_spec.rb
+++ b/lib/generators/rspec/mailer/templates/mailer_spec.rb
@@ -1,10 +1,10 @@
 require "rails_helper"
 
 <% module_namespacing do -%>
-RSpec.describe <%= class_name %><%= Rails.version.to_f >= 5.0 ? "Mailer" : "" %>, <%= type_metatag(:mailer) %> do
+RSpec.describe <%= Rails.version.to_f >= 5.0 ? class_name.sub(/(Mailer)?$/, 'Mailer') : class_name %>, <%= type_metatag(:mailer) %> do
 <% for action in actions -%>
   describe "<%= action %>" do
-    let(:mail) { <%= class_name %><%= Rails.version.to_f >= 5.0 ? "Mailer" : "" %>.<%= action %> }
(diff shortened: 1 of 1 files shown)


JonRowe:
Hi, this needs to be here as the recommended way to add a mailer in Rails 5 is `rails g mailer comment`, which will add the Mailer part automatically. I'd support a change to see if it has a `Mailer` end part or not however.

hanumakanthvvn:
@JonRowe  Oh ok got it will change accordingly, Thank you.

hanumakanthvvn:
@JonRowe Fixed. Please check it.

lib/generators/rspec/mailer/templates/mailer_spec.rb
```diff
@@ -1,10 +1,10 @@
 require "rails_helper"
 
 <% module_namespacing do -%>
-RSpec.describe <%= class_name %><%= Rails.version.to_f >= 5.0 ? "Mailer" : "" %>, <%= type_metatag(:mailer) %> do
+RSpec.describe <%= class_name.gsub(/Mailer$/i, '') %><%= Rails.version.to_f >= 5.0 ? "Mailer" : "" %>, <%= type_metatag(:mailer) %> do
```
JonRowe: I'd prefer it if we did this like this:

```
<%= Rails.version.to_f >= 5.0 ? class_name.gsub(/(Mailer)?$/i, 'Mailer') : class_name %>
```

As this will only run the modification on the edge case we know exists.

JonRowe:
Thanks, I've suggested a slight improvement, see what you think :)

hanumakanthvvn:
@JonRowe Modified. Please review. Thanks

mltsy:
:+1: :smile:

PR #1730, opened 2016-10-24
Add support for ActiveJob argument matchers

antstorm:
Hi,

This PR adds support for argument list matcher when matching ActiveJob arguments. Not only it is convenient, but also provides consistency with RSpec's method arguments matcher. Here's an example:

```ruby
expect {
  hello_job.perform_later(42, "David")
}.to have_enqueued_job.with(instance_of(Fixnum), instance_of(String))
```


samphippen:
LGTM

antstorm:
@samphippen cheers mate!

PR #1731, opened 2016-10-24
Add note about different between helpers

JonRowe:
Alternative wording to #1726 ping @myronmarston @sdhull (@samphippen)


sdhull:
This works too `¯\_(ツ)_/¯`

PR #1760, opened 2016-12-16
Get rspec-rails running on Rails 5-0-stable

maclover7:
- Added missing require
- Change Travis CI configuration to use `5-0-stable` instead of `5.0.0`, since new `5.0.x` versions (for example, 5.0.1) will be landing soon

cc @samphippen


maclover7:
Looks like the `3.2.x` builds were already failing on master, so no new failures introduced here.

JonRowe:
Yep, LGTM, thanks

PR #1765, opened 2016-12-27
[wip] Ruby 2.4.0 support

JonRowe:


myronmarston:
You need to prevent rubocop from installing on 2.4 to prevent the incompatible json from installing, or upgrade rubocop. (I opted for the former in the other repos since I didn't want to mess with upgrading rubocop as part of this).

Also, is the 1.8.7 segfault a new thing?

JonRowe:
@myronmarston this is passing on 2.4.0 and should get other builds back to green because of it, we should merge if the build failures match master which is also still failing (compare with https://travis-ci.org/rspec/rspec-rails/builds/184730807), let me know if you approve or otherwise :)

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -211,8 +211,8 @@ def perform; end
 
     it "passes with provided argument matchers" do
       expect {
-        hello_job.perform_later(42, "David")
-      }.to have_enqueued_job.with(instance_of(Fixnum), instance_of(String))
+        ignoring_warnings { hello_job.perform_later(42, "David") }
```
myronmarston: It'd be nice to document what warnings are being ignored here with a comment.

myronmarston:
If this fixes the builds for the other repos, I'm in favor of merging it.  I notice that on 1.8.7 we get some segfaults against certain versions of rails.  should we remove those from the build matrix or something?

spec/rspec/rails/matchers/active_job_spec.rb
JonRowe: We actually don't need that any more, git derp.

JonRowe:
@myronmarston I'd have to defer to @samphippen on that, Rails supports 1.8.7 on those versions so we really should check it, the seg fault seems intermittent perhaps we should retry them? We also need to fix 1.9.2 on master but both those things are for another PR, this should fix our builds.

samphippen:
@JonRowe it looks like it's cucumber: https://travis-ci.org/rspec/rspec-rails/jobs/189500018#L4231 can you try bumping the cucumber version?

myronmarston:
Thanks, @JonRowe.  Want to open a PR in each repo triggering a travis build so we can see if they go green now?  Feel free to use the rspec-dev task to do that.

JonRowe:
I choose to restart the cron'd builds instead

myronmarston:
> I choose to restart the cron'd builds instead

What cron are you talking about?  I've been getting periodic build emails from travis for old commits of mine and had no idea why.  Travis support mentioned a cron as the cause but I didn't know we setup a cron...

https://twitter.com/travisci/status/812100226051932161

JonRowe:
Yeah it's a travis feature that runs master every so often unless a build has been run in that period, I set it up as an experiment as we have in the past come across broken master from downstream dependencies. Bit noisy when the build is broken though.

PR #1768, opened 2017-01-06
Add relish to gemfile to publish to relish

JonRowe:
The rake task for publishing to relish doesn't work for me without this (as the install is isolated I guess?) /cc @myronmarston


samphippen:
LGTM

PR #1769, opened 2017-01-07
[wip] Attempt to fix build

JonRowe:


JonRowe:
Hmm, @samphippen (given you're way more active with rspec-rails than I) what do you want to do about the cucumber segfaults, we can't update cucumber, we're on the last version supporting 1.8.7 anyway

samphippen:
@JonRowe if this is still relevant I can take a look at it. Is this still a thing?

JonRowe:
We still need to fix the 1.8.7 build so yes? I've added a commit which skips cucumber on 1.8.7 which should help.

JonRowe:
A wild green Rails build appears! This is just skipping the acceptance on 1.8.7 though, @samphippen and @myronmarston how do you feel about this? I think given that 1.8.7 is so EOL and this only affects Rails 3 I'm ok with it, we're still running unit specs and confirming cross compatibility with the other gems and its unlikely any 1.8.7 specific errors will crop up here?

myronmarston:
> we're still running unit specs and confirming cross compatibility with the other gems and its unlikely any 1.8.7 specific errors will crop up here?

The problem is that the unit spec coverage in rspec-rails is pretty poor, from what I remember (although I haven't worked much in this code base), and the cukes do a much better job of coverage.

Also, 1.8.7 specific errors are _very_ easy to do accidentally, e.g. wrong hash syntax or whatever.

Couldn't we just add a `Gemfile.lock` for 1.8.7 that contains all the gems that were working previously?

JonRowe:
No the issue here is the segfaults, the gem versions haven't changed.

myronmarston:
> No the issue here is the segfaults, the gem versions haven't changed.

But what is different in the environment from when 1.8.7 builds passed?

JonRowe:
Nothing as far as I know

script/custom_build_functions.sh
```diff
@@ -1,5 +1,10 @@
 function run_cukes {
-  bin/rake acceptance --trace
+  if is_mri_192_plus; then
+    bin/rake acceptance --trace
+    return $?
+  else
+    return 0
```
myronmarston: Might be worth echoing a warning explaining that this part of the build is being skipped and why.

myronmarston:
OK. Ideally, it would be nice to just skip the part of the acceptance cukes that trigger this instead of all of them, but I'll defer to you and @samphippen on this.

JonRowe:
I don't think its that predictable, I'm happy for this to be a temporary measure but I don't have the time to investigate any further and it'd be nice to get them as green as possible.

myronmarston:
> I don't think its that predictable, I'm happy for this to be a temporary measure but I don't have the time to investigate any further and it'd be nice to get them as green as possible.

:+1:

PR #1770, opened 2017-01-07
Default rails install to the latest version

JonRowe:
This will fix the remaining rspec-rails failures for 2.4.0 on the other repos, but doesn't actually affect the travis build for rspec-rails (as we always specify a version). /cc @samphippen @myronmarston


myronmarston:
LGTM

PR #1771, opened 2017-01-09
Fix build

JonRowe:
/cc @myronmarston this is the hopefully final step in getting 1.9.2 working across the board, we also need to merge #1770 to get the main builds working, although that won't fix rspec-rails which is still segfaulting on 1.8.7


myronmarston:
LGTM

JonRowe:
Merging as 1.9.2 is fixed, leaving only 1.8.7

PR #1772, opened 2017-01-09
Pin rails on older rubies

JonRowe:
Again this doesn't affect our rails build, but our other builds, Rails 5 of course only supports Ruby 2.2+ so use 4.2 elsewhere. /cc @myronmarston @samphippen


myronmarston:
LGTM

PR #1777, opened 2017-01-20
`with_routing` not working with `get :index`

### LINKED ISSUE #1652: Invoking `with_routing` does not work with the `get :index` style of invoking controller actions in controller specs
Original context: https://github.com/rspec/rspec-rails/issues/817

App that demonstrates issue: https://github.com/samphippen/test_issue_817

In order for this issue to be fixed, all specs in the app linked above should pass.

sclinede:
As guys previously mentioned in the thread the problem is that `#get` is called on other context then `#with_routing`.

It is caused by https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/example/controller_example_group.rb#L5.
ActionDispatch::Assertions::RoutingAssertions is adding #with_routing to an assertion_instance, and ActionController::TestCase::Behavior is adding #get to ControllerExampleGroup.
But they are included to different classes and do not affect each other.

So I decided to add an ControllerAssertionDelegator which will include both of them. Actually AssertionDelegator missed some methods, so I featured them in that delegator.

Closes #1652


sclinede:
Hi @samphippen. This is my first try for rspec and also just a working draft rather then full solution.
Just want to be sure that I'm on right track)

sclinede:
On the other hand we can do it like that https://gist.github.com/sclinede/3dc0565e5ef9cb9d99bbcf6719302c5d

pirj:
@sclinede Can you please add a test that would fail without the patch?

sclinede:
@pirj You could find failing spec in an additional commit https://github.com/rspec/rspec-rails/pull/1777/commits/a9adc657db9e0d79ef5b072d36c18c5ed7aa7189

sclinede:
The solution was tested against  https://github.com/samphippen/test_issue_817 provided by @penelopezone

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -90,6 +90,30 @@ def my_helper
 
         example.foos_url
       end
+
+      context "when controller is not a stub" do
+        class SomeController < ActionController::Base
```
pirj: Please [stub a const instead](https://rspec.rubystyle.guide/#declare-constants).

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -90,6 +90,30 @@ def my_helper
 
         example.foos_url
       end
+
+      context "when controller is not a stub" do
```
pirj: This spec seems to fail with the fix as well. Might be some trivial thing is missing.

sclinede: True, it's the Rails version updated :)
 Will fix soon

sclinede:
I don't get it. The test suit is failing around at `teardown` method somewhere deep. If you have any suggestions I'll appreciate.
Meanwhile, trying to figure it out by myself...

lib/rspec/rails/adapters.rb
```diff
@@ -60,6 +60,48 @@ def assertion_instance
...
+
+          def init_setup
+            @setup_methods ||= []
+            @setup_blocks ||= []
+          end
+
+          def setup(*methods, &block)
+            @setup_methods += methods
+            @setup_blocks << block if block
+          end
```
pirj: You may have found it yourself already, but anyway - if you add a `def teardown(*); end` here, the spec starts failing with a different error.

sclinede:
So... it looks like I need more time to invest in that change. Hope I'll find it soon :)
I don't give up

pirj:
@sclinede Have you had a chance to take a closer look at `teardown` problem?

PR #1778, opened 2017-01-21
Remove redundant expectation in README.md

tobyhs:
The `be_success` checks that the HTTP status is between 200 and 299, but
the next line already checks that it is 200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,9 +175,8 @@ require "rails_helper"
 
 RSpec.describe PostsController, :type => :controller do
   describe "GET #index" do
-    it "responds successfully with an HTTP 200 status code" do
+    it "responds with an HTTP 200 status code" do
       get :index
-      expect(response).to be_success
       expect(response).to have_http_status(200)
     end
 


JonRowe:
It's not redundant, because it's showing two ways of doing the same thing, its documentation after all :)

tobyhs:
@JonRowe If it is two ways of doing the same thing, would you mind if I changed it so it is commented out, resulting in something like:
```
# Alternatively, if you just want to check for 2xx:
#expect(response).to be_success
```

JonRowe:
Thats not been the style of our README's in the past

tobyhs:
After looking at the README.md files of https://github.com/rspec/rspec-core, https://github.com/rspec/rspec-expectations, https://github.com/rspec/rspec-mocks, https://github.com/rspec/rspec-activemodel-mocks, https://github.com/rspec/rspec-collection_matchers, https://github.com/rspec/rspec-its, it looks like code snippets come in 2 forms:
1) Complete snippets (that one can copy into a file and it can be run with `rspec`)
2) Partial snippets

It looks like the complete snippets are close to resembling the style of real life examples, while the partial snippets are the ones that demonstrate alternatives.

My original motivation for the change in this pull request was to make the snippet closer to what an actual example would look like (because I don't think it is necessary to have both `expect(response).to be_success` and `expect(response).to have_http_status(200)` in actual test code).

PR #1779, opened 2017-01-23
Fix other builds by pinning rest client

JonRowe:


Gemfile
```diff
@@ -12,6 +12,9 @@ group :documentation do
   gem 'redcarpet',     '2.3.0'
   gem 'github-markup', '1.0.0'
   gem 'relish'
+  if RUBY_VERSION '< 2.0.0'
```
myronmarston: This seems to have broken things worse.  Every entry in the build matrix is failing during the `bundle install` phase with:

```
Undefined local variable or method `RUBY_VERSION' for Gemfile
```

JonRowe: It seems I typo'd this

myronmarston:
These changes look fine, although were you expecting them to fix 1.8.7?

JonRowe:
Yeah I did, but bundler is doing something weird. The conflict its complaining about doesn't exist :/

JonRowe:
@myronmarston this is now failing for the usual reasons on 1.8.7, and so should fix the other builds (huzzah :/) turns out theres a bundler issue with displaying dependency resolution on 1.14, 1.13 enabled me to find out what the issue was.

myronmarston:
LGTM.  Do you want to merge this and then kick the builds of the other repos to see if they are indeed green?

PR #1785, opened 2017-01-31
ActiveJob performed matchers

isaacseymour:
* `have_been_performed` is like `have_been_enqueued`, but for performed
  jobs
* `have_performed_job` is like `have_enqueued_job`, but for performed
  jobs

It's somewhat unusual to use `perform_enqueued_jobs`, but it seems nice if 
rspec-rails supports it!

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -67,7 +67,7 @@ module RSpec
           end
 
           def failure_message
-            "expected to enqueue #{base_message}".tap do |msg|
+            "expected to #{self.class::FAILURE_MESSAGE_EXPECTATION_ACTION} #{base_message}".tap do |msg|
               if @unmatching_jobs.any?
                 msg << "\nQueued jobs:"
                 @unmatching_jobs.each do |job|
@@ -78,7 +78,7 @@ module RSpec
           end
 
           def failure_message_when_negated
-            "expected not to enqueue #{base_message}"
+            "expected not to #{self.class::FAILURE_MESSAGE_EXPECTATION_ACTION} #{base_message}"
           end
 
           def message_expectation_modifier
@@ -119,7 +119,7 @@ module RSpec
               msg << " with #{@args}," if @args.any?
               msg << " on queue #{@queue}," if @queue
               msg << " at #{@at.inspect}," if @at
-              msg << " but enqueued #{@matching_jobs_count}"
+              msg << " but #{self.class::MESSAGE_EXPECTATION_ACTION} #{@matching_jobs_count}"
             end
           end
 
@@ -193,6 +193,9 @@ module RSpec
 
         # @private
         class HaveEnqueuedJob < Base
+          FAILURE_MESSAGE_EXPECTATION_ACTION = 'enqueue'.freeze
+          MESSAGE_EXPECTATION_ACTION = 'enqueued'.freeze
+
           def initialize(job)
             super()
             @job = job
@@ -217,6 +220,9 @@ module RSpec
 
         # @private
         class HaveBeenEnqueued < Base
+          FAILURE_MESSAGE_EXPECTATION_ACTION = 'enqueue'.freeze
+          MESSAGE_EXPECTATION_ACTION = 'enqueued'.freeze
+
           def matches?(job)
             @job = job
             check(queue_adapter.enqueued_jobs)
@@ -226,6 +232,38 @@ module RSpec
             set_expected_number(:at_least, 1)
 
             !matches?(proc)
+          end
+        end
+
+        # @private
+        class HavePerformedJob < Base
+          FAILURE_MESSAGE_EXPECTATION_ACTION = 'perform'.freeze
+          MESSAGE_EXPECTATION_ACTION = 'performed'.freeze
+
+          def initialize(job)
+            super()
+            @job = job
+          end
+
+          def matches?(proc)
+            raise ArgumentError, "have_performed_job only supports block expectations" unless Proc === proc
+
+            original_performed_jobs_count = queue_adapter.performed_jobs.count
+            proc.call
+            in_block_jobs = queue_adapter.performed_jobs.drop(original_performed_jobs_count)
+
+            check(in_block_jobs)
+          end
+        end
+
+        # @private
+        class HaveBeenPerformed < Base
+          FAILURE_MESSAGE_EXPECTATION_ACTION = 'perform'.freeze
+          MESSAGE_EXPECTATION_ACTION = 'performed'.freeze
+
+          def matches?(job)
+            @job = job
+            check(queue_adapter.performed_jobs)
           end
         end
       end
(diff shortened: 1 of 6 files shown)


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -174,7 +176,7 @@ def queue_adapter
         # @private
         class HaveEnqueuedJob < Base
           def initialize(job)
-            super()
+            super("enqueue", "enqueued")
```
BobbyMcWho: I could be argued either way, but I think instead of parameters, adding the 
```ruby
@verb_present_tense = verb_present_tense
@verb_past_tense = verb_past_tense
```
after the super here would be preferred.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -269,6 +305,75 @@ def have_been_enqueued
...
+      #         HelloJob.set(wait_until: Date.tomorrow.noon, queue: "low").perform_later(42)
+      #       }
+      #     }.to have_performed_job.with(42).on_queue("low").at(Date.tomorrow.noon)
+      def have_performed_job(job = nil)
+        check_active_job_adapter
+        ActiveJob::HavePerformedJob.new(job)
+      end
+
+      # @api public
+      # Passes if a job has been preformed. May chain at_least, at_most or exactly to specify a number of times.
```
BobbyMcWho:
```suggestion
      # Passes if a job has been performed. May chain at_least, at_most or exactly to specify a number of times.
```

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -11,7 +11,9 @@ module ActiveJob
         # rubocop: disable Style/ClassLength
         # @private
         class Base < RSpec::Matchers::BuiltIn::BaseMatcher
-          def initialize
+          def initialize(verb_present_tense, verb_past_tense)
+            @verb_present_tense = verb_present_tense
+            @verb_past_tense = verb_past_tense
```
JonRowe: How do you feel about making these constants and defined appropriately as frozen strings in the sub matchers?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -193,6 +193,9 @@ def queue_adapter
 
         # @private
         class HaveEnqueuedJob < Base
+          VERB_PRESENT_TENSE = 'enqueue'
```
pirj: Note to self: `.freeze`

pirj: I'm on the fence between using cryptic `self.class::VERB_PRESENT_TENSE` and defining methods in matchers, and calling them from the base class:
```
def verb_present_tense
  'enqueue'.freeze
end
```

JonRowe: I mean naming wise they could be improved, but the reason is the later form causes new allocations even with the freeze I believe.

lib/rspec/rails/matchers/active_job.rb
pirj: I see three viable options:

1. keep as is (only add `.freeze`)

2. define methods with `freeze`
```ruby
def x
  'qwrqwfqw'.freeze
end
x.equal? x # => true
x.object_id == x.object_id
```
Seems that some Ruby magic is going on, and it's being optimized to reduce allocations.
The downside is that we have two method definitions for no good reason, and that doubles the number of methods in those matchers.

3. define class-level `attr_reader`, and use `self.class.verb_present_tense`

Do you have a preference here?

pirj: Speaking about the names, I'm out of good ideas.
`message_expectation_action`/`failure_message_expectation_action`?

lib/rspec/rails/matchers/active_job.rb
pirj: Sorry, responded in the wrong thread.

I see three viable options:

1. keep as is (only add `.freeze`)

2. define methods with `freeze`
```ruby
def x
  'qwrqwfqw'.freeze
end
x.equal? x # => true
x.object_id == x.object_id # => true
```
Seems that some Ruby magic is going on, and it's being optimized to reduce allocations.
The downside is that we have two method definitions for no good reason, and that doubles the number of methods in those matchers.

3. define class-level `attr_reader`, and use `self.class.verb_present_tense`

Do you have a preference here?

pirj: Speaking about the names, I'm out of good ideas.
`message_expectation_action`/`failure_message_expectation_action`?

benoittgt: @pirj I am in favor of 1. Also more in favor of `message_expectation_action`. 😉

pirj:
Also squashed commits.

lib/rspec/rails/matchers/active_job.rb
pirj: Fixed!

pirj:
Squashed on [@isaacseymour's branch](https://github.com/isaacseymour/rspec-rails/tree/activejob-performed-matchers), but that did not update the PR 8E

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -228,6 +234,38 @@ def does_not_match?(proc)
...
+            super()
+            @job = job
+          end
+
+          def matches?(proc)
+            raise ArgumentError, "have_performed_job only supports block expectations" unless Proc === proc
+
+            original_performed_jobs_count = queue_adapter.performed_jobs.count
+            proc.call
+            in_block_jobs = queue_adapter.performed_jobs.drop(original_performed_jobs_count)
```
JonRowe: Would it not be safer to use `-` here?
```
original_performed_jobs = queue_adapter.performed_jobs
proc.call
in_block_jobs = queue_adapter.performed_jobs - original_performed_jobs
```
I mean are jobs guaranteed to be ordered?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -67,7 +67,7 @@ def thrice
           end
 
           def failure_message
-            "expected to enqueue #{base_message}".tap do |msg|
+            "expected to #{self.class::FAILURE_MESSAGE_EXPECTATION_ACTION} #{base_message}".tap do |msg|
```
JonRowe: Is this scoping definitely necessary?

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -405,4 +405,261 @@ def self.name; "LoggingJob"; end
...
   end
+
+  describe "have_performed_job" do
+    before do
+      ActiveJob::Base.queue_adapter.perform_enqueued_jobs = true
+      ActiveJob::Base.queue_adapter.perform_enqueued_at_jobs = true
+
+      stub_const('HeavyLiftingJob', heavy_lifting_job)
+      stub_const('HelloJob', hello_job)
+      stub_const('LoggingJob', logging_job)
```
JonRowe: Just curious what purpose this serves? We don't seem to use the constants so I'm curious...

lib/rspec/rails/matchers/active_job.rb
pirj: I believe it's a copy-paste from [here](https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/matchers/active_job.rb#L206).
It mostly worked like this (one old spec fails, one new):
```ruby
            original_performed_jobs = queue_adapter.performed_jobs.dup
            proc.call
            in_block_jobs = queue_adapter.performed_jobs - original_performed_jobs
```

However, since `drop` is non-destructive, it's more like a `tail`.

The culprit here is that if there were jobs of some type enqueued, and the block enqueues an identical job, `-` reports there was nothing enqueued, like in:
```ruby
[1, 1] - [1] # => []
```
With debug:
```
original_performed_jobs: {:job=>HeavyLiftingJob, :args=>[], :queue=>"default"}
queue_adapter.performed_jobs: {:job=>HeavyLiftingJob, :args=>[], :queue=>"default"}, {:job=>HeavyLiftingJob, :args=>[], :queue=>"default"}
    counts only jobs performed in block (FAILED - 1)

Failures:

  1) ActiveJob matchers have_performed_job counts only jobs performed in block
     Failure/Error:
       expect {
         heavy_lifting_job.perform_later
       }.to have_performed_job.exactly(1)

       expected to perform exactly 1 jobs, but performed 0
 ```

I really hope the ordering behaviour of `queue_adapter.performed_jobs` is the same as `queue_adapter.enqueued_jobs`, and we've relied on the latter for quite some time.

File: lib/rspec/rails/matchers/active_job.rb (hunk 6 above)
Comment (pirj) [reply lib/rspec/rails/matchers/active_job.rb]: Yes, it ends up with an error otherwise:
```
NameError: uninitialized constant RSpec::Rails::Matchers::ActiveJob::Base::FAILURE_MESSAGE_EXPECTATION
_ACTION
```

File: spec/rspec/rails/matchers/active_job_spec.rb (hunk 7 above)
Comment (pirj) [reply spec/rspec/rails/matchers/active_job_spec.rb]: With `perform_enqueued_jobs` we're including ActiveJob's in the loop, that does its [`constantize` here](https://github.com/rails/rails/blob/af2129b/activejob/lib/active_job/core.rb#L51):
```ruby
job = job_data["job_class"].constantize.new
```
so we have to stub the constants.

JonRowe: Copy, fancy adding a comment like this?
```
# stub_const is used so `job_data["job_class"].constantize` works
```

lib/rspec/rails/matchers/active_job.rb
JonRowe: :👍:

lib/rspec/rails/matchers/active_job.rb
JonRowe: Ok leave as is then.

spec/rspec/rails/matchers/active_job_spec.rb
pirj: Makes sense. Added.

benoittgt:
@isaacseymour could you rebase? We fixed the CI.

pirj:
Rebased 👍

benoittgt:
@pirj I didn't you can rebase and push from a fork you don't own. 🤔

pirj:
It seems it's possible by default with this setting:
image

benoittgt:
Thanks! Good to know.

PR #1788, opened 2017-02-08
Change hardcoded path to the pathname.join

shlima:


JonRowe:
`.join` is not always available so thank you but this isn't necessary.

mikegee:
I started looking to see if `Rails.root` was an instance of `Pathname` in the oldest supported Rails, but then I noticed the specs all passes. When is `.join` not available?

PR #1795, opened 2017-03-22
Load spec_helper before Rails

kjg:
so that things in spec_helper, especially coverage tools,
have a chance to load before the Rails code loads

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -1,9 +1,9 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
+require 'spec_helper'
 ENV['RAILS_ENV'] ||= 'test'
 require File.expand_path('../../config/environment', __FILE__)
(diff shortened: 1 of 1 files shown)


samphippen:
@kjg I'm not entirely sure about this. Can you explain your use case or need a little more?

kjg:
@samphippen [Simplecov](https://github.com/colszowka/simplecov), for example, [says that it must be loaded before any of the application code in order to properly track code coverage.](https://github.com/colszowka/simplecov#getting-started)

If you have your Simplecov init in your spec_helper (which makes sense because you want coverage on all files, not just files that need rails_helper), you have Simplecov init at the top of spec_helper, but rails_helper is loading your Rails files before spec_helper and therefor before Simplecov.

Does that make sense?

JonRowe:
@samphippen I would recommend this change as well to be honest, given `spec_helper` should be rails free I think it's safe to merge this. Thanks @kjg

JonRowe:
TIL hitting merge doesn't also post your comment :)

PR #1799, opened 2017-03-26
Audit tweaks

myronmarston:
I'm doing an audit of rspec-rails as part of putting together content for the new RSpec book.  I ran across a couple things that seemed slightly wrong and/or unnecessary and figured I'd fix them while I'm in there.   Check individual commits for explanations of each change.

I don't have a working rspec-rails dev environment (nor do I care to set one up) and I don't really have the time to do any follow up on this, so this can be closed or taken over by another maintainer as desired.  I'm mostly opening this so these things don't get forgotten.


samphippen:
LGTM

PR #1803, opened 2017-04-18
Consistently use `record.new_record?`.

### LINKED ISSUE #1801: Matcher inconsistencies
A couple things I'm noticing:

- [x] `be_a_new(model_class)` checks `new_record?`, but is documented as checking `!persisted?`.  These should agree.
- [x] `be_new_record` checks `!persisted?`. It should probably use the same method that `be_a_new` uses. (`new_record?` makes sense to me).
- [ ] `be_a_new(model_class).with(some_attributes)` does not support composable matching by accepting matchers in the `some_attributes` hash.

myronmarston:
`be_a_new(model_class)` and `be_new_record` were
inconsistent in that one of them checked `!persisted?`
and one checked `new_record?`. The docs also said that
both checked `!persisted?` even only one did.

This changes both to use `new_record?` and to be documented
as such.

Fixes #1801.

Note that I'm not sure this is the right fix but I figured I'd offer to get the conversation on #1801 rolling.  [My book](https://pragprog.com/book/rspec3/effective-testing-with-rspec-3) has an rspec-rails cheat sheet and documents the current inconsistency which is confusing some readers, so I'd like to get this fixed by publication if possible.

/cc @rspec/rspec

### DIFF
--- a/lib/rspec/rails/matchers/be_a_new.rb
+++ b/lib/rspec/rails/matchers/be_a_new.rb
@@ -61,8 +61,8 @@ module RSpec
       end
 
       # @api public
-      # Passes if actual is an instance of `model_class` and returns `false` for
-      # `persisted?`. Typically used to specify instance variables assigned to
+      # Passes if actual is an instance of `model_class` and returns `true` for
+      # `new_record?`. Typically used to specify instance variables assigned to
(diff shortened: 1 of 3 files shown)


samphippen:
LGTM

Issue #1160, opened 2014-08-27
Improve route spec generation for controller generators


cupakromer:
If you use the Rails generator to create a controller with some actions, it will automatically add basic routes for it. This feature would need to do two things:
1. Automatically add passing specs for the generated routes
2. If no actions / routes were specified, include our default filler message

From #1155 (see [discussion](https://github.com/rspec/rspec-rails/pull/1155#issuecomment-53575521))

b-lambert:
Looks like @glittershark gave up on this? I'd be happy to take it on!

samphippen:
@b-lambert go ahead! As far as I'm aware we've not seen any further action on this pull request :)

siakaramalegos:
I'm not sure I quite understand what the request is - rspec already generates a controller test for any route specified in the generator. What's an example of an improved test?

For example, if you run `rails g controller Delete test`, it generates a controller spec file with the content:
```
require 'rails_helper'

RSpec.describe DeleteController, type: :controller do

  describe "GET #test" do
    it "returns http success" do
      get :test
      expect(response).to have_http_status(:success)
    end
  end

end
```

If the `test` route is removed from the generated command, this is what you get instead:
```
require 'rails_helper'

RSpec.describe DeleteController, type: :controller do

end
```

What do we want instead?

samphippen:
@siakaramalegos I believe the idea is to generate routing specs: https://www.relishapp.com/rspec/rspec-rails/v/2-4/docs/routing-specs

siakaramalegos:
Oh interesting! I'll wait a bit to see if @b-lambert picks it up.

glittershark:
I had actually done this (two years ago D:) but gave up when I couldn't manage to make the tests pass. Maybe I can see if I can dredge it up from a backup on whatever computer I had then

b-lambert:
@glittershark see what you can dredge up!

00dav00:
@cupakromer @b-lambert @glittershark If you guys don't I would like to work on this one :)

b-lambert:
I'm no longer working on it so it's up to @glittershark

00dav00:
@cupakromer [This PR](https://github.com/rspec/rspec-rails/pull/1806) should cover the requested feature.

CC  @b-lambert

klyonrad:
Seems like this issue can be closed with https://github.com/rspec/rspec-rails/pull/2134, right?

PR #1806, opened 2017-04-21
Create controller generator for routing specs

00dav00:
This PR covers #1160

This creates routing specs for controller generator

/cc @rspec/rspec

### DIFF
--- a/lib/generators/rspec/controller/controller_generator.rb
+++ b/lib/generators/rspec/controller/controller_generator.rb
@@ -7,8 +7,9 @@ module Rspec
       argument :actions, :type => :array, :default => [], :banner => "action action"
 
       class_option :template_engine, :desc => "Template engine to generate view files"
-      class_option :controller_specs, :type => :boolean, :default => true
-      class_option :view_specs,       :type => :boolean, :default => true
+      class_option :controller_specs, :type => :boolean, :default => true,  :desc => "Generate controller specs"
+      class_option :view_specs,       :type => :boolean, :default => true,  :desc => "Generate view specs"
+      class_option :routing_specs,    :type => :boolean, :default => false,  :desc => "Generate routing specs"
 
       def generate_controller_spec
         return unless options[:controller_specs]
@@ -28,6 +29,13 @@ module Rspec
           template 'view_spec.rb',
                    File.join("spec", "views", file_path, "#{@action}.html.#{options[:template_engine]}_spec.rb")
         end
+      end
+
+      def generate_routing_spec
+        return unless options[:routing_specs]
+
+        template 'routing_spec.rb',
+                 File.join('spec/routing', class_path, "#{file_name}_routing_spec.rb")
       end
     end
   end
--- a/lib/generators/rspec/controller/templates/routing_spec.rb
+++ b/lib/generators/rspec/controller/templates/routing_spec.rb
@@ -1,0 +1,13 @@
+require 'rails_helper'
+
+<% module_namespacing do -%>
+RSpec.describe <%= class_name %>Controller, <%= type_metatag(:routing) %> do
+  describe 'routing' do
+<% for action in actions -%>
+    it 'routes to #<%= action %>' do
+      expect(:get => "/<%= class_name.underscore %>/<%= action %>").to route_to("<%= class_name.underscore %>#<%= action  %>")
+    end
+<% end -%>
+  end
+end
+<% end -%>
--- a/spec/generators/rspec/controller/controller_generator_spec.rb
+++ b/spec/generators/rspec/controller/controller_generator_spec.rb
@@ -90,4 +90,41 @@ RSpec.describe Rspec::Generators::ControllerGenerator, :type => :generator do
       end
     end
   end
+
+  describe 'routing spec' do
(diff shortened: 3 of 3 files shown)


samphippen:
@00dav00 Thanks for this PR! This looks fine to me and I like most of the PR implementation, I do have a few notes. Before we go ahead with this I need to think about whether or not this is a feature we want to support going forward.

I rarely if ever use a routing spec. Do you personally find a lot of value from them?

seuros:
@samphippen I use routing specs for API endpoints since they should follow the specification and not change.

lib/generators/rspec/controller/controller_generator.rb
```diff
@@ -7,8 +7,9 @@ class ControllerGenerator < Base
       argument :actions, :type => :array, :default => [], :banner => "action action"
 
       class_option :template_engine, :desc => "Template engine to generate view files"
-      class_option :controller_specs, :type => :boolean, :default => true
-      class_option :view_specs,       :type => :boolean, :default => true
+      class_option :controller_specs, :type => :boolean, :default => true,  :desc => "Generate controller specs"
+      class_option :view_specs,       :type => :boolean, :default => true,  :desc => "Generate view specs"
+      class_option :routing_specs,    :type => :boolean, :default => true,  :desc => "Generate routing specs"
```
seuros: maybe we could have default : false ?

00dav00:
@samphippen I agree with @seuros, this specs can be useful when the routes does not follow rest conventions.

JonRowe:
@seuros @00dav00 sure they *can* be useful, but that doesn't mean they should be recommended by default, generated specs have a habit of becoming the default, personally I'd welcome adding a seperate generator, for those that so desire, but not adding them as the default for a controller.

seuros:
@JonRowe what about having the option default to false ?

00dav00:
@JonRowe I think @seuros 's option may be a good fit, I mean not adding them as default. WDYT @samphippen ?

juank-pa:
I noticed routing specs are being created by the scaffold generator. So isn't this contribution helping us have a sense of completeness? I think this could work but in any case, I'm also open to not having it by default. WDYT? @JonRowe @samphippen

samphippen:
Hi @00dav00 I've found some time to come back to this. I think I like this overall, but I have some questions:

* if I don't specify an action, what happens?
* if I use a scaffold generator to do a full controller backed by a model, what happens?

00dav00:
Hi @samphippen, thanks for puting some time into this, highly appreciated.

1) `if I don't specify an action, what happens?`

If `--routing-specs` is passed to the generator but no action is given the spec file will be created but will contain no specs. Should I change this behavior to create route specs for the CRUD operations (like the scaffold routing generator) or the creation of this file should be skipped (like in the method for view specs)?

2) `if I use a scaffold generator to do a full controller backed by a model, what happens?`

The scaffold generator is creating the routing specs by default. In opposition with the current PR, the scaffold generator will only skip the routing specs if the flag `--no-routing-specs` is provided. Also if no actions are sent the CRUD operations will be used to create the routing specs. I think this should be consistent between the two generators, WDYT @samphippen? Which approach should be used?

00dav00:
@samphippen This has been here for a while now, do you think we will be able to move it forward or you consider it is not required?

lib/generators/rspec/controller/controller_generator.rb
```diff
@@ -7,8 +7,9 @@ class ControllerGenerator < Base
       argument :actions, :type => :array, :default => [], :banner => "action action"
 
       class_option :template_engine, :desc => "Template engine to generate view files"
-      class_option :controller_specs, :type => :boolean, :default => true
-      class_option :view_specs,       :type => :boolean, :default => true
+      class_option :controller_specs, :type => :boolean, :default => true,  :desc => "Generate controller specs"
+      class_option :view_specs,       :type => :boolean, :default => true,  :desc => "Generate view specs"
+      class_option :routing_specs,    :type => :boolean, :default => false,  :desc => "Generate routing specs"
```
JonRowe: Double spacing here, unless you were intending to align the `:desc`, in which case the latst needs a space removed.

lib/generators/rspec/controller/controller_generator.rb
```diff
@@ -29,6 +30,13 @@ def generate_view_specs
                    File.join("spec", "views", file_path, "#{@action}.html.#{options[:template_engine]}_spec.rb")
         end
       end
+
+      def generate_routing_spec
+        return unless options[:routing_specs]
+
```
JonRowe: This should probably return is actions is empty like views...

lib/generators/rspec/controller/templates/routing_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+<% module_namespacing do -%>
+RSpec.describe <%= class_name %>Controller, <%= type_metatag(:routing) %> do
```
JonRowe: This describe should be a doc string, we don't need the constant here.

00dav00:
Closing this one in favor of #2134

PR #1811, opened 2017-05-02
Allow use of composable matchers inside be_a_new#with

britnia:
The be_a_new(model_class).with(attributes) did not allow the use of composable matchers inside the with method.

Fixes the third and final item in issue [#1801](https://github.com/rspec/rspec-rails/issues/1801).

This appears to be working as expected per my discussion with @myronmarston in the issue thread but I am, of course, open to feedback.

### DIFF
--- a/lib/rspec/rails/matchers/be_a_new.rb
+++ b/lib/rspec/rails/matchers/be_a_new.rb
@@ -49,7 +49,7 @@ module RSpec
 
         def attributes_match?(actual)
           attributes.stringify_keys.all? do |key, value|
-            actual.attributes[key].eql?(value)
+            actual.attributes[key].eql?(value) || (value.respond_to?(:expected, true) && actual.attributes[key].include?(value.expected))
           end
         end
 
--- a/lib/rspec/rails/matchers/be_a_new.rb
+++ b/lib/rspec/rails/matchers/be_a_new.rb
@@ -32,10 +32,11 @@ module RSpec
               message << "expected #{actual.inspect} to be a new #{expected.inspect}"
             end
             unless attributes_match?(actual)
+              describe_unmatched_attributes = surface_descriptions_in(unmatched_attributes)
               if unmatched_attributes.size > 1
-                message << "attributes #{unmatched_attributes.inspect} were not set on #{actual.inspect}"
+                message << "attributes #{describe_unmatched_attributes.inspect} were not set on #{actual.inspect}"
               else
-                message << "attribute #{unmatched_attributes.inspect} was not set on #{actual.inspect}"
+                message << "attribute #{describe_unmatched_attributes.inspect} was not set on #{actual.inspect}"
               end
             end
           end.join(' and ')
--- a/lib/rspec/rails/matchers/be_a_new.rb
+++ b/lib/rspec/rails/matchers/be_a_new.rb
@@ -49,13 +49,13 @@ module RSpec
 
         def attributes_match?(actual)
           attributes.stringify_keys.all? do |key, value|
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/matchers/be_a_new.rb
```diff
@@ -49,7 +49,7 @@ def attributes
 
         def attributes_match?(actual)
           attributes.stringify_keys.all? do |key, value|
-            actual.attributes[key].eql?(value)
+            actual.attributes[key].eql?(value) || (value.respond_to?(:expected, true) && actual.attributes[key].include?(value.expected))
```
myronmarston: To work with all matchers, you need to delegate to the logic of each matcher.  The way we do that for composable matchers is via the `values_match?(expected, actual)` method:

https://github.com/rspec/rspec-expectations/blob/0ad53b7b0a6e34b1d20f64b1c6588a26648cf6ed/lib/rspec/matchers/composable.rb#L51-L69

That said, the `HaveAttributes` matcher already does this (and more) and given that the attribute matching here should have the same semantics, I think it might be best if we just delegate to that matcher here:

``` ruby
def attributes_match?(actual)
  HaveAttributes.new(attributes).matches?(actual)
end
```

If we go that route, we'll probably want to delegate to `HaveAttributes` for part of the failure message, too.

All that said: I don't know much about rails these days and I'm not sure if the `attributes` method here needs to be used.  @JonRowe or @samphippen -- can one of you advise?

britnia: Thanks! I'll dig in tomorrow and see where I get in that direction. I messed with it just now thinking it might just be that straight-forward...but then a bunch of tests blew up :/

britnia: One thing I've noticed upon further inspection is that while have_attributes does appear to use nested matchers the way you and the docs have said, there doesn't seem to actually be any testing around that in [the have_attributes_spec](https://github.com/rspec/rspec-expectations/blob/0ad53b7b0a6e34b1d20f64b1c6588a26648cf6ed/spec/rspec/matchers/built_in/have_attributes_spec.rb). There are a couple a_string_includings in there but those are just in "fail_with." It might be good to add some test coverage around that. Thoughts?

myronmarston: I believe it is covered by tests here:

https://github.com/rspec/rspec-expectations/blob/0ad53b7b0a6e34b1d20f64b1c6588a26648cf6ed/spec/rspec/matchers/built_in/have_attributes_spec.rb#L83-L99

The test coverage there is pretty light because all the heavy lifting is done by other logic the matcher delegates to.

britnia: Eek, how did I miss that? Thanks, ugh.

myronmarston: It's a big spec file, and easy to miss!

britnia:
I took a second stab using values_match? instead eql?. I might try to further refactor this later. I tried to directly use HaveAttributes and that's something else I might circle back to but I think this implementation does All The Things™

britnia:
Yeah, totally. Thanks for the feedback.

britnia:
@myronmarston I made the failure messages more accurate and caught my branch up to master. I also removed my use of regex matching however there is still quite a bit of it in the file. I am happy to edit the tests to not have them but that seems like a separate task from this issue.

britnia:
I had some super weird merge/rebase issues that I just fixed so the `git log` should look a bit cleaner. Apparently I have now written these commits with....myself? Gotta keep it weird somehow I guess.

myronmarston:
Well done, @britnia!

PR #1814, opened 2017-05-09
Fixed AJ job name in generated error message

morgoth:
It was not caught before because of no class name for the anonymous class.


JonRowe:
Thanks @morgoth!

PR #1815, opened 2017-05-11
Correct return value of `path_for`

hlcfan:
Method `ViewPathBuilder#path_for` is actually returning correct url or exception error message.

It's `nil` in the documentation, guess people forgot this when working on #1402. Correcting now.

### DIFF
--- a/lib/rspec/rails/view_path_builder.rb
+++ b/lib/rspec/rails/view_path_builder.rb
@@ -18,7 +18,7 @@ module RSpec
       # @example
       #     # path cannot be built because the params are missing a required element (:id)
       #     view_path_builder.path_for({ :controller => 'posts', :action => 'delete' })
-      #     # => nil
+      #     # => ActionController::UrlGenerationError: No route matches {:action=>"delete", :controller=>"posts"}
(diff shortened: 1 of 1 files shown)


JonRowe:
Is that the correct message? I wouldn't expect it to look like an error...

hlcfan:
@JonRowe 

```
[3] pry(#<RSpec::Rails::ViewPathBuilder>)> path_params
=> {:controller=>"rooms", :action=>"show"}
[4] pry(#<RSpec::Rails::ViewPathBuilder>)> url_for(path_params.merge(:only_path => true))
ActionController::UrlGenerationError: No route matches {:action=>"show", :controller=>"rooms"}
from /Users/thatsme/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/actionpack-5.0.2/lib/action_dispatch/journey/formatter.rb:51:in `generate'
```

JonRowe:
TIL, take out the `<`,'`>` and it'll be all good then

hlcfan:
Updated.

JonRowe:
Thanks :heart:

PR #1818, opened 2017-05-17
[WIP] Fix NoMethodError: render_views?

yujinakayama:

Issue #1800, opened 2017-03-29
NoMethodError: undefined method `example_group' for nil:NilClass


masterkain:
I'm not sure what's triggering this, yet.

```
        # @private
        class LogSubscriber < ::ActiveSupport::LogSubscriber
          def current_example_group
            RSpec.current_example.example_group
          end

          def render_template(_event)
            return if current_example_group.render_views?
            info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
          end
        end
```

```
[127.0.0.1] [bc765ac7-e5e6-4d8f-894f-707b2ce4b309]   Rendered layouts/authentication.html.haml (17.1ms)
[127.0.0.1] [bc765ac7-e5e6-4d8f-894f-707b2ce4b309] Could not log "render_template.action_view" event. NoMethodError: undefined method `render_views?' for RSpec::ExampleGroups::OAuth2Authorization:Class 
 Could not log "render_template.action_view" event. NoMethodError: undefined method `example_group' for nil:NilClass
[
    "/Users/kain/.rvm/gems/ruby-2.4.1/gems/rspec-rails-3.6.0.beta2/lib/rspec/rails/view_rendering.rb:67:in `current_example_group'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/gems/rspec-rails-3.6.0.beta2/lib/rspec/rails/view_rendering.rb:71:in `render_template'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/bundler/gems/rails-ec9b4d39108d/activesupport/lib/active_support/subscriber.rb:95:in `finish'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/bundler/gems/rails-ec9b4d39108d/activesupport/lib/active_support/log_subscriber.rb:83:in `finish'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/bundler/gems/rails-ec9b4d39108d/activesupport/lib/active_support/notifications/fanout.rb:102:in `finish'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/bundler/gems/rails-ec9b4d39108d/activesupport/lib/active_support/notifications/fanout.rb:46:in `block in finish'",
    "/Users/kain/.rvm/gems/ruby-2.4.1/bundler/gems/rails-ec9b4d39108d/activesupport/lib/active_support/notifications/fanout.rb:46:in `each'",
[...]

PR #1827, opened 2017-05-31
Fix undefined method `example_group' for nil & NoMethodError: render_views? for feature specs

aCandidMind:
This is my stab at issue #1800.

In my case the `ViewRenderer` code that was added with PR #1732 was triggered via the capybara webdriver and `RSpec.current_example` is nil then causing the error in the title. @yujinakayama's spec repro in #1818 points to the cause being that `FeatureExampleGroup` doesn't include `RSpec::Rails::ViewRendering` while `ControllerExampleGroup` does.

This commit makes the tests in PR #1818 pass (via the `respond_to?`) and running the whole rspec-rails suite only caused bundler errors for me (I tweaked the gemspec for local gem development). Using this local commit also got rid of loads of errors (those from the title) when running feature specs in my own project.

I have to say that this is my very first experience with RSpec code and that I just deemed the `included` block to be a good place to do this. Bear with me, if I forgot something.

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -77,8 +77,6 @@ module RSpec
           end
         end
 
-        LogSubscriber.attach_to(:action_view)
-
         # Delegates all methods to the submitted resolver and for all methods
         # that return a collection of `ActionView::Template` instances, return
         # templates with modified source
@@ -150,6 +148,9 @@ module RSpec
       end
 
       included do
+        # Make sure to log when template rendering was prevented by rspec-rails
+        EmptyTemplateResolver::LogSubscriber.attach_to(:action_view)
+
         before do
           unless render_views?
             @_original_path_set = controller.class.view_paths
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -64,12 +64,16 @@ module RSpec
         # @private
         class LogSubscriber < ::ActiveSupport::LogSubscriber
           def current_example_group
-            RSpec.current_example.example_group
+            # When running feature specs current_example can be nil if the subscriber runs in a different context
+            # than the specs themselves. E.g. in a capybara thread.
+            RSpec.current_example.try!(:example_group)
           end
 
           def render_template(_event)
-            return if current_example_group.render_views?
-            info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
+            example_group = current_example_group
+            if example_group.respond_to?(:render_views?) && !example_group.render_views?
+              info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
+            end
           end
         end
 
--- a/spec/rspec/rails/view_rendering_spec.rb
+++ b/spec/rspec/rails/view_rendering_spec.rb
@@ -183,7 +183,7 @@ module RSpec::Rails
         ActiveSupport::LogSubscriber.logger
       end
 
-      pending 'does not cause an error' do
+      it 'does not cause an error' do
         expect(logger).not_to receive(:error).with(a_string_starting_with('Could not log "render_template.action_view" event.'))
         view.render(body: 'foo')
       end


aCandidMind:
Maybe the author of #1732 @andymorris can also have a look at it, whether I preserved all behavior. As I said tests are green and to me it looks good. But I'm not deep into RSpec behavior.

aCandidMind:
ad40aae is only an optimization for the case when no controller specs are run

lib/rspec/rails/view_rendering.rb
```diff
@@ -64,17 +64,19 @@ def self.nullify_template_rendering(templates)
...
-            return if current_example_group.render_views?
-            info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
+            example_group = current_example_group
+            if example_group.respond_to?(:render_views?) && !example_group.render_views?
+              info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
+            end
           end
         end
 
-        LogSubscriber.attach_to(:action_view)
```
JonRowe: Why did this have to move?

lib/rspec/rails/view_rendering.rb
```diff
@@ -64,17 +64,19 @@ def self.nullify_template_rendering(templates)
...
+            # When running feature specs current_example can be nil if the subscriber runs in a different context
+            # than the specs themselves. E.g. in a capybara thread.
+            RSpec.current_example.try!(:example_group)
           end
 
           def render_template(_event)
-            return if current_example_group.render_views?
-            info("  Template rendering was prevented by rspec-rails. Use `render_views` to verify rendered view contents if necessary.")
+            example_group = current_example_group
+            if example_group.respond_to?(:render_views?) && !example_group.render_views?
```
JonRowe: We prefer guard condition style returns, prehaps just change the first line to:
`return if !example_group.respond_to?(:render_views?) || !example_group.render_views?`

lib/rspec/rails/view_rendering.rb
```diff
@@ -64,17 +64,19 @@ def self.nullify_template_rendering(templates)
         # @private
         class LogSubscriber < ::ActiveSupport::LogSubscriber
           def current_example_group
-            RSpec.current_example.example_group
+            # When running feature specs current_example can be nil if the subscriber runs in a different context
+            # than the specs themselves. E.g. in a capybara thread.
+            RSpec.current_example.try!(:example_group)
```
JonRowe: Please don't use `try!` here, we try to avoid the usage of non standard Ruby, you could add a guard to return nil if theres no current example but really we should figure out why its nil.

lib/rspec/rails/view_rendering.rb
aCandidMind: It's only an optimization for the case when no controller specs are run. E.g. when only running feature specs it should make things a little faster since not every template rendering is going through that Subscriber and so on. I could easily refrain from moving it.

lib/rspec/rails/view_rendering.rb
aCandidMind: I'll address both points. RSpec.current_example is nil in the case where the request came from the (Selenium) webdriver out of the browser (e.g. via AJAX) and not from an RSpec example group.

lib/rspec/rails/view_rendering.rb
aCandidMind: `RSpec.current_example` is `nil` in the case where the request came from the (Selenium) webdriver out of the browser (e.g. via AJAX) and not from an RSpec example group. I'll add that little extra info to the code comment.

JonRowe: No don't thats not the cause, this is only executed during a test process when this should have a thread value, its not selenium causing it.

b-wen:
now I also confronted with this error message.
and I have no idea why this error happen and how to avoid it under latest rspec-rails of version 3.6.0.

xaviershay:
Fixed in https://github.com/rspec/rspec-rails/pull/1831

xaviershay:
(sorry I didn't see this PR until after I merged the other one, but I do think it addresses the feedback in here)

Issue #1870, opened 2017-09-13
Negative `have_enqueued_job` does not work as expected with multiple jobs of same kind


povilasjurcys:
Here is my proof of concept:

```
require 'spec_helper'

class MyJob < ActiveJob::Base
  def perform
    OtherJob.perform_later
    OtherJob.perform_later
  end
end

class OtherJob < ActiveJob::Base
end

describe MyJob do
  subject(:my_job) { MyJob.new }

  it 'enqueues job' do
    expect { my_job.perform }.to have_enqueued_job(OtherJob).at_least(:once)
    expect { my_job.perform }.not_to have_enqueued_job(OtherJob)
  end
end
```

As I understand, `have_enqueued_job` checks if job was performed exactly once, so `expect { my_job.perform }.to have_enqueued_job(OtherJob)` is same as `expect { my_job.perform }.to have_enqueued_job(OtherJob).exactly(:once)`. 

But in negative case `expect { my_job.perform }.not_to have_enqueued_job(OtherJob)` becomes same as `expect { my_job.perform }.not_to have_enqueued_job(OtherJob).exactly(:once)`. So it's quite clear why `expect { my_job.perform }.not_to have_enqueued_job(OtherJob)` returns true when there is multiple jobs of same kind, but behavior is confusing.

Tried with rspec-rails 3.6.1.

JonRowe:
I'm not sure what you're asking here, it seems your code snippet *should* fail, as `expect { my_job.perform }.not_to have_enqueued_job(OtherJob)` means you haven't enqueued any jobs, and you have. Expectations don't have any knowledge of each other.

Are you saying it passes? Because thats a bug.

povilasjurcys:
I'm saying that `expect { my_job.perform }.to have_enqueued_job(OtherJob).at_least(:once)` passes and `expect { my_job.perform }.not_to have_enqueued_job(OtherJob)` passes

whatdacode:
I faced some issue. You can hack it with:
```ruby
expect{ my_job.perform }.not_to have_enqueued_job(OtherJob).exactly(0).times
```

marciojg:
Try this. Worked for me

`expect{ my_job.perform }.to have_enqueued_job(OtherJob).at_least(0).times`

beeplove:
> Try this. Worked for me
> 
> ```ruby
> expect{ my_job.perform }.to have_enqueued_job(OtherJob).at_least(0).times
> ```

at_least(0).times ?
[...]

PR #1871, opened 2017-09-14
Make sure that negative matcher fails on multiple enqueues of the same kind

povilasjurcys:
Fix for #1870. 

Without this fix test passes for given example
```
expect {
  job.perform_later
  job.perform_later
}.not_to have_enqueued_job
```

This happens because by default we check using options `:exactly, 1`. That is, under the hood 
`}.not_to have_enqueued_job`
is same as
`.not_to have_enqueued_job.exactly(:once)`. 

In negative examples this is incorrect - we should use `:exaclty, 0` instead. 
Negative example 
`.not_to have_enqueued_job`
 should be equal to positive example with conditions
`.to have_enqueued_job.exactly(0)`

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -16,7 +16,8 @@ module RSpec
             @queue = nil
             @at = nil
             @block = Proc.new {}
-            set_expected_number(:exactly, 1)
+            @using_default_expected_number = true
+            silently_set_expected_number(:exactly, 1)
           end
 
           def with(*args, &block)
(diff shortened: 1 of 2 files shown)


samphippen:
can you provide full examples of what the various kinds of error messages look like when this matcher fails?

povilasjurcys:
@samphippen did you had something like this in mind: https://github.com/rspec/rspec-rails/pull/1871/commits/38801acd0c0bd00af29d9d22eae7c88251ccc083

pirj:
This is pretty much covered in #2069 
Please feel free to reopen if you're up to improve current failure messages.

In general, the use of quantitive qualifiers with negated version of those matchers reads weird.

PR #1879, opened 2017-10-10
Fixed documentation for controller specs

andresakata:
Now it's mandatory to explicitly send `params` in the keyword arguments.


JonRowe:
This is a Rails post 5 change yes? The tests/docs you've changed are for pre 5!

andresakata:
You're right. I didn't pay attention to it.

PR #1884, opened 2017-10-17
Fix puma loading in Rails 5.1

samphippen:


JonRowe:
This LGTM @samphippen, merge on green and get out a bugfix :)

guilleiguaran:
I'm trying to fix this on Rails side to make this work even when puma gem isn't present (making Puma the default but allowing alternative servers as WebRick and Unicorn)

We aren't requiring to Puma gem directly in SystemTestCase anymore: https://github.com/rails/rails/commit/50f697664edf0d2deff22f3f1a1c8e01d54a74ca

and Puma is set as default server through `Capybara.server`:  https://github.com/rails/rails/blob/50f697664edf0d2deff22f3f1a1c8e01d54a74ca/actionpack/lib/action_dispatch/system_testing/server.rb#L23

I'll create a test app with rails master and rspec 3.7.0 to check if the error persists and hopefully, we can backport this to 5.1.x

guilleiguaran:
I've created a test app with rails master, rspec 3.7.0, added a system spec and removed puma gem and after of setting in `spec_helper.rb`:

```ruby
config.before(:suite)   { Capybara.server = :webrick }
```

The spec worked as expected.

JonRowe:
@guilleiguaran any idea when that'll be released? And will it be a minor or patch level increment.

ansonhoyt:
@guilleiguaran thanks for this. I was just trying out a system test with `Capybara.server = :passenger`, and noticed the hardcoded `require "rack/handler/puma"`.

A 5.1 backport would be much appreciated, as it would let my tests more closely match production.

PR #1890, opened 2017-10-19
Make requiring order of `spec/support/**/*.rb` to be fixed

meganemura:
Dir.glob is not guaranteed order.

This change may help who faced the problem about test success and failure across some environments.


JonRowe:
Thanks for your suggestion, this is best left to the projects to decide, if they require files in a specific order they should require them in that order, `sort` merely makes them alphabetical which is also not necessarily the correct order.

meganemura:
ok. thanks.

PR #1892, opened 2017-10-20
Move let statement "new_attributes" under "PUT #update with valid par…

monmonmon:
…ams" to "POST #create with valid params"
because "new_attributes" should be used to create "new" records


JonRowe:
In this context new means "changed" attributes.

PR #1893, opened 2017-10-22
Check for puma and capybara when defining system tests

JonRowe:
Rather than skipping the definition altogether define a fake module that
warns about the dependencies.


pat:
@JonRowe Just gave this a spin in a Rails app that doesn't have/need Capybara, and it didn't have any issues, nor did I see any warning (which is good, because I don't have any system tests).

JonRowe:
Thanks so much @pat!

JonRowe:
@myronmarston can you give me a quick sanity check in lieu of @samphippen (who's in hospital right now), as I'd like to release it to help rails devs

PR #1903, opened 2017-11-03
Delay loading system test integration

JonRowe:
Delay loading system tests as late as possible in order to allow people who arent using them to not have to turn off puma or other such things. It also enables other webservers to be configured other
than puma which is needed in Rails as of https://github.com/rails/rails/commit/50f697664edf0d2deff22f3f1a1c8e01d54a74ca

This should help people who have issues with puma loading when they're not using system test integration. /cc @myronmarston @samphippen @sj26

### DIFF
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -1,113 +1,96 @@
-can_load_system_tests = false
-begin
-  require 'puma'
-  require 'capybara'
-  if ActionPack::VERSION::STRING >= "5.1"
-    require 'action_dispatch/system_test_case'
-    can_load_system_tests = true
-  end
-# rubocop:disable Lint/HandleExceptions
-rescue LoadError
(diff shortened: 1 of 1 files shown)


myronmarston:
This seems reasonable, but I don't think I have sufficient context to be the only one to review this.

sj26:
This looks right to me. 👍 

The main thing is making sure `action_dispatch/system_test_case` isn't loaded unless it's used which [this seems to do](https://github.com/rspec/rspec-rails/pull/1903/files#diff-779a18a53423de804d5cbdc1edc6552dR43).

Thanks for making this happen, @JonRowe!

samphippen:
👍

JonRowe:
Going to merge this as the build failure is unrelated, will attempt to resolve that when I can

sj26:
:tada:

JonRowe:
This got released as 3.7.2, hopefully helps!

PR #1911, opened 2017-11-14
Don't load ActionPack and ActionView examples, when they are not available

knapo:
When I use rails without `action_pack` or `action_view`, `rspec-rails` still  expects them to be both loaded and available. This PR fixes the issue, similar way it's handled for `ActiveJob` or `ActionMailer`

Better to ignore whitespaces when looking on code changes here https://github.com/rspec/rspec-rails/pull/1911/files?w=1 :)

### DIFF
--- a/lib/rspec/rails/example.rb
+++ b/lib/rspec/rails/example.rb
@@ -1,13 +1,16 @@
 require 'rspec/rails/example/rails_example_group'
-require 'rspec/rails/example/controller_example_group'
-require 'rspec/rails/example/request_example_group'
-require 'rspec/rails/example/helper_example_group'
-require 'rspec/rails/example/view_example_group'
-require 'rspec/rails/example/mailer_example_group'
-require 'rspec/rails/example/routing_example_group'
 require 'rspec/rails/example/model_example_group'
-require 'rspec/rails/example/job_example_group'
 require 'rspec/rails/example/feature_example_group'
-require 'rspec/rails/example/system_example_group'
 require 'rspec/rails/example/channel_example_group'
 require 'rspec/rails/example/mailbox_example_group'
+require 'rspec/rails/example/mailer_example_group' if defined?(ActionMailer)
+require 'rspec/rails/example/job_example_group' if defined?(ActiveJob)
+require 'rspec/rails/example/view_example_group' if defined?(ActionView)
+
+if defined?(ActionPack)
+  require 'rspec/rails/example/controller_example_group'
+  require 'rspec/rails/example/helper_example_group'
+  require 'rspec/rails/example/request_example_group'
+  require 'rspec/rails/example/routing_example_group'
+  require 'rspec/rails/example/system_example_group'
+end
--- a/lib/rspec/rails/example/job_example_group.rb
+++ b/lib/rspec/rails/example/job_example_group.rb
@@ -1,23 +1,9 @@
 module RSpec
   module Rails
-    # @api public
-    # Container module for job spec functionality. It is only available if
-    # ActiveJob has been loaded before it.
+    # Container module for job spec functionality.
     module JobExampleGroup
-      # This blank module is only necessary for YARD processing. It doesn't
-      # handle the conditional `defined?` check below very well.
-    end
-  end
-end
-
-if defined?(ActiveJob)
-  module RSpec
-    module Rails
-      # Container module for job spec functionality.
-      module JobExampleGroup
-        extend ActiveSupport::Concern
-        include RSpec::Rails::RailsExampleGroup
-      end
+      extend ActiveSupport::Concern
+      include RSpec::Rails::RailsExampleGroup
     end
   end
 end
--- a/lib/rspec/rails/example/mailer_example_group.rb
+++ b/lib/rspec/rails/example/mailer_example_group.rb
@@ -1,36 +1,22 @@
 module RSpec
   module Rails
-    # @api public
-    # Container module for mailer spec functionality. It is only available if
-    # ActionMailer has been loaded before it.
+    # Container module for mailer spec functionality.
     module MailerExampleGroup
-      # This blank module is only necessary for YARD processing. It doesn't
-      # handle the conditional `defined?` check below very well.
-    end
-  end
-end
+      extend ActiveSupport::Concern
+      include RSpec::Rails::RailsExampleGroup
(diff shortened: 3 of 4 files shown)


pirj:
@knapo Are you up to finish this? Some things have changed since the moment you've opened the pull request. Can you please rebase?

knapo:
@pirj omg, it's been a while, better later than never :) rebased!

pirj:
I believe YARD complains about undocumented methods/classes.

knapo:
It's actually pretty weird and it complains about: `# @!attribute [r]`, which probably are [ not valid ](https://github.com/lsegal/yard/issues/787#issuecomment-46745104), but not idea why yard didn't complained about them before. I wonder, if current master would pass right now.

knapo:
Did an experiment https://github.com/rspec/rspec-rails/pull/2257

pirj:
The problem seems to disappear if you change the conditional to a guard statement:
```ruby
- if defined?(ActionPack)
+ return unless defined?(ActionPack)
```

Wondering if it's good like that, or it would be better to put conditionals in `lib/rspec/rails/example.rb`.

Also, with this approach empty modules preceeding the conditional become redundant.

Highly appreciate if along with your changes you also change the rest of files in `lib/rspec/rails/example/` to match the style you settle on for this change.

knapo:
> Wondering if it's good like that, or it would be better to gather all those conditionals in

I think it's a way better approach. I did it the way it was done for `JobExampleGroup` and `MailerExampleGroup` . Gonna give it a try.

knapo:
@pirj I think it's good now!

lib/rspec/rails/example/job_example_group.rb
```diff
@@ -1,23 +1,9 @@
 module RSpec
   module Rails
-    # @api public
-    # Container module for job spec functionality. It is only available if
-    # ActiveJob has been loaded before it.
```
benoittgt: Are we sure we want to remove those comments?

pirj:
So much better code-wise 👍 

Have you tested it against your project?

Just wondering if the following will fail:
```ruby
# lib/rspec/rails/configuration.rb
config.include RSpec::Rails::SystemExampleGroup, type: :system
```
There are conditional checks wether to include e.g. `JobExampleGroup`:
```ruby
        config.include RSpec::Rails::JobExampleGroup, type: :job
      end
```
maybe add something similar?

lib/rspec/rails/example/job_example_group.rb
knapo: @benoittgt The diff is misleading, we removed empty modules, not comments only.

pirj: I think @benoittgt meant that those comments would still be helpful in the description of what's left, i.e.:
```diff
  module RSpec
    module Rails
-     # Container module for job spec functionality.
+     # Container module for job spec functionality. It is only available if
+     # ActiveJob has been loaded before it.
```

lib/rspec/rails/example.rb
```diff
@@ -1,13 +1,16 @@
...
-require 'rspec/rails/example/job_example_group'
 require 'rspec/rails/example/feature_example_group'
-require 'rspec/rails/example/system_example_group'
 require 'rspec/rails/example/channel_example_group'
 require 'rspec/rails/example/mailbox_example_group'
+require 'rspec/rails/example/mailer_example_group' if defined?(ActionMailer)
+require 'rspec/rails/example/job_example_group' if defined?(ActiveJob)
+require 'rspec/rails/example/view_example_group' if defined?(ActionView)
+
+if defined?(ActionPack)
```
JonRowe: If we go ahead with this, this is not my preferred way to handle this, the constants must always exist because extensions may expect them to exist. They are part of our public API and semver means we must maintain this.

spec/rspec/rails/example/job_example_group_spec.rb
```diff
@@ -1,10 +1,8 @@
-require "spec_helper"
+require 'spec_helper'
```
JonRowe: Revert arbitary changes.

spec/rspec/rails/example/job_example_group_spec.rb
```diff
@@ -1,10 +1,8 @@
...
 
 module RSpec::Rails
   describe JobExampleGroup do
-    if defined?(ActiveJob)
-      it_behaves_like "an rspec-rails example group mixin", :job,
-        './spec/jobs/', '.\\spec\\jobs\\'
-    end
+    it_behaves_like 'an rspec-rails example group mixin', :job,
+      './spec/jobs/', '.\\spec\\jobs\\'
```
JonRowe: Revert arbitary changes.

lib/rspec/rails/example/job_example_group.rb
benoittgt: Yes sorry @knapo. @pirj explained it properly.

pirj:
@knapo Have you had a chance to reproduce the problem that this pull request addresses?

pirj:
Even though the condition checking if `actionpack` is present might not be required, this pull request cleans some things up, and it would be nice to wrap it up.

knapo:
@pirj Sorry, no. I've been quite busy, will be back to it in a next few days!

pirj:
Just a friendly reminder @knapo

lib/rspec/rails/example.rb
knapo: Agree, valid poin.

knapo:
@pirj Thanks for reminder :)

It was a long time ago since I experienced the issue and don't remember exact errors and reasons I created the PR. I no longer use `rails` (so `rspec-rails` too) in my current projects, but I revisited the issue from scratch again. I created a test project which has only `railties` loaded and can see that [actionpack is its dependency](https://github.com/rails/rails/blob/40b7d93c5bf04974cfd0201a8e391c499c046f0b/railties/railties.gemspec#L40), so it's always loaded - no matter is used/needed or not (it sucks btw, especially it's the heaviest rails component).
  
Taking account of that, I don't see benefits in adding checks introduced by this PR, since the key assumption of creating the PR **_When I use rails without action_pack or action_view_** is never met.

pirj:
Thanks for double-checking this @knapo !

PR #1913, opened 2017-11-14
Fix Ruby warning: character class has duplicated range

mattbrictson:
When running the specs in master, I see the following warning printed:

```
lib/rspec/rails/example/system_example_group.rb:52: warning: character class has duplicated range: /[\n\s]+/
```

This PR fixes the warning by simplifying the offending regex to remove the redundant `\n`.


JonRowe:
Thanks, I wasn't aware `\s` now matched new lines.

PR #1914, opened 2017-11-14
Fix "uninitialized constant Gem" and other Travis errors

mattbrictson:
The Travis builds have been failing for the past couple weeks. This PR makes them green again ✅ 

It seems that the trigger for the failures was the recent release of Bundler 1.16.0, which brings some subtly different behavior regarding environment variables and binstubs. Here are the changes I made in this PR to get Travis working:

1. Fix a RuboCop violation that snuck in with #1913.
2. Add `bundler` to the list of binstubs that we explicitly generate when creating the test Rails app. The `bin/bundle` binstub that ships as part of `rails new` is in the pre-1.16 format, and conflicts with the 1.16 format binstubs that are generated for rake and rspec (see https://github.com/bundler/bundler/issues/6149). Fix this by replacing the default `bin/bundle` with a fresh one generated by `bundle binstubs bundler --force`.
3. The `Bundle.with_clean_env` wrapper seems to remove some important information from the environment that causes binstubs to fail with errors like `uninitialized constant Gem (NameError)`. Fix this by using `Bundle.with_original_env` instead (which is what the Bundler documentation [recommends](https://github.com/bundler/bundler/blob/d17f2e43236ba53f05b70e7317d218ba94e293c8/lib/bundler.rb#L281)). Because Travis also sets its own `BUNDLER_GEMFILE` variable, I had to also reset that one manually.
4. The `RAILS_VERSION` variable was not being parsed correctly, causing incorrect Cucumber tags to be used. Fixed by using a regex to extract the version number.
5. Part of the Travis script runs `bin/yard doc`. This was failing because gems where missing. Fix by removing `--without documentation` from the bundler command.

Along the way I also added `--no-rc` to the sample app setup command to fix conflicts in my local development environment.


JonRowe:
:heart:

mattbrictson:
Getting close! 🤞

JonRowe:
Zomg amazing :) Thanks! ❤️

Issue #1919, opened 2017-11-25
Model Spec example


twest7:
This example does not appear to work...

```
require "rails_helper"

RSpec.describe User, :type => :model do
  it "orders by last name" do
    lindeman = User.create!(first_name: "Andy", last_name: "Lindeman")
    chelimsky = User.create!(first_name: "David", last_name: "Chelimsky")

    expect(User.ordered_by_last_name).to eq([chelimsky, lindeman])
  end
end
```

error:
```
1) User orders by last name
     Failure/Error: expect(User.ordered_by_last_name).to eq([chelimsky, lindeman])

     NoMethodError:
       undefined method `ordered_by_last_name' for #<Class:0x00000004ff12f8>
     # ./spec/models/user_spec.rb:24:in `block (2 levels) in <top (required)>'
```

Changing `User.ordered_by_last_name` to `User.order(:last_name)` resolves the error and spec passes.  I see this .ordered_by_last_name example posted elsewhere as well back to 2.x versions... is this a reference to removed syntax or is something missing?  Tested against RSpec 3.7 release and github master with same results.

mikegee:
I agree that this is confusing. The example you posted is [copied directly from the README](https://github.com/rspec/rspec-rails#model-specs). 

I don't think Rails ever provided methods like `ordered_by_last_name` as described in that example. I think it is implying that the example is describing a method explicitly defined in the application.

Perhaps that example could be introduced differently? Instead of 
> For example:

something like

> Here is an example specifying the behavior of a `User` model that defines `.ordered_by_last_name`:

joeltaylor:
Opened a PR that updates the example with the one being used on Relish. Hopefully it'll help reduce some of the confusion.

samphippen:
This was fixed by #1920

PR #1920, opened 2017-11-26
Update model example to match Relish (#1919)

joeltaylor:
Updated the model example to match the one that is being used in the Relish documentation since the original example can cause potential confusion.


JonRowe:
Thanks!

PR #1932, opened 2017-12-20
Remove useless ivar assign in controller spec

riseshia:
This assignment is not necessary here.
It maybe need to be someone, but there is no reason to assign in generated code at least ;)

before rspec-rails generates(model: Person, Rails > 5):

```ruby
describe "GET #index" do
  it "returns a success response" do
    person = Person.create! valid_attributes
    get :index, params: {}, session: valid_session
    expect(response).to be_success
  end
end
```

after:

```ruby
describe "GET #index" do
  it "returns a success response" do
    Person.create! valid_attributes
    get :index, params: {}, session: valid_session
    expect(response).to be_success
  end
end
```

How do you think?


JonRowe:
Thanks :heart: although this was a local var not an ivar for future reference :)

riseshia:
😭 Thanks to fix my poor english

PR #1933, opened 2017-12-28
generator for system spec

andrzejsliwa:
Rspec since 3.7 is supporting the system specs, so would be great to add generator for it.

### DIFF
--- a/lib/generators/rspec/system/system_generator.rb
+++ b/lib/generators/rspec/system/system_generator.rb
@@ -1,0 +1,24 @@
+require 'generators/rspec'
+
+module Rspec
+  module Generators
+    # @private
+    class SystemGenerator < Base
+      class_option :system_specs, :type => :boolean, :default => true,  :desc => "Generate system specs"
+
+      def generate_feature_spec
+        return unless options[:system_specs]
+
+        template template_name, File.join('spec/system', class_path, filename)
+      end
+
+      def template_name
+        'system_spec.rb'
+      end
+
+      def filename
+        "#{table_name}_spec.rb"
+      end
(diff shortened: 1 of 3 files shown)


JonRowe:
I think you need to make sure these specs only run on rails versions that support system specs.

joshRpowell:
@andrzejsliwa #2071 looks like a good candidate for including this PR

JonRowe:
No this can be independant, it just needs to be gated on Rails version

lib/generators/rspec/system/system_generator.rb
```diff
@@ -0,0 +1,24 @@
+require 'generators/rspec'
+
+module Rspec
+  module Generators
+    # @private
+    class SystemGenerator < Base
+      class_option :system_specs, :type => :boolean, :default => true,  :desc => "Generate system specs"
+
+      def generate_feature_spec
+        return unless options[:system_specs]
```
joshRpowell:
```ruby
@andrzejsliwa perhaps you'll have something more elegant in mind, but this is a start:
begin
  raise ArgumentError.new('System Tests are available for Rails >= 5.1') if Rails::VERSION::STRING < '5.1'
rescue ArgumentError => e
  puts e.message
end
```

joshRpowell:
> No this can be independant, it just needs to be gated on Rails version

Thanks @JonRowe. Dropped a something to jumpstart the conversation again.

benoittgt:
Closed via: https://github.com/rspec/rspec-rails/pull/2075 thanks @andrzejsliwa and @joshRpowell 🎉

PR #1938, opened 2017-12-31
Upgrade yard to 0.9.12, motivated by CVE.

xaviershay:


xaviershay:
I don't know WTF is going on with the build but I don't think it's green on master.

pirj:
Rebased on `master`.

pirj:
Doc diff is 2k+ lines, needs to be closely looked at.

JonRowe:
😬 Can you push the diff up to a branch on rspec/rspec.github.io ?

pirj:
@JonRowe Pushed [here](https://github.com/rspec/rspec.github.io/tree/yard-0.9.20/source/documentation/3.9/rspec-rails).

JonRowe:
@pirj there is a task in the `rspec-dev` repo for creating docs, if you use that the diff size will reduce as a lot of lines will be incorrectly linking to assets

pirj:
There might be some difference due to:
```
[error]: Error loading plugin 'yard-rspec-docs-template'
```

I'll get back to this a bit later.

JonRowe:
Heh, that might not be relevant :)

pirj:
@JonRowe [Here](https://github.com/rspec/rspec.github.io/commit/523121adde0b446075211e43cb2b7b72feb02288) it is.

Gemfile
```diff
@@ -7,7 +7,7 @@ gemspec
 rspec_dependencies_gemfile = File.expand_path("../Gemfile-rspec-dependencies", __FILE__)
 eval_gemfile rspec_dependencies_gemfile
 
-gem 'yard', '~> 0.8.7', require: false
+gem 'yard', '~> 0.9.20', :require => false
```
benoittgt: I think we can now jump to 0.9.22 :)

JonRowe: Same problem as before though

benoittgt:
So the issue is that the diff is too big? 

I ran the command `bundle exec rake update_docs` from `rspec-dev` but the diff is way too big. 🧐
https://github.com/rspec/rspec.github.io/compare/re-enable-yard-check?expand=1

I didn't specify release tag in `rake` command argument, but even without release tags I think I miss something.


Also I may have found a way to get rid of this error https://github.com/rspec/rspec-rails/pull/1938#issuecomment-563269962 by installing the gem from `rspec.github.io` `gem build` then `gem install`. But I am not sure.

JonRowe:
The diffs are large because they're html and the format will have changed, whats unfortunately needed is to check that our docs are still valid, its probably easier to compare the rendered result than the diff itself...

benoittgt:
I opened a draft PR with the changes. https://github.com/rspec/rspec.github.io/pull/131

As I mentioned in the draft PR. I am wondering if this PR should go to maintenance branch instead of master if we regenerate the documentation.

JonRowe:
For rspec-rails it will need to go to `3-9-maintenance` for sure.

benoittgt:
Closed by https://github.com/rspec/rspec-rails/pull/2285

PR #1942, opened 2018-01-06
delegate find_all directly

ahorek:
this will allow ruby to delegate find_all method more effectively

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -71,6 +71,11 @@ module RSpec
             @resolver = resolver
           end
 
+          def find_all(*args, &block)
+            result = @resolver.find_all(*args, &block)
+            nullify_templates(result)
+          end
+
           def method_missing(name, *args, &block)
             result = @resolver.send(name, *args, &block)
             nullify_templates(result)


JonRowe:
Hi @ahorek can you elaborate on the problem you're trying to solve is? In general yes defining methods is more specific and faster, but if we did that everywhere we'd have a huge maintenance burden in freezing ourselves to Rails, so unless this fixes something I'm inclined to close it.

ahorek:
hi @JonRowe, it's a micro-opt, but I think it's safe. This rails api wasn't changed for a very long time and even if it will be, it won't brake anything, otherwise I wouldn't even propose it. It's just an alias for the most used scenario. So I don't think there will be any maintanance burden due to the rails api.
In my case (big app) this method is called 50 times for each test case scenario, so it's used heavily.

Method missing isn't for free. Another option is a lazy method definition, it's the best practice to use a delegator, but I think it's useless for this case, because we know the api, we can define this method right away.
```ruby
          def method_missing(name, *args, &block)
            if @resolver.respond_to?(name)
              self.class.class_eval <<-end_eval
                def #{name}(*args, &block)
                  result = @resolver.__send__(:#{name}, *args, &block)
                  nullify_templates(result)
                end
              end_eval
              __send__(name, *args, &block)
            else
              super
            end
          end 
```
I can't use a delegator api which could be simplier because it's available since ruby 2.0 and this gem still support ancient rubies right?

On the other hand, even if this is more effective and compatible, the performance benefit may not worth for 4 more (unnecessary) lines of code. I understand if you decide to decline this proposal.

JonRowe:
Given that this is a pretty minor optimisation and causes some duplication, I'm going to close this. Thanks very much for taking the time to contribute, and please contribute again soon!

Issue #1857, opened 2017-08-09
Rails 6 will remove success? predicate in favor of successful?


ybart:
While running against Rails 5.2.0.alpha, I encountered the following message when running my test suite:

```
.DEPRECATION WARNING: The success? predicate is deprecated and will be removed in Rails 6.0. Please use successful? as provided by Rack::Response::Helpers. (called from matches? at rspec/rails/matchers/have_http_status.rb:263)
```

The message is coming from https://github.com/rails/rails/pull/30104 (https://github.com/rails/rails/commit/af3500b18817a48d7ec83812a0d4869f3fe33b2f)

trev:
Hi @ybart, happy to help with this. What's your question?

ybart:
I'm not sure how to deal with this. Maybe we may edit https://github.com/rspec/rspec-rails/blob/f33dde7519433f59fa21778aba953f2237dfacdb/lib/rspec/rails/matchers/have_http_status.rb#L245 to allow statuses referenced here http://www.rubydoc.info/gems/rack/Rack/Response/Helpers as well.

trev:
Basically the whole [`GenericStatus`](https://github.com/rspec/rspec-rails/blob/f33dde7519433f59fa21778aba953f2237dfacdb/lib/rspec/rails/matchers/have_http_status.rb#L20) class functionality needs to be deprecated and removed in lockstep with Rails 6.0.

trev:
Probably something like:

```ruby
        def self.matcher_for_status(target)
          if GenericStatus.valid_statuses.include?(target)
            RSpec.deprecate(
              GenericStatus.valid_statuses.map { |c| "`:#{c}`" }.join(', '),
              replacement: "the Rack::Response::Helpers predicates as symbols",
            )
            GenericStatus.new(target)
          elsif Symbol === target
            SymbolicStatus.new(target)
          else
            NumericCode.new(target)
          end
        end
```

alternatively we could just leave as is since the Rails deprecation takes care of it.

In your case you're probably using something like `expect(response).to have_http_status(:success)` in your tests. To get rid of the deprecation just use `expect(response).to have_http_status(:successful)`

ybart:
Thanks for the answer, just tried that and got the following error :

``` 
1) HomeController GET #show returns http success
     Failure/Error: expect(response).to have_http_status(:successful)

     ArgumentError:
       Invalid HTTP status: :successful
[...]

PR #1945, opened 2018-01-15
Prevent "have_http_status" from using deprecated methods in Rails 5.2+

wbreeze:
Provides a minimum change to solve the problem of response status methods deprecated in Rails 5.2, as raised by issue https://github.com/rspec/rspec-rails/issues/1857

This does so without deprecating or changing any status matchers in RSpec, so a Rails 5.2 upgrade will be transparent, at least with respect to not causing deprecation warnings for the :success, :error, and :missing values of `has_http_status`.

Another change could support the new :successful, :server_error, and :not_found status symbols. Or if wanted, another commit to this change.


lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -255,12 +255,32 @@ def initialize(type)
...
+          # @api private
+          # Convert a symbol representing the expected response class into
+          #   the name of the test method to call on the response object
+          #
+          # @param response_symbol [Symbol] representing the expected http
+          #   response class
+          # @return [String] name of the response method, minus the ending
+          #   question mark
+          def response_method(response_symbol)
+            if 5 < ::Rails::VERSION::MAJOR ||
```
JonRowe: We prefer conditionally defining methods rather than running `if` checks like this at runtime, after all it won't change for the duration of the spec suite.

spec/rspec/rails/matchers/have_http_status_spec.rb
```diff
@@ -7,6 +7,28 @@ def create_response(opts = {})
     }
   end
 
+  shared_examples_for "does not use deprecated methods for Rails 5.2+" do
+    it "does not use deprecated method for Rails >= 5.2" do
+      if 5 < Rails::VERSION::MAJOR ||
```
JonRowe: Similarly might as well not define the test when not in the right rails version.

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -255,12 +255,32 @@ def initialize(type)
...
+                success: 'successful',
+                error: 'server_error',
+                missing: 'not_found'
+              }.fetch(response_symbol, response_symbol)
+            else
+              response_symbol
+            end
+          end
+
+          # @return [Boolean] value of response status method
```
JonRowe: Can we flesh this out a bit more back closer to it's original docs.

wbreeze:
I hope that these changes hit the target for you. I looked for examples in the code.

If they don't hit the target, do you have an example in mind where this style is followed? But hopefully I understood.

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -327,6 +337,33 @@ def type_codes
...
+          # Convert a symbol representing the expected response class into
+          #   the name of the test method to call on the response object
+          #
+          # @param response_symbol [Symbol] representing the expected http
+          #   response class
+          # @return [String] name of the response method, minus the ending
+          #   question mark
+          def response_method(response_symbol)
+            {
+              success: 'successful',
```
JonRowe: I'm wondering if we should match our naming conventions to Rails going forward

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -18,7 +18,12 @@ module HaveHttpStatus
         # @return response matcher instance
         def self.matcher_for_status(target)
           if GenericStatus.valid_statuses.include?(target)
-            GenericStatus.new(target)
+            if 5 < ::Rails::VERSION::MAJOR ||
+               (::Rails::VERSION::MAJOR == 5 && 2 <= ::Rails::VERSION::MINOR)
```
JonRowe: This is not quite what I meant, I meant conditionally defining the method on the status itself to match the new names, so that we're not doing version checks all the time.

spec/rspec/rails/matchers/have_http_status_spec.rb
```diff
@@ -434,4 +434,46 @@ def create_response(opts = {})
...
+          expect(splitter.has_output?).to be false
+        ensure
+          $stderr = previous_stderr
+        end
+      end
+    end
+
+    context 'http status :missing' do
+      it_behaves_like "does not use deprecated methods for Rails 5.2+" do
+        puts "RAILS VERSION #{Rails.version}"
```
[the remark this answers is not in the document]
wbreeze: Gah. Sorry

lib/rspec/rails/matchers/have_http_status.rb
wbreeze: I'm trying to make the minimum change to get over the deprecation warnings. I'm trying not to be progressive about the function -- new naming conventions and such. Unfortunately, before 5.2 does not have successful (is that right?) 5.2 and after squawks if you don't use it.

lib/rspec/rails/matchers/have_http_status.rb
wbreeze: I don't like version checks. But I also don't like breaking tests for a whole lot of people because of a name change. I think that would mean at least a minor version upgrade, not a maintenance patch. I'm trying to get by with a maintenance patch. But perhaps I'm not understanding well the strategy you would prefer, and I'm entirely missing the boat.

wbreeze:
This is a request to merge to 3-7-maintenance, from which the branch is based, not to master. My idea was to allow a maintenance patch. If it's the strategy you want going to 3.8 or later, then it can also be merged upstream into master; but, that isn't what I'm asking to do. And I'm hearing that you want to go a different route in the master branch.

JonRowe:
Ah, well we'd want this based of `master` please, we don't generally accept PRs to the maintenance branches, we backport bug fixes ourselves but not enchancements like this.

wbreeze:
I don't think we have the same world view. I am trying to avoid a release that causes anyone using 5.2 to have to change all of their tests. The fix to do that would be different than a fix that either:

- requires 5.2 or later, or
- requires anyone upgrading to 5.2 to change all of their :success tests in order to avoid the deprecation warnings

Either one of those would require a minor or major version release. I don't think that backporting a change will do the trick. It isn't a bug. It's a migration strategy.

I don't think that I'm understanding your migration plan for this change in Rails. Would you explain it? Understanding it, I can either help you or get out of the way.

wbreeze:
PR #1951 references master, as you asked, without arguing. It still has the version check.

wbreeze:
Closing because this PR now goes to the wrong branch.

Issue #586, opened 2012-08-01
RAILS_ENV gets set to 'development' (not 'test') when running rake spec


joehorsnell:
Hi,

When running "rake spec" with a Rails 3.2.7 app and RSpec 2.11.0, Rails.env is getting set to 'development', rather than defaulting to 'test'. The spec_helper has the ENV["RAILS_ENV"] ||= 'test' line right at the top (as generated by "rails g rspec:install") but it's too late at that point, that environment variable has already been set.

I will zip and attach (if possible) a sample Rails app to recreate the issue, but a bit of debugging shows that the RAILS_ENV environment variable is being set from the following before spec_helper is loaded:

Setting RAILS_ENV = development
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/gems/1.9.1/gems/railties-3.2.7/lib/rails/tasks/misc.rake:4:in `block in <top (required)>'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:205:in`call'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:205:in `block in execute'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:200:in`each'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:200:in `execute'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:158:in`block in invoke_with_call_chain'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/monitor.rb:211:in `mon_synchronize'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:151:in`invoke_with_call_chain'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:176:in `block in invoke_prerequisites'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:174:in`each'
/Users/joe/.rbenv/versions/1.9.3-p194/lib/ruby/1.9.1/rake/task.rb:174:in `invoke_prerequisites'
[...]

Issue #70, opened 2010-06-06
RSpec not setting environment to test


samflores:
I'm trying Rspec-2.0.0.beta.10 with Rails-3.0.0.beta3 and Ruby-1.9.2-preview3 and my environment is not getting set to 'test' as it should. I tryed using the spec command, rake spec and autotest but both Rails.env and ENV['RAILS_ENV'] are always 'development'.

At first I thought it was an Mongoid issue, but I created a new project using ActiveRecord and sqlite3 and the problem still exists.

PR #1947, opened 2018-01-22
Set RAILS_ENV to test in rails_helper

rodrigobdz:
This is a solution to tests running in development environment as stated in issues #70 and #586.
Before Rails 3.2.7, `ENV['RAILS_ENV'] ||= 'test'` ran tests without issues in test env as discussed in https://github.com/rails/rails/issues/7175#issue-5869126.

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -1,6 +1,6 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require 'spec_helper'
-ENV['RAILS_ENV'] ||= 'test'
+ENV['RAILS_ENV'] = 'test'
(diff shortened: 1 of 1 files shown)


mikegee:
I always make this change in my `rails_helper`. The justification in #70 seems kinda weak:

> I don't want to force it to 'test' because that ties everybody's hands.

This file is in my app, not in RSpec.

JonRowe:
The justification behind this is that it's a template, and this is the friendliest, safest option. Its perfect ok to modify in your own projects, but it's not our recommended thing to do, thus closing. Thanks for taking the time to contribute :)

rodrigobdz:
It should at least be warned about its implications in the [README](https://github.com/rspec/rspec-rails/blob/master/README.md). That caused my tests to run in development environment and the entire database was wiped out.

PR #1935, opened 2017-12-30
Rake 12.3 cannot install on Ruby 1.9.3.

myronmarston:

PR #1949, opened 2018-01-26
Update Gemfile to correctly install rake per ruby version.

JonRowe:
We do this on rspec/rspec-core and I believe it's the source of the cron build failures on expectations and support. Care to review @rspec/rspec ?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,10 +28,12 @@ if RUBY_VERSION >= '2.4.0'
   gem 'json', '>= 2.0.2'
 end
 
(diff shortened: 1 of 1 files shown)


xaviershay:
LGTM, though lol the build failed :)

myronmarston:
LGTM...after all, this is what I had tried to do in #1935 :).

JonRowe:
@myronmarston doh! Going to merge this as the build is broken regardless but I prefer this approach over sticking with Rake 10, will try @samphippen's bundler fix on my `fix-ci` branch

PR #1950, opened 2018-01-26
Pin bundler to fix CI

JonRowe:


JonRowe:
Following on from @samphippen's suggestion in #1935 I hypothesised that if we "down grade" bundler we could fix our build, and it worked, thoughts about merging this and then working on fixing the bundler issue later on? @rspec/rspec?

I got a green build with the version of bundler we use elsewhere, but I've pushed up a commit updating this for `rspec-rails` to `1.15.x` 🤞 that goes green too...

JonRowe:
Woohoo! Green build, can someone review @rspec/rspec ?

JonRowe:
@samphippen @myronmarston can I get a quick review?

PR #1951, opened 2018-01-31
Prevent "have_http_status" from using deprecated methods in Rails 5.2+

wbreeze:
This alternative PR (to #1945) Provides a minimum change to solve the problem of response status methods deprecated in Rails 5.2, as raised by issue #1857

This does so without deprecating or changing any status matchers in RSpec, so a Rails 5.2 upgrade will be transparent, at least with respect to not causing deprecation warnings for the :success, :error, and :missing values of has_http_status.

A major or minor release that eliminates the version check will either require Rails 5.2 and map :success to the `successful` call, or require everyone to change their tests.

### DIFF
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -289,8 +289,22 @@ module RSpec
 
         protected
 
-          def check_expected_status(test_response, expected)
-            test_response.send("#{expected}?")
+          if 5 < ::Rails::VERSION::MAJOR ||
+             (::Rails::VERSION::MAJOR == 5 && 2 <= ::Rails::VERSION::MINOR)
+            def check_expected_status(test_response, expected)
+              test_response.send("#{expected}?")
+            end
+          else
+            RESPONSE_METHODS = {
+              success: 'successful',
+              error: 'server_error',
+              missing: 'not_found'
+            }.freeze
+
+            def check_expected_status(test_response, expected)
+              test_response.send(
+                "{RESPONSE_METHODS.fetch(response_symbol, response_symbol)}?")
+            end
           end
 
         private
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -18,7 +18,12 @@ module RSpec
         # @return response matcher instance
         def self.matcher_for_status(target)
           if GenericStatus.valid_statuses.include?(target)
-            GenericStatus.new(target)
+            if 5 < ::Rails::VERSION::MAJOR ||
+               (::Rails::VERSION::MAJOR == 5 && 2 <= ::Rails::VERSION::MINOR)
+              GenericStatusWithLookup.new(target)
+            else
+              GenericStatus.new(target)
+            end
           elsif Symbol === target
             SymbolicStatus.new(target)
           else
@@ -255,12 +260,11 @@ module RSpec
             @invalid_response = nil
           end
 
-          # @return [Boolean] `true` if Rack's associated numeric HTTP code matched
-          #   the `response` code
+          # @return [Boolean] value of response status method
           def matches?(response)
             test_response = as_test_response(response)
             @actual = test_response.response_code
-            test_response.send("#{expected}?")
+            check_expected_status(test_response, expected)
           rescue TypeError => _ignored
             @invalid_response = response
             false
@@ -281,6 +285,12 @@ module RSpec
           def failure_message_when_negated
             invalid_response_type_message ||
             "expected the response not to have #{type_message} but it was #{actual}"
+          end
+
+        protected
+
+          def check_expected_status(test_response, expected)
+            test_response.send("#{expected}?")
           end
 
         private
@@ -325,6 +335,33 @@ module RSpec
                             when :redirect
                               "3xx"
                             end
+          end
+        end
+
+        class GenericStatusWithLookup < GenericStatus
+
+        protected
+
+          def check_expected_status(test_response, expected)
+            test_response.send("#{response_method(expected)}?")
+          end
+
+        private
+
+          # @api private
+          # Convert a symbol representing the expected response class into
+          #   the name of the test method to call on the response object
+          #
+          # @param response_symbol [Symbol] representing the expected http
+          #   response class
+          # @return [String] name of the response method, minus the ending
+          #   question mark
+          def response_method(response_symbol)
+            {
+              success: 'successful',
+              error: 'server_error',
+              missing: 'not_found'
+            }.fetch(response_symbol, response_symbol)
           end
         end
       end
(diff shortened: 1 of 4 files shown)


JonRowe:
You'll need to rebase this so you don't bring across a load of `3-7-maintenance` commits across to master

wbreeze:
Thank you for prompting that, Jon. Agreed and done. The commit history looked terrible, but the diff was okay. With your prompting, I rebased, squashed, and force pushed. Now you have one commit to backport if you want it.

My view of the migration strategy is:
- Now: Patch or minor release with the version check
- With Rails 6 or someday: major release that requires 5.2 (or later), eliminates the version check, and maps the old syntax to the new method.

With that, no-one has to port forty tests when they upgrade. (I can only guess the number of extant `has_http_status :success` tests, but doubt that it's a trivial count.)

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -18,7 +18,12 @@ module HaveHttpStatus
         # @return response matcher instance
         def self.matcher_for_status(target)
           if GenericStatus.valid_statuses.include?(target)
```
wbreeze: Rather than check valid statuses and then doing the mapping, this could do the mapping, then check valid statuses. `status_method = GenericStatus.map_status(target); if (status_method) ...` Then GenericStatus encapsulates all of the logic. I'm willing to make that change.

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -255,12 +260,11 @@ def initialize(type)
             @invalid_response = nil
           end
 
-          # @return [Boolean] `true` if Rack's associated numeric HTTP code matched
-          #   the `response` code
+          # @return [Boolean] value of response status method
```
JonRowe: I'd like to see the original doc restored, with maybe an additional " or the named response status matches"

lib/rspec/rails/matchers/have_http_status.rb
JonRowe: I'd still like to see this class done away with, and the lookup method defined depending on the version of rails. See below.

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -283,6 +287,12 @@ def failure_message_when_negated
             "expected the response not to have #{type_message} but it was #{actual}"
           end
 
+        protected
+
+          def check_expected_status(test_response, expected)
+            test_response.send("#{expected}?")
+          end
```
JonRowe: Basically this would become something like:

```Ruby
if if 5 < ::Rails::VERSION::MAJOR || (::Rails::VERSION::MAJOR == 5 && 2 <= ::Rails::VERSION::MINOR)
  def check_expected_status(test_response, expected)
    test_response.send("#{expected}?")
  end
else
  RESPONSE_METHODS = {
    success: 'successful',
    error: 'server_error',
    missing: 'not_found'
  }.freeze

  def check_expected_status(test_response, expected)
     test_response.send("#{RESPONSE_METHODS.fetch(response_symbol, response_symbol)}?")
  end
end
```

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -327,6 +337,33 @@ def type_codes
                             end
           end
         end
+
+        class GenericStatusWithLookup < GenericStatus
```
JonRowe: And this would go away.

wbreeze:
Jon:

Thank you for the further guidance on the change that you would like to see.

I'm afraid I went a little further, and made the change forward compatible as well as backward compatible. Allowing the new syntax for any version didn't seem right unless it worked for any version.

The two variants failing on Travis are failing in build.

wbreeze:
Thank you for prompting those additional tests. What could go wrong? It turns out that the type codes came back empty in the messages for those new generic methods.

I couldn't resist drying up the tests with a shared_examples block. They were very repetitive.

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -243,7 +243,11 @@ class GenericStatus < RSpec::Matchers::BuiltIn::BaseMatcher
           #   code "group"
           # @see https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/testing/test_response.rb `ActionDispatch::TestResponse`
           def self.valid_statuses
-            [:error, :success, :missing, :redirect]
+            [
+              :error, :success, :missing,
+              :server_error, :successful, :not_found,
+              :redirect
```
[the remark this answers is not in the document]
wbreeze: Thank you for prompting those additional tests. What could go wrong? It turns out that the type codes came back empty in the messages for those new generic methods.

I couldn't resist drying up the tests with a shared_examples block. They were very repetitive.

spec/rspec/rails/matchers/have_http_status_spec.rb
```diff
@@ -177,6 +177,51 @@ def create_response(opts = {})
...
+    end
+
+    it "has a negated failure message reporting the expected and actual status codes" do
+      response = create_response(:status => expected_code)
+      expect{ matcher.matches? response }.
+        to change(matcher, :failure_message_when_negated).
+        to(negated_message)
+    end
+  end
+
```
wbreeze: The diff from here out looks terrible. All that happened is:

- the repeated code in this shared example was replaced with invocation of this shared example
- new tests using the shared example were added for :not_found, :server_error, and :successful

spec/rspec/rails/matchers/have_http_status_spec.rb
```diff
@@ -381,57 +408,92 @@ def create_response(opts = {})
...
 
-    describe "matching a response" do
-      it "returns true for a response with a 3xx status code" do
-        any_3xx_code = 308
-        response     = create_response(:status => any_3xx_code)
+  context "with a nil status" do
+    it "raises an ArgumentError" do
+      expect{ have_http_status(nil) }.to raise_error ArgumentError
+    end
+  end
```
wbreeze: Everything from here on is actually new, not changed lines.

wbreeze:
@JonRowe The two failing combinations are gem dependency errors from bundler. If you hate the refactoring, I'll redo it copy-paste.

Let me know if you would like any further changes.

JonRowe:
Thanks for the effort getting this over the line!

jesperronn:
In my opinion, this calls for a release, since the deprecation warning is REALLY annoying when upgrading rails projects.

I would suggest a minor version bump, since it's an added feature for Rails 5.2 and it doesn't break existing behaviour.

jfi:
Any news on a release for this?

Aesthetikx:
@samphippen Sorry to bump but I would also appreciate a point release including this for our Rails 5.2 upgrade :heart:

LeeXGreen:
I would also really appreciate a point release with this fix.

We're currently pointing to 3.8.pre via Git SHAs, but it would be much nicer to just pull this from Rubygems.

PR #1953, opened 2018-02-09
improve documentation - add common example

BKSpurgeon:
* it's too common for 401 not be added hence its inclusion

* This documentation renders as intended. It is also tested against my rspec code.

### DIFF
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -233,6 +233,7 @@ module RSpec
         #   expect(response).to have_http_status(:error)
         #   expect(response).to have_http_status(:missing)
         #   expect(response).to have_http_status(:redirect)
+        #   expect(response).to have_http_status(:unauthorized)   
         #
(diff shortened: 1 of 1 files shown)


lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -233,6 +233,7 @@ def set_expected_code!
         #   expect(response).to have_http_status(:error)
         #   expect(response).to have_http_status(:missing)
         #   expect(response).to have_http_status(:redirect)
+        #   expect(response).to have_http_status(:unauthorized)
```
JonRowe: There is trailing whitespace on this line which is why the build is failing.

BKSpurgeon: Ok. Will fix - I didn't know that a comment would fail the test.

JonRowe: Its not the comment that fails the test, its the whitespace, we have linters :)

BKSpurgeon:
will submit a new pull request.

PR #1957, opened 2018-02-19
Refactor spec for have_http_status.rb

BKSpurgeon:
**What are the proposed changes?** 

Refactor the spec file – eliminate needless and painful repetition to simply reading and understanding the spec.

**Checklist:**

- New behavior is covered by tests and all tests are passing - well at least the changes i made are passing. this is my first pull request to rspec so i'll probably need some pointers.
- No warnings (that I'm aware of) results from changes to the spec files.
- There are no changes to the documentation.
- Can confirm: Rubocop passes (e.g. bundle exec rubocop lib).
- I've squashed all the commits to just one commit.

### DIFF
--- a/spec/rspec/rails/matchers/have_http_status_spec.rb
+++ b/spec/rspec/rails/matchers/have_http_status_spec.rb
@@ -177,210 +177,74 @@ RSpec.describe "have_http_status" do
     end
   end
 
-  context "with general status code group", ":error" do
-    # The error query is an alias for `server_error?`:
-    #
-    # - https://github.com/rails/rails/blob/ca200378/actionpack/lib/action_dispatch/testing/test_response.rb#L27
-    # - https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/testing/test_response.rb
-    #
-    # `server_error?` is part of the Rack Helpers and is defined as:
-    #
-    #     status >= 500 && status < 600
-    #
-    # See:
-    #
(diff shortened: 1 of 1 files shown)


BKSpurgeon:
hi folks it's failing on some code pertaining to the mailer - amongst other things - but i haven't touched that code. Some advice on how to get it to pass would be much appreciated.

spec/rspec/rails/matchers/have_http_status_spec.rb
```diff
@@ -177,210 +177,74 @@ def create_response(opts = {})
...
-        expect( have_error_status.matches?(response) ).to be(true)
-      end
-
-      it "returns false for a response with a different code" do
-        client_error_code = 400
-        response          = create_response(:status => client_error_code)
-
-        expect( have_error_status.matches?(response) ).to be(false)
+    it "returns true for a response which starts with the same number (i.e. 4XX and 4ZZ) " do
+      if code == 404
```
JonRowe: We don't want if's inside our tests, we like the seperate tests, you've not removed duplication here.

JonRowe:
Sorry this doesn't make things more readable at all, not all repetition is bad, the fact you had to add conditionals to make this work shows that its not as repeated as you think.

Additionally there has been work over in #1951 which will supersedes this so its not worth continuing right now. After #1951 is merged if you think there are improvements that can be made by all means open another PR based off the result of that.

PR #1960, opened 2018-03-02
Update rails_helper.rb template

coorasse:
The following change follows best practices of Rails.root.join and avoids linting errors when uncommented. See also http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/FilePath


JonRowe:
Thanks

PR #1965, opened 2018-03-12
Remove the version check in have_http_status

wbreeze:
The successful?, server_error?, and not_found? methods for ActionDispatch::TestResponse have existed in Rails for a very long time. It isn't necessary to check for Rails 5.2 or greater in order to use them.

This is an improvement on https://github.com/rspec/rspec-rails/pull/1951 which has been merged.

### DIFF
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -289,20 +289,11 @@ module RSpec
 
         protected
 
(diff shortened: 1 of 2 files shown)


wbreeze:
@JonRowe So happy to be able to do this.

JonRowe:
Thanks, sorry this slipped through the cracks

benoittgt:
Thanks!

PR #1966, opened 2018-03-14
Add fscenario alias to focus example

swebb:
This alias adds the focus: true metadata to scenarios. Most users
receive get this functionality via the Capybara gem; however, this
patch makes it available to all rspec-rails users.

I'm not sure how to test this change. Capybara uses [this scenario](https://github.com/teamcapybara/capybara/blob/e6e132ca636f1e81b944066250aa302348cfdf50/spec/rspec/scenarios_spec.rb#L10). However, because rspec-rails has [filter_run: :focus](https://github.com/rspec/rspec-rails/blob/1d2935861c89246236b46f77de753cda5ea67d61/spec/spec_helper.rb#L23) that won't work.


JonRowe:
You can add a test for this by creating an example using the alias and checking it's metadata

swebb:
Closing as I'm not likely to have time to work on this.

PR #1967, opened 2018-03-16
RFC: versioning strategy

samphippen:
[view rendered](https://github.com/rspec/rspec-rails/blob/rfc-versioning-strategy/rfcs/versioning-strategy.md)

Please leave comments on the PR. The render link will update as new commits get pushed.

### DIFF
--- a/rfcs/versioning-strategy.md
+++ b/rfcs/versioning-strategy.md
@@ -1,0 +1,119 @@
+
+# RFC: RSpec Rails new versioning strategy
+
+Hi Folks,
+
+This RFC captures a proposal for RSpec Rails' new versioning strategy. Specifically, this represents two things:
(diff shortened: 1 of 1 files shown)


myronmarston:
:+1:

> For the purposes of this versioning strategy, we consider removing a Rails or Ruby version to be a breaking change. 

It's worth calling out that we're discussing changing this for RSpec (see rspec/rspec#25).  Not sure what affect (if any) that should have on rspec-rails, though.

JonRowe:
I entirely support this, my only comment is how will `rspec-rails` handle major releases of RSpec going forward, and what affect if any RSpec dropping ruby version support (as discussing in rspec/rspec#25) have on this? (I think none as Rails only supports in life Ruby?)

samphippen:
I think we can consider this accepted. Merging.

PR #1977, opened 2018-03-27
Add support for asserting job is enqueued with no wait.

bcharna:


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -149,12 +149,17 @@ def other_attributes_match?(job)
 
           def serialized_attributes
             {}.tap do |attributes|
-              attributes[:at]    = @at.to_f if @at
+              attributes[:at]    = serialized_at
```
JonRowe: This is not quite the same, can we do  `serialized_at if @at` and eliminate the nil check in the method, we don't want to se this at all if it hasn't been set.

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -196,6 +196,19 @@ def self.name; "LoggingJob"; end
...
+        hello_job.perform_later
+      }.to have_enqueued_job.at(:no_wait)
+    end
+
+    it "does not have an enqueued job when providing at of :no_wait and there is a wait" do
+      date = Date.tomorrow.noon
+      expect {
+        hello_job.set(:wait_until => date).perform_later
+      }.to_not have_enqueued_job.at(:no_wait)
+    end
```
JonRowe: There's another scenario here, not setting at `at` all and there being a wait

bcharna: good callouts 👍

PR #1979, opened 2018-03-30
Cache ActionView::FixtureResolvers created by stub_template

urbanautomaton:
The `#stub_template` view spec helper works by prepending an [`ActionView::FixtureResolver`](https://github.com/rails/rails/blob/9f65d2a08bc80a94bbb2c0b6e00957c7059aed25/actionview/lib/action_view/testing/resolvers.rb#L6-L9) to the view instance's `view_paths`. This intercepts queries for specific template files, providing a canned response that's used to instantiate a dummy `ActionView::Template`.

These canned responses still require compilation and [finalization](https://github.com/rails/rails/blob/9f65d2a08bc80a94bbb2c0b6e00957c7059aed25/actionview/lib/action_view/template.rb#L118-L126), however. As the resolver holds all the references to `ActionView::Template` instances, and a new resolver is created for each stub in every example, this means that a single `stub_template` call in a `before` block covering 20 examples will result in the same template being compiled and finalized 20 times.

Both compilation and finalization can be expensive. To avoid this expense where possible, this change adds a global resolver cache for stubbed templates.

By way of anecdote ('cos everyone loves those), this dropped our ~4k example view spec suite with ~500 `stub_template` calls from 6 min to 2 min total runtime.

If there's anything you'd like done differently, or if there are any negative consequences of this change I haven't thought of, just let me know; I'll be happy to update the PR accordingly.

Cheers,
Simon

edit: ugh, the [only build job that failed](https://travis-ci.org/rspec/rspec-rails/jobs/360415431) looks transient...

### DIFF
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -14,6 +14,14 @@ module RSpec
       include RSpec::Rails::Matchers::RenderTemplate
 
       # @private
+      module StubResolverCache
+        def self.resolver_for(hash)
+          @resolvers ||= {}
+          @resolvers[hash] ||= ActionView::FixtureResolver.new(hash)
+        end
+      end
+
+      # @private
       module ClassMethods
         def _default_helper
           base = metadata[:description].split('/')[0..-2].join('/')
@@ -84,7 +92,7 @@ module RSpec
         #
         #     stub_template("widgets/_widget.html.erb" => "This content.")
         def stub_template(hash)
-          view.view_paths.unshift(ActionView::FixtureResolver.new(hash))
+          view.view_paths.unshift(StubResolverCache.resolver_for(hash))
         end
 
         # Provides access to the params hash that will be available within the
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -254,5 +254,25 @@ module RSpec::Rails
         view_spec.template
       end
     end
+
+    describe '#stub_template' do
+      let(:view_spec) do
+        Class.new do
+          include ViewExampleGroup::ExampleMethods
+          def _view
+            @_view ||= Struct.new(:view_paths).new(['some-path'])
+          end
+        end.new
+      end
+
+      it 'prepends an ActionView::FixtureResolver to the view path' do
+        view_spec.stub_template('some_path/some_template' => 'stubbed-contents')
+
+        result = view_spec.view.view_paths.first
+
+        expect(result).to be_instance_of(ActionView::FixtureResolver)
+        expect(result.hash).to eq('some_path/some_template' => 'stubbed-contents')
+      end
+    end
   end
 end
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -256,22 +256,33 @@ module RSpec::Rails
     end
 
     describe '#stub_template' do
-      let(:view_spec) do
+      let(:view_spec_group) do
         Class.new do
           include ViewExampleGroup::ExampleMethods
           def _view
             @_view ||= Struct.new(:view_paths).new(['some-path'])
           end
-        end.new
+        end
       end
 
       it 'prepends an ActionView::FixtureResolver to the view path' do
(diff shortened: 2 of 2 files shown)


lib/rspec/rails/example/view_example_group.rb
```diff
@@ -13,6 +13,14 @@ module ViewExampleGroup
       include RSpec::Rails::ViewAssigns
       include RSpec::Rails::Matchers::RenderTemplate
 
+      # @private
+      module StubResolverCache
+        def self.resolver_for(hash)
+          @resolvers ||= {}
+          @resolvers[hash] ||= ActionView::FixtureResolver.new(hash)
```
JonRowe: My only concern is that this introduces shared state across tests (I know this is by design) but if we were to freeze this object it wouldn't be mutable which would alleviate this concern

urbanautomaton: Sure, that's a straightforward change to make, have pushed with the extra `.freeze`. 👍 

However, there's some further mutable internal state in the resolver that I don't think we'll be able to freeze; it only builds and caches `Template` instances when it receives queries for a specific path, so this happens on first render, not when we're instantiating the resolver.

My intuition is that this shared state should be safe, though. Since we're indexing the cache on the full stub definition (i.e. filename and stubbed contents), and the output of a compiled template should be a function of its contents, I don't think there's any way for the cache to return incorrect values.

urbanautomaton:
Thanks for the feedback @JonRowe - I've made the requested change.

I also got bored and wrote a [small repro](https://github.com/urbanautomaton/stub_template_perf) that constructs a pathological case for the performance problem the caching addresses. The caching turns a 4 minute run into a 13s one, which is nice. 😄

urbanautomaton:
Ah; it seems that freezing the `FixtureResolver` isn't an option in Rails 3.0.20 / ruby 1.9 (Failing jobs [1](https://travis-ci.org/rspec/rspec-rails/jobs/360590385#L4685), [2](https://travis-ci.org/rspec/rspec-rails/jobs/360590389#L4641)).

On the grounds that freezing it doesn't actually prevent its internal state from being mutated anyway, I'd be tempted to argue that it's okay to leave it unfrozen, for the reasons expressed earlier. By way of comparison, the `OptimizedFileSystemResolver` instances that constitute the default `view_paths` are already shared between examples, seemingly without problems. What do you think?

JonRowe:
My hesitation is I have no idea if there will be any unintended consequences for allowing it to be mutated, I believe our `OptimisedFileSystemResolver` took it into consideration from the outset. If you're more familiar with the `FixtureResolver` internals to say it won't be a problem I'm happy to accept your judgement :)

urbanautomaton:
I definitely share your concerns; it'd be nice to have more confidence than just a code reading to assure us this is safe.

From the code reading perspective: the only significant internal state to both [`ActionView::OptimizedFileSystemResolver`](https://github.com/rails/rails/blob/50bfb263a130d781e992a1933589a451d54ba212/actionview/lib/action_view/template/resolver.rb#L364-L378) and [`ActionView::FixtureResolver`](https://github.com/rails/rails/blob/50bfb263a130d781e992a1933589a451d54ba212/actionview/lib/action_view/testing/resolvers.rb#L10-L46) is the resolver's [template cache itself](https://github.com/rails/rails/blob/50bfb263a130d781e992a1933589a451d54ba212/actionview/lib/action_view/template/resolver.rb#L39-L128). This is built into the base `ActionView::Resolver` class from which both resolvers ultimately inherit, so my understanding is that there shouldn't be any difference, state-wise, in behaviour between the two resolvers; fundamentally they only differ in how they map input paths to template source strings, and the rest of their functionality is shared. So any state mutation problem that affects `FixtureResolver` *ought* to affect `OptimizedFileSystemResolver` too. FWIW the latter is defined in Rails itself, so as best I can tell it's not had any specific attention paid to this issue.

So I'm pretty confident it's okay, but that's obviously not the same as "there can't be a problem".

I've monkey-patched this change into the app I'm currently working on. While this wouldn't conclusively prove the absence of problems either, would it be worth sitting on this PR for a week or so to see if any issues crop up in our heavily-trafficked build?

urbanautomaton:
Okeydoke - thanks very much! ❤️

PR #1986, opened 2018-04-11
Test against the released version of Rails 5.2

### LINKED ISSUE #1985: Rails 5.2.0 support
Rails 5.2.0 is out and stable. Could you upgrade the Rails dependency version to 5.2?

mikegee:
fixes #1985


.travis.yml
```diff
@@ -141,7 +141,7 @@ matrix:
     - rvm: 2.4.4
       env: RAILS_VERSION='~> 3.1.12'
     - rvm: 2.4.4
-      env: RAILS_VERSION=5.2.0.rc1
+      env: RAILS_VERSION=5.2.0
```
JonRowe: Ah hem, this is in the wrong place, it actually prevents `5.2.0` from running on `2.4.4`, it should be include with an entry for each version it supports

mikegee: Indeed this appears to be at end of some exclusions. But, why is it here? I don’t see anything that would test against 5.2 in the first place. 

Looks like I should have added entries under the include on line 60.

JonRowe: Yes it's in the wrong place and no one noticed :)

mikegee:
@JonRowe I think Travis is running the expected jobs now. Please have a look.

mikegee:
I'm having a hard time deciphering the build failure. Does that stuff about system tests stop the spec run with a non-zero exit code?

JonRowe:
Yes it does, I'm not sure why it's being triggered, there must be some setup thats done for 5.1 that isn't working on 5.2 for some reason

Issue #1984, opened 2018-04-08
SourceAnnotationExtractor moved in edge Rails causing error


jasnow:
**Steps to Reproduce**

 1. Clone repo: https://github.com/jasnow/show-me-the-food5edge
 2. Run "bundle ; rake" (all green)
 3. Run "bundle update ; rake" and get this error:
```
Bundler::GemRequireError: There was an error while trying to load the gem 'rspec-rails'.
Gem Load Error is: Rails::SourceAnnotationExtractor is not a class/module
Backtrace for gem load error is:
```
 4. Think this is caused by this code change:    https://github.com/rails/rails/commit/0ec23effa74e8e9da6ebb14afbd335bea470ab3e
 5. Feel free to add comment for more information if needed.

Rails:   6.0.0.alpha
Ruby:   2.6.0-preview1

jasnow:
Checked today and it affects 4 of my "canary"/edge Rails repos.

nbulaj:
The same problem [here](https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/366415756). Even for all the RSpec stack from the `master`.

jasnow:
@JonRowe - Is there anything else you need from me regarding this issue?

JonRowe:
Try out #1989 and see if that fixes it for you, really someone should tell the Rails team about this too as the intent is to deprecate the class not break it.

jasnow:
I did the following:
 1. Changed Gemfile to use
```
gem 'rspec-rails', git: 'https://github.com/rspec/rspec-rails'
```
 2. Ran "bundle update" and got lots of dependency errors.

jasnow:
@JonRowe - Added a comment to Gitter. Can't cut/paste into Google Group and did not yet want to delete rails/rails issue template.

nbulaj:
I'm using master branch of Rspec with Rails master and still see this error:

https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/380759360

PR #1997, opened 2018-05-24
Fix SourceAnnotationExtractor on Rails 6

brian-kephart:
Without this fix, my tests fail when loading RSpec.
`uninitialized constant RSpec::Rails::SourceAnnotationExtractor`

With the fix, the constant is appropriately nested under the top-level `Rails` constant instead of trying to nest under `RSpec`, so my tests pass again.

### DIFF
--- a/lib/rspec-rails.rb
+++ b/lib/rspec-rails.rb
@@ -10,7 +10,7 @@ module RSpec
       if ::Rails::VERSION::STRING >= '5.1'
         require 'rails/source_annotation_extractor'
         if ::Rails::VERSION::STRING >= '6.0'
-          Rails::SourceAnnotationExtractor::Annotation.register_directories("spec")
+          ::Rails::SourceAnnotationExtractor::Annotation.register_directories("spec")
         else
           SourceAnnotationExtractor::Annotation.register_directories("spec")
         end


brian-kephart:
See build errors here: https://travis-ci.com/brian-kephart/camaleon-cms/jobs/125697021

And the last comment here: https://github.com/rspec/rspec-rails/issues/1984

brian-kephart:
When I submitted this PR, the rails-controller-testing gem would not load in Rails 6, so the effect of this PR on the test suite could not be evaluated directly. This changed allowed a repo that depends on rspec-rails to build correctly, which was indirect evidence of its effect.

Since then, rails-controller-testing has been updated, so now the tests can be run on Rails master (at least until another problem is encountered). See results [here](https://travis-ci.com/brian-kephart/rspec-rails/jobs/127532088).

With the updated rails-controller-testing gem but without the change from this PR, [this is the result](https://travis-ci.com/brian-kephart/rspec-rails/jobs/127526567#L1000).

nbulaj:
I'm using master branch of RSpec with Rails master and still see this error:

https://travis-ci.org/doorkeeper-gem/doorkeeper/jobs/380759360

JonRowe:
Thanks for this!

dsandstrom:
If you encounter a similar error after updating to Rails 6.1, try updating rspec to 3.10+ and rspec-rails to 4.0+.

PR #1998, opened 2018-06-02
Update and clean up Travis build

brian-kephart:
Sorry for the big commit, but these changes are all pretty inter-related. Splitting it up seemed like it would make things more complicated, not less.

1. Reworked .travis.yml. It used to be an auto-generated matrix, but the number of specific inclusions and exclusions made it confusing to try and update. I changed it to specifically list each build, ordered by Ruby version first and Rails version second.
2. Added Ruby 2.5 to supported Rails versions in .travis.yml.
3. Added stable git branches for Rails 5.1 and 5.2.
4. Removed redundant Rails git branches. Versions below 4.2 receive no updates, so testing these git branches was redundant with testing the released versions.
5. Updated each Ruby/Rails to latest patch version.
6. Fixed some dependency issues with newer Rails versions that were causing build errors.
7. Fixed an issue in the Rakefile that caused some versions to be tagged incorrectly, causing build errors.
8. Added a build for development version of Ruby using Rails 5.2 (failures allowed).
9. Removed tests of Ruby < 2.4 for Rails master.

Currently all released versions and stable branches build correctly. Rails master has a dependency issue causing failures, and Ruby-head has a segfault.

Please note that I did not intentionally remove any builds for old versions. The only builds I removed were git branches that were redundant with released gem versions. This PR is not meant to remove support for anything, only to improve coverage of new versions and maintainability of the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,129 +34,201 @@ before_script:
 
 script: "script/run_build 2>&1"
 
-rvm:
-  - 1.8.7
-  - 2.4.4
-  - 2.3.7
-  - 2.2
-  - 2.1
-  - 2.0.0
-  - 1.9.3
-  - 1.9.2
-
-env:
-  - RAILS_VERSION='~> 4.2.0'
-  - RAILS_VERSION=4-2-stable
-  - RAILS_VERSION='~> 4.1.0'
-  - RAILS_VERSION=4-1-stable
-  - RAILS_VERSION='~> 4.0.4'
-  - RAILS_VERSION=4-0-stable
-  - RAILS_VERSION='~> 3.2.17'
-  - RAILS_VERSION=3-2-stable
-  - RAILS_VERSION='~> 3.1.12'
-  - RAILS_VERSION='~> 3.0.20'
-
 matrix:
   include:
-    # Rails 5.x only supports 2.2+
-    - rvm: 2.2.2
-      env: RAILS_VERSION=master
-    - rvm: 2.2.2
-      env: RAILS_VERSION=5-0-stable
-    - rvm: 2.2.2
-      env: RAILS_VERSION=5.1.0
-    - rvm: 2.3.7
-      env: RAILS_VERSION=master
-    - rvm: 2.3.7
-      env: RAILS_VERSION=5-0-stable
-    - rvm: 2.3.7
-      env: RAILS_VERSION=5.1.0
-    - rvm: 2.3.7
-      env: RAILS_VERSION=5.2.0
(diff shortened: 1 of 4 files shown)


sikachu:
> Rails master has a dependency issue causing failures.

I recently submitted a patch to fix this (https://github.com/rails/rails-controller-testing/pull/42) and it got merged in today. Do you mind trying to re-run the build to see how it goes?

brian-kephart:
Thanks @sikachu for addressing that. See build log [here](https://travis-ci.com/brian-kephart/rspec-rails). Now rails-controller-testing loads, but building with Rails master still fails because of the problem addressed in #1997.

JonRowe:
Wow ok, so first of all thanks for taking the time to do this but I'm going to need some help here..

> Reworked .travis.yml. It used to be an auto-generated matrix, but the number of specific inclusions and exclusions made it confusing to try and update. I changed it to specifically list each build, ordered by Ruby version first and Rails version second.

Great, probably about time since we keep adding specific rails exclusions with each new version but I'd like this to be a seperate PR, and I'll need to verify no build version changes.

> Removed redundant Rails git branches. Versions below 4.2 receive no updates, so testing these git branches was redundant with testing the released versions.

Please restore them, the point is to prove our compatibility with Rails, not visa versa, we can introduce regressions in new versions of RSpec that break older Rails. (We are changing our support policy for Rails see, #1967)

> Added a build for development version of Ruby using Rails 5.2 (failures allowed).

Can you remove this, our build is too large as it is.

brian-kephart:
Please explain about the old Rails git branches. No older Rails were removed, only development branches for versions no longer in development, none of which have received a commit in years. Current master essentially tests these twice, once using RubyGems, and once using GitHub.

JonRowe:
No the branches are the stable release branches, we maintain support for rspec-rails with those branches currently, we are hoping to reduce this at our next major release but for now we are committed to maintaining that compatibility no matter when they were last committed to.

brian-kephart:
I still don't get it, but I guess that's your call.

I submitted a separate PR that's just changes to .travis.yml. I pushed the other requested changes to this branch.

brian-kephart:
I forgot that some of the other file changes assumed that those git release branches were no longer needed. This needs further reworking if you're intent on preserving those, and I don't have the time or enthusiasm for it.

This PR still fixes a lot of things that needed to be fixed to support more recent versions of Rails. I'll leave it here for others to reference or build on to implement that support in a way that you'll find more acceptable.

JonRowe:
Closed by #2001

PR #2000, opened 2018-06-19
Explicitly list Ruby/Rails combinations in .travis.yml

brian-kephart:
This PR separates out some of the changes originally included in #1998. Instead of applying inclusions and exclusions to an auto-generated matrix, now every Ruby/Rails combination to be tested is explicitly listed.

Other changes:
1. Patch version bumps
2. Specific versions changed to tilde version specifiers
3. Tests for Ruby < 2.4 not tested against Rails master, which now requires Ruby 2.4+

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,129 +34,170 @@ before_script:
 
 script: "script/run_build 2>&1"
 
-rvm:
-  - 1.8.7
-  - 2.4.4
-  - 2.3.7
-  - 2.2
-  - 2.1
-  - 2.0.0
-  - 1.9.3
-  - 1.9.2
-
-env:
-  - RAILS_VERSION='~> 4.2.0'
-  - RAILS_VERSION=4-2-stable
-  - RAILS_VERSION='~> 4.1.0'
-  - RAILS_VERSION=4-1-stable
-  - RAILS_VERSION='~> 4.0.4'
-  - RAILS_VERSION=4-0-stable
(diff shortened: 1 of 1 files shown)


JonRowe:
I appreciate the enthusiasm with the extra builds, but I was hoping for the build matrix to remain the same bar version upgrades! Note the deliberate `2.x` builds where because travis automatically picks the latest, that said I'm happy to replace them with FQ'd versions.

JonRowe:
I'd love to get a green build matrix here to, I believe it's just Rake causing problems on certain combinations again which can be resolved with the Gemfile.

brian-kephart:
Please note that there's a pre-existing issue here. The first time I cloned master and ran a Travis build on my account with no changes, the issue with rake appeared. If you look at the first two failures in the list, rake installs correctly from the Gemfile the first time, then when gems are reinstalled from the build script the wrong version is retrieved. This happened to me repeatedly, always on different jobs in the build (seemingly at random), and in each case the job completed successfully when I re-ran it.

Again, this happened BEFORE I made any changes. Attempting to resolve it via the Gemfile was not successful. I suspect the problem is actually [here](https://github.com/rspec/rspec-rails/blob/master/example_app_generator/generate_app.rb#L44), but a few attempts to fix it here weren't successful either.

I haven't seen the rake issue in the final build job before. No idea about that one.

JonRowe:
Closing in favour of #2001, thanks for your time and effort here!

PR #2007, opened 2018-07-01
Update supported Rails versions in Readme

hayesr:
Clear-up that `rspec-rails` works with Rails versions above 5.0


benoittgt:
Thanks @hayesr

PR #2010, opened 2018-07-13
Clarify that binstub can be created on Rails 4+

olivierlacan:
It's not just Rails 4.

I know this is a baby change but people could be confused into thinking it only works on Rails 4, not 5.


benoittgt:
Hello Olivier

Thanks for the patch. Why not mentioning Rails 5 too?

Thanks 😄

olivierlacan:
> Thanks for the patch. Why not mentioning Rails 5 too?

Pas bête. 😸

PR #2013, opened 2018-07-15
Change minimum Ruby version for Puma to 2.2 + specify rdoc version under 2.2

benoittgt:
Fix #2012

Puma now requires Ruby 2.2 minimum https://github.com/puma/puma/pull/1506/files

PR #2014, opened 2018-07-15
Avoid deprecated TestResponse#success? method

### LINKED ISSUE #2009: Scaffold controller spec template uses deprecated 'success?` method
When creating a resource with `rails generate scaffold`, we use this template for the rspec controller specs:
https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec/scaffold/templates/controller_spec.rb#L67

But the `success?` method on the response was deprecated:
https://github.com/rails/rails/blob/20c91119903f70eb19aed33fe78417789dbf070f/actionpack/lib/action_dispatch/testing/test_response.rb#L23

This may be as simple as changing the controller spec template to use `be_successful` instead of `be_success`. I can make a PR for this next week if nobody else does.

You can see the deprecation warning by running the specs here: https://github.com/composerinteralia/rspec-deprecated-method-app

Ruby version: 2.5.1
Rails version: 5.2.0
Rspec version: 3.7

composerinteralia:
Closes #2009

`TestResponse#success?` was deprecated with [this commit](https://github.com/rails/rails/commit/af3500b18817a48d7ec83812a0d4869f3fe33b2f). Instead, we can use `#successful?` from `Rack::Response::Helpers`. This method has been available for a long time. I tried this change with Rails 3.2, 4.2, and 5.2 and everything seemed to work fine.

### DIFF
--- a/lib/generators/rspec/scaffold/templates/api_controller_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/api_controller_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe <%= controller_class_name %>Controller, <%= type_metatag(:control
 <% else -%>
       get :index, params: {}, session: valid_session
 <% end -%>
-      expect(response).to be_success
+      expect(response).to be_successful
     end
   end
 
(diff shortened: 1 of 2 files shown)


JonRowe:
Great, we're just working on fixing our ci (#2013), and then I'd like this rebased off a working build to check it's all green and I'll merge it.

benoittgt:
CI is now green. Do you think you can rebase a re-push? I will look at this, this afternoon.

composerinteralia:
Hm, I rebased but I am now seeing several failures with `Gem::RuntimeRequirementNotMetError: rdoc requires Ruby version >= 2.2.2`

benoittgt:
Thanks a lot @composerinteralia

PR #2026, opened 2018-09-18
Fix `NoMethodError` for `Pathname` fixture paths

cupakromer:
When the `fixture_path` is defined as a `Pathname` object the `rails_fixture_file_wrapper` fails with:

    NoMethodError:
      undefined method `strip' for #<Pathname:0x00007fab0a970e70>

This resolves the issue by ensuring the fixture path is converted into a string prior to resolving it.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,8 @@ Enhancements:
 
 Bug Fixes:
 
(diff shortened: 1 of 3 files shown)


JonRowe:
Good catch!

Issue #2020, opened 2018-09-04
Strange behavior with have_enqueued_job


corentindms:
Hello everyone !
I've got to report a strange behavior that I observed today with the use of `have_enqueud_job`

Ruby version: ruby 2.4.1
Rails version: 4.2.10
Rspec version: 3.7

## Observed behaviour

Consider the following code : 
```ruby
class MyJob < ActiveJob::Base
  def perform(param)
    puts "Hello world ! #{param}"
  end
end

class JobLauncher
  def launch
    MyJob.perform_later("first job")
    MyJob.perform_later("second job")
  end
end

RSpec.describe JobLauncher do
  describe "#launch" do
    it "works" do
      expect {
        JobLauncher.new.launch
      }.to have_enqueued_job(MyJob)
    end
  end
end
```

This spec is failing because two jobs have been enqueued with the following message : 
```
expected to enqueue exactly 1 jobs, but enqueued 2
```

There is nothing wrong with that and that's the behavior expected.
But when you change the expectation line with : 
```ruby
      expect {
        JobLauncher.new.launch
      }.to have_enqueued_job(MyJob).with("first job")
```
then the spec does not fail anymore, even though there is still two jobs that have been enqueued.

## Expected behaviour

I find this difference of behavior very strange. I would have expected that `have_enqueued_job(Job).with(params)` would behave the same as without the `with`, just checking the job params extra.

Maybe it is not a bug but a feature. If so, is there a way to check that only one job have been enqueued with RSpec ?

Thank you

JonRowe:
The issue is we don't have a registry of enqueued jobs used / expected unlike how we do with expectations of methods etc, so we don't have global knowledge of whats expected, so "unexpected" jobs are just allowed through, this might take some work to fix.

As a work around this should work for now:

```
expect {
        JobLauncher.new.launch
}.to have_enqueued_job(MyJob) { |args| expect(args).to eq ['first job'] }
```

BobbyMcWho:
We could [make the default expected number relativity](https://github.com/rspec/rspec-rails/blob/master/lib/rspec/rails/matchers/active_job.rb#L19) `:at_least`- This breaks 3 specs, but only the expected error message. This changes the default behavior of the HaveEnqueuedJob matcher, but that can be restored by chaining the `#exactly` method on the matcher if someone is relying on the default exactly. 

I've opened PR https://github.com/rspec/rspec-rails/pull/2033 to continue the conversation

[...]

PR #2033, opened 2018-10-17
Add with_only method

BobbyMcWho:
In response to #2020, looking to spark discussion. 

Not sure if this solves the root of the question, but it seems useful to have a strict argument matcher in addition to a more lenient one. 

First time contributing to rspec project code, so feel free to let me know if there's anything I missed style wise, or in the contributing guide.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -16,12 +16,20 @@ module RSpec
             @queue = nil
             @at = nil
             @block = Proc.new {}
-            set_expected_number(:exactly, 1)
+            @strict_args = false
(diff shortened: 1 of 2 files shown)


JonRowe:
:wave: Hello, thanks for contributing here, `with` is actually supposed to be strict, but isn't due to it's implementation, however this implementation is too strict, it prevents other job classes from being enqueued which is not ideal. I'm not even sure this implementation would allow multiple job expectations...

I propose what needs to happen is the matcher needs to match jobs based on class, then check arguments when `with` is used. Some sort of registry may be required to allow multiple expectations to occur, similar to how we use our "space" object in rspec-mocks.

PR #2034, opened 2018-10-17
[Fixes #1877] Do not duplicate namespace in the path name

### LINKED ISSUE #1877: Generator Path
Hi friends at RSpec,

First of all, thank you so much for your hard work. I'm an avid fan and I just picked up the new book.

I wanted to bring to your attention something that been bothering me over the last few days. When I type

```
$ bin/rspec g rspec:feature folder/resource
```

The path of the generated file is

```
#{Rails.root}/spec/features/folder/folder_resource_spec.rb
```

Perhaps I'm not thinking clearly about this but I think I want

```
#{Rails.root}/spec/features/folder/resource_spec.rb
```

I would love to hear your feedback. If you find my proposal to be in agreeance, I would be happy to contribute a PR. If not, I totally understand.

jekuta:
If a feature spec is generated with a namespaced name, the namespace gets duplicate in the filename as well.

The `table_name` method returns a string that [is already a combination](Link) of the `class_path` and `file_name`. We could used `file_name` directly instead of the `table_name`, which is also the [approach taken in the controller generator](https://github.com/rspec/rspec-rails/blob/master/lib/generators/rspec/controller/controller_generator.rb#L17).

Closes #1877

### DIFF
--- a/lib/generators/rspec/feature/feature_generator.rb
+++ b/lib/generators/rspec/feature/feature_generator.rb
@@ -19,9 +19,9 @@ module Rspec
 
       def filename
         if options[:singularize]
-          "#{table_name.singularize}_spec.rb"
+          "#{file_name.singularize}_spec.rb"
         else
-          "#{table_name}_spec.rb"
+          "#{file_name}_spec.rb"
         end
       end
     end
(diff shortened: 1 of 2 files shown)


benoittgt:
Thanks a lot @jekuta for this PR. It looks ok for me.

I'm still a junior here so the merge button will be probably push by @JonRowe

JonRowe:
@benoittgt you can merge this if you feel its good to go, don't forget a change log :)

Issue #1876, opened 2017-09-21
Add a global switch for ActiveRecord


Jack12816:
Hey guys, first thank you for this great project! :)

I encountered lately [an issue like this](https://stackoverflow.com/questions/46204197/testing-a-databaseless-rails-5-application-with-rspec-rails), where the `ActiveRecord` constant was loaded,
 but on purpose not used on an app. The `rails_helper.rb` was configured
[not to use transactions, etc](https://github.com/rspec/rspec-rails/issues/1416#issuecomment-120055707). 

Unfortunately, the `FixtureSupport` module was configured for ActiveRecord
and errored like this:
```
   ActiveRecord::ConnectionNotEstablished:
     No connection pool with 'primary' found.     
```

I figured out that ActiveStorage had a dependency on ActiveRecord, which I do not use.
The app just makes use of the ActiveStorage service, without any database records.
So I had to unload the `ActiveRecord` constant on my `rails_helper.rb` like this:

```
# This file is copied to spec/ when you run 'rails generate rspec:install'
require 'spec_helper'
ENV['RAILS_ENV'] ||= 'test'
require File.expand_path('../../config/environment', __FILE__)

# Remove the ActiveRecord constant, because it is autloaded by
# ActiveStorage and not needed for our application. The presence
# of the ActiveRecord constant causes rspec-rails to include
# extra fixture support (see: https://goo.gl/6Ed17Y), which
# results in:
#
#   ActiveRecord::ConnectionNotEstablished:
#     No connection pool with 'primary' found.
#
Object.send(:remove_const, :ActiveRecord)   
```
 
It would be better to have a configuration switch on rspec-rails to turn off any
handling of ActiveRecord. The guessing based on the availability of the constant
results in false-positive cases like mine.

Something like that would be helpful (guessing by default, but configurable):
```
RSpec.configure do |config| 
  # Disable all ActiveRecord handling
  config.use_active_record = false
end
```

I'm also willing to send in a PR, if this is proposal is accepted. :)

JonRowe:
Hmm this feels like a bug in active storage, if it's not using Active Record it shouldn't load ActiveRecord, can you not only load the parts you need? I'm not overall against the idea of a flag for it but we'd need to keep the auto detection as well and that feels like extra maintenance burden.

Jack12816:
[...]

PR #2038, opened 2018-10-25
Optional ActiveRecord config

Jack12816:
**- What is it good for**

This PR ships a solution for #1876.

**- What I did**

I added a new option `use_active_record` to the RSpec config. This option is also added to the generator for new `spec/rails_helper.rb` files. The config option is read on the `FixtureSupport` module before the ActiveRecord loading happens. Tests with and without this setting were added in the case a database connection is not established.

**- A picture of a cute animal (not mandatory but encouraged)**

download

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -2,14 +2,15 @@ module RSpec
   module Rails
     # @private
     module FixtureSupport
-      if defined?(ActiveRecord::TestFixtures)
-        extend ActiveSupport::Concern
-        include RSpec::Rails::SetupAndTeardownAdapter
-        include RSpec::Rails::MinitestLifecycleAdapter
-        include RSpec::Rails::MinitestAssertionAdapter
-        include ActiveRecord::TestFixtures
+      extend ActiveSupport::Concern
 
-        included do
+      included do
+        if RSpec.configuration.use_active_record? && defined?(ActiveRecord::TestFixtures)
+          include RSpec::Rails::SetupAndTeardownAdapter
+          include RSpec::Rails::MinitestLifecycleAdapter
+          include RSpec::Rails::MinitestAssertionAdapter
+          include ActiveRecord::TestFixtures
+
           self.fixture_path = RSpec.configuration.fixture_path
           if ::Rails::VERSION::STRING > '5'
             self.use_transactional_tests = RSpec.configuration.use_transactional_fixtures
--- a/spec/rspec/rails/fixture_support_spec.rb
+++ b/spec/rspec/rails/fixture_support_spec.rb
@@ -2,7 +2,8 @@ module RSpec::Rails
   RSpec.describe FixtureSupport do
     context "with use_transactional_fixtures set to false" do
       it "still supports fixture_path" do
-        allow(RSpec.configuration).to receive(:use_transactional_fixtures) { false }
+        allow(RSpec.configuration).to \
+          receive(:use_transactional_fixtures) { false }
         group = RSpec::Core::ExampleGroup.describe do
           include FixtureSupport
         end
@@ -20,6 +21,58 @@ module RSpec::Rails
       end
 
       expect { group.new.setup_fixtures }.to_not raise_error
+    end
+
+    context "without database available" do
+      let(:example_group) do
+        RSpec::Core::ExampleGroup.describe("FixtureSupport") do
+          include FixtureSupport
+          include RSpec::Rails::MinitestLifecycleAdapter
+        end
+      end
+      let(:example) do
+        example_group.example("foo") do
+          expect(true).to be(true)
+        end
+      end
+
+      RSpec.shared_examples_for "unrelated example raise" do
+        it "raise due to no connection established" do
+          expect(example_group.run).to be(false)
+          expect(example.execution_result.exception).to \
+            be_a(ActiveRecord::ConnectionNotEstablished)
+        end
+      end
+
+      RSpec.shared_examples_for "unrelated example does not raise" do
+        it "does not raise" do
+          expect(example_group.run).to be(true)
+          expect(example.execution_result.exception).not_to \
+            be_a(ActiveRecord::ConnectionNotEstablished)
+        end
+      end
+
+      before { clear_active_record_connection }
+
+      after { establish_active_record_connection }
+
+      context "with use_active_record set to false" do
+        before { RSpec.configuration.use_active_record = false }
+
+        after { RSpec.configuration.use_active_record = true }
+
+        include_examples "unrelated example does not raise"
+      end
+
+      context "with use_active_record set to true" do
+        before { RSpec.configuration.use_active_record = true }
+
+        if Rails.version.to_f >= 4.0
+          include_examples "unrelated example does not raise"
+        else
+          include_examples "unrelated example raise"
+        end
+      end
     end
   end
 end
--- a/spec/support/ar_classes.rb
+++ b/spec/support/ar_classes.rb
@@ -1,7 +1,24 @@
-ActiveRecord::Base.establish_connection(
-  :adapter => 'sqlite3',
-  :database => ':memory:'
-)
+FileUtils.rm_f('./test.db')
+
+def establish_active_record_connection
+  ActiveRecord::Base.establish_connection(
+    :adapter => 'sqlite3',
+    :database => './test.db'
+  )
+end
+
+def clear_active_record_connection
+  ActiveRecord::Base.connection_handler.clear_all_connections!
+  ActiveRecord::Base.connection_handler.connection_pool_list.each do |pool|
+    if Rails.version.to_f >= 5.0
+      ActiveRecord::Base.connection_handler.remove_connection(pool.spec.name)
+    else
+      ActiveRecord::Base.connection_handler.remove_connection(ActiveRecord::Base)
+    end
+  end
+end
+
+establish_active_record_connection
 
 module Connections
   def self.extended(host)
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ bundle
 specs.out
 spec/examples.txt
 specs.out
+test.db
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -33,6 +33,10 @@ rescue ActiveRecord::PendingMigrationError => e
 end
 <% end -%>
 RSpec.configure do |config|
+  # You can explicitly turn off ActiveRecord support on rspec-rails by
+  # uncommenting the following line.
+  # config.use_active_record = false
+
 <% if RSpec::Rails::FeatureCheck.has_active_record? -%>
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
   config.fixture_path = "#{::Rails.root}/spec/fixtures"
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -64,6 +64,7 @@ module RSpec
       config.add_setting :infer_base_class_for_anonymous_controllers, default: true
 
       # fixture support
+      config.add_setting :use_active_record, default: true
       config.add_setting :use_transactional_fixtures, alias_with: :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -64,6 +64,9 @@ RSpec.describe "Configuration" do
                      :use_transactional_fixtures,
                      alias_with: :use_transactional_examples
 
+    include_examples "adds setting", :use_active_record,
+                     :default => true
+
     include_examples "adds setting", :use_instantiated_fixtures
 
     include_examples "adds setting", :global_fixtures


spec/rspec/rails/matchers/has_spec.rb
```diff
@@ -1,27 +1,31 @@
...
+      )
+    SQL
+
+    AssociatedItem.connection.execute <<-SQL
+      CREATE TABLE associated_items (
+        id integer PRIMARY KEY AUTOINCREMENT,
+        collection_owner_id integer
+      )
+    SQL
+  end
```
JonRowe: Why have you moved these?

lib/rspec/rails/fixture_support.rb
```diff
@@ -2,14 +2,15 @@ module RSpec
...
-        include RSpec::Rails::MinitestAssertionAdapter
-        include ActiveRecord::TestFixtures
+      extend ActiveSupport::Concern
+
+      included do
+        if RSpec.configuration.use_active_record? && defined?(ActiveRecord::TestFixtures)
+          include RSpec::Rails::SetupAndTeardownAdapter
+          include RSpec::Rails::MinitestLifecycleAdapter if ::ActiveRecord::VERSION::STRING > '4'
+          include RSpec::Rails::MinitestAssertionAdapter
+          include ActiveRecord::TestFixtures
```
JonRowe: Moving these includes is undesired

spec/support/ar_classes.rb
```diff
@@ -1,12 +1,35 @@
...
+  )
+
+  [:MockableModel,
+   :SubMockableModel,
+   :AssociatedModel,
+   :AlternatePrimaryKeyModel].each do |host|
+    next unless Kernel.const_defined? host
+    Connections.extended(Kernel.const_get(host))
+  end
+end
```
JonRowe: Why?

spec/support/ar_classes.rb
```diff
@@ -1,12 +1,35 @@
...
+
+establish_active_record_connection
 
 module Connections
   def self.extended(host)
     host.connection.execute <<-eosql
-      CREATE TABLE #{host.table_name} (
+      CREATE TABLE IF NOT EXISTS #{host.table_name} (
```
JonRowe: When would this exist?

lib/rspec/rails/fixture_support.rb
Jack12816: I think it's needed due to:

```
An error occurred while loading spec_helper.
Failure/Error: if RSpec.configuration.use_active_record? && defined?(ActiveRecord::TestFixtures)

NoMethodError:
  undefined method `use_active_record?' for #<RSpec::Core::Configuration:0x0000557710919768>
```

when the module looks like this:

```ruby
module FixtureSupport
  if RSpec.configuration.use_active_record? && defined?(ActiveRecord::TestFixtures)
    include RSpec::Rails::SetupAndTeardownAdapter
    include RSpec::Rails::MinitestLifecycleAdapter if ::ActiveRecord::VERSION::STRING > '4'
```

spec/rspec/rails/matchers/has_spec.rb
Jack12816: Because of the randomly running tests, the memory sqlite connection/spec is going to be removed. That's why the table is missing sometimes in the middle of the run. The SQL migration was inlined to the example group due to this fact.

lib/rspec/rails/fixture_support.rb
JonRowe: Ok so you had a load order issue, you need to make sure the config extension is loaded before this file, then have it like this snippet rather than moving the includes. Moving the includes has the effect of having them all in a different precedence order.

spec/support/ar_classes.rb
Jack12816: See the spec. 
```ruby
    context "without database available" do
      before { clear_active_record_connection }
       after { establish_active_record_connection }
       context "with use_active_record set to true" do
```
I wrote a spec which tests exactly this behavior.

`establish_active_record_connection` takes care of creating AR connection, but when it run for the first time the models are not yet defined and later in the middle of the suite the connection is dropped which leads to the fact that the db tables are missing and need to be migrated again.

Maybe switching to a sqlite file will be an better option, then the initial migrations stay in place after dropping the connections.

lib/rspec/rails/fixture_support.rb
Jack12816: Hmm, yeah but that's a chicken egg issue. When I load the config before the fixture_support when it yells at me with:

```
An error occurred while loading spec_helper.
Failure/Error: config.include RSpec::Rails::FixtureSupport, :use_fixtures

NameError:
  uninitialized constant RSpec::Rails::FixtureSupport
```

samphippen:
@Jack12816 any chance that you take a look at the travis build and get it to a passing state?

Thank you!

mrgordon:
This is a workaround for ActiveStorage bugs. Ideally it should be fixed there instead of adding new code to rspec-rails but I'd support this to save people time if we don't think they'll fix it

JonRowe:
@mrgordon If you'd like to use this as a base and fix it up I'm sure @Jack12816 won't mind, its gone rather stale. Please note the existing feedback still needs taking into account and a build passing.

spec/support/ar_classes.rb
benoittgt: It has been removed. Marked as *resolved*.

lib/rspec/rails/fixture_support.rb
benoittgt: I tried this

```diff
+++ b/lib/rspec/rails/fixture_support.rb
@@ -1,58 +1,57 @@
+require 'rspec/rails/configuration'
 module RSpec
   module Rails
     # @private
     module FixtureSupport
-      extend ActiveSupport::Concern
+      include RSpec::Rails::Configuration
+      if RSpec.configuration.use_active_record? && defined?(ActiveRecord::TestFixtures)
+        include RSpec::Rails::SetupAndTeardownAdapter
+        include RSpec::Rails::MinitestLifecycleAdapter if ::ActiveRecord::VERSION::STRING > '4'
+        include RSpec::Rails::MinitestAssertionAdapter
+        include ActiveRecord::TestFixtures
+        extend ActiveSupport::Concern
```

Without the require because I know I should not do that... 😉
```
An error occurred while loading spec_helper.
Failure/Error: include RSpec::Rails::Configuration

NameError:
  uninitialized constant RSpec::Rails::Configuration
# ./spec/spec_helper.rb:11:in `require'
# ./spec/spec_helper.rb:11:in `<top (required)>'
No examples found.
```

And I have the same error as @Jack12816. Sadly I don't see other way to fix this.

@JonRowe maybe you could tell me what do you mean by
> you need to make sure the config extension is loaded before this file

JonRowe: What about if we wrap:
```
config.include RSpec::Rails::FixtureSupport, :use_fixtures
```
In the config if block?

benoittgt: I tried

```diff
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -23,6 +23,7 @@ RSpec.configure do |config|
   config.filter_run :focus
   config.run_all_when_everything_filtered = true
   config.order = :random
+  config.include RSpec::Rails::FixtureSupport, :use_fixtures
```

And I get
```
An error occurred while loading spec_helper.
Failure/Error: config.include RSpec::Rails::FixtureSupport, :use_fixtures

NameError:
  uninitialized constant RSpec::Rails::FixtureSupport
# ./spec/spec_helper.rb:11:in `require'
# ./spec/spec_helper.rb:11:in `<top (required)>'
No examples found.


Finished in 0.00008 seconds (files took 1.57 seconds to load)
0 examples, 0 failures, 1 error occurred outside of examples
```

`# ./spec/spec_helper.rb:11:in `require'` is `require `rspec/rails'`

benoittgt:
@mrgordon you said
> This is a workaround for ActiveStorage bugs

Do you have more infos? Maybe it is fixed and we don't need this patch anymore? Especially with rspec-rails 4

mrgordon:
I have no other info. Just seems logical that ActiveStorage should either not require ActiveRecord at all or should automatically require it if necessary so it functions. Instead there is an unstated (?) / unclear dependency that is only noticed if you remove ActiveRecord because things break.

benoittgt:
@mrgordon thanks

@Jack12816 Do you think you can rebase.

Jack12816:
@benoittgt I've rebased the branch and did some changes to make it work with Rails 3.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,54 @@ module RSpec::Rails
...
+        # +ActiveRecord::ConnectionNotEstablished+ exception at this point.
+        # (Eg. Rails 3.x, 4.0, 5.0)
+        it "raise due to no connection established" do
+          example_group = RSpec::Core::ExampleGroup.describe("FixtureSupport") do
+            include FixtureSupport
+            include RSpec::Rails::MinitestLifecycleAdapter
+          end
+
+          test = example_group.example("foo") do
+            expect(true).to be(true)
```
benoittgt: Why are we doing this?

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,54 @@ module RSpec::Rails
...
+          end
+
+          success = example_group.run
+          expect(test.execution_result.exception).to \
+            be_a(ActiveRecord::ConnectionNotEstablished) unless success
```
benoittgt: I found this test quite hard to read. Does the `unless success` refers to your comment from line 25 to 27?

If yes, maybe wrap the entire test to the specific Rails version? Like https://github.com/rspec/rspec-rails/blob/85e23f36e1595a3ee37c4eb8271fe459e2030716/spec/rspec/rails/configuration_spec.rb#L259

Also I see more often in RSpec code base
```ruby
expect(
  test.execution_result.exception
).to be_a(ActiveRecord::ConnectionNotEstablished)
```
or
```ruby
 expect(test.execution_result.exception)
   .to be_a(ActiveRecord::ConnectionNotEstablished)
```

instead of 
```ruby
 expect(test.execution_result.exception).to \
            be_a(ActiveRecord::ConnectionNotEstablished)
```

pirj:
@Jack12816 are you up to wrap this up?
Just in case, Rails 3 support has been dropped.

pirj:
@Jack12816 a friendly reminder.

Jack12816:
@pirj I've rebased it to upstream master and now all tests passing.

spec/support/ar_classes.rb
JonRowe: This should only be done in the specs necessary and not globally.

spec/support/ar_classes.rb
```diff
@@ -1,7 +1,24 @@
...
+  ActiveRecord::Base.connection_handler.connection_pool_list.each do |pool|
+    if Rails.version.to_f >= 5.0
+      ActiveRecord::Base.connection_handler.remove_connection(pool.spec.name)
+    else
+      ActiveRecord::Base.connection_handler.remove_connection(ActiveRecord::Base)
+    end
+  end
+end
+
+establish_active_record_connection
```
JonRowe: This should be removed, we don't want every other test having a database connection.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -21,5 +22,57 @@ module RSpec::Rails
...
+      before { clear_active_record_connection }
+
+      after { establish_active_record_connection }
+
+      context "with use_active_record set to false" do
+        before { RSpec.configuration.use_active_record = false }
+
+        after { RSpec.configuration.use_active_record = true }
+
+        include_examples "unrelated example does not raise"
```
JonRowe: This spec does nothing to demonstrate that the support works when active record is available but not configured, you should be able to demonstrate the issue without adding database support as its supposed to be broken without database support.

spec/support/ar_classes.rb
Jack12816: Good point, but it's not a subject of the PR. Actually all tests require a database connection due to the `*Model` classes at `ar_classes.rb`. When disabling AR support on the specs overall, everything is broken. Disabling the database connection on all tests sounds like a bigger refactoring.

JonRowe: > Actually all tests require a database connection due to the *Model classes at ar_classes.rb. When disabling AR support on the specs overall, everything is broken. 

Currently there is no database configured so I'm not sure thats true, if your changes cause things to fail your changes are at fault, applying a global config to the specs is not a valid fix.

spec/rspec/rails/fixture_support_spec.rb
Jack12816: Unfortunately, I don't get your point. When `use_active_record` is set to `true` and no AR connection is available, it raises an `ActiveRecord::ConnectionNotEstablished`. When `use_active_record` is set to `false` and no AR connection is available it does not raise. Could you explain in detail what you have in mind for another test?

JonRowe: Its artificial because you have configured a database then manually cleared it.

The default is no configuration, the changes adding config should be removed and these spec should demonstrate that `requiring' active_record'` causes an error, and that setting `use_active_record` false prevents it.

JonRowe: Sorry, I've been trying to find time to work on this myself because I believe strongly that adding database configuration for all the specs is not the way to go! :)

spec/support/ar_classes.rb
Jack12816: > Currently there is no database configured so I'm not sure thats true

https://github.com/rspec/rspec-rails/blob/master/spec/support/ar_classes.rb#L1-L19

JonRowe: Hmm ok, well make it fail with that.

JonRowe: We have `in_sub_process` helpers if you need to change something, but I don't want global configuration changes for just this spec, it shouldn't be necessary.

JonRowe:
Thanks for your hard work getting this started, closing in favour of #2266, would you be able to check that works for you? Once green I can ship as a beta if you'd like, this is one of the last things we'd like included in `4.0.0` final.

PR #2039, opened 2018-10-29
Fix typo in changelog

mschnitzer:
There seems to be a typo in the changelog for version 3.8.1


JonRowe:
Yes that was my mistake, thanks for catching that!

PR #2047, opened 2018-12-11
Feature/have enqueued mail matcher

### LINKED ISSUE #1901: Feature request: have_enqueued_email matcher to deal with ActionMailer::DeliveryJob
I use a lot the `have_enqueued_job` matcher because I like its syntax and it also perform params serialization that allows me to catch invalid arguments given to background jobs such as symbol serialization.

However I think that it's a little bit too verbose when used with `ActionMailer::DeliveryJob` e.g.

```ruby
expect {
  subject.perform request
}.to have_enqueued_job(ActionMailer::DeliveryJob).with(
    'SubscriptionMailer',
    'welcome',
    'deliver_now',
    hash_including('email_address' => email),
    'en'
)
```

It would be very nice to have a specialized wrapper around it, something like this

```ruby
expect {
  subject.perform request
}.to have_enqueued_email(SubscriptionMailer, :welcome).with(hash_including('email_address' => email), 'en')
```

What do you think?

jdlubrano:
Work in progress to add a `have_enqueued_mail` RSpec matcher

Fixes #1901

### DIFF
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -28,10 +28,7 @@ module RSpec
 
         def with(*args, &block)
           @mail_args = args
-
-          super(*mailer_args) do |*job_args|
-            block.call(*(job_args - base_mailer_args)) if block.present?
-          end
+          block.nil? ? super(*mailer_args) : super(*mailer_args, &yield_mail_args(block))
         end
 
         def matches?(block)
@@ -84,6 +81,10 @@ module RSpec
 
         def base_mailer_args
           [@mailer_class.name, @method_name.to_s, MAILER_JOB_METHOD]
+        end
+
+        def yield_mail_args(block)
+          Proc.new { |*job_args| block.call(*(job_args - base_mailer_args)) }
         end
 
         def check_active_job_adapter
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -14,6 +14,8 @@ module RSpec
           @mailer_class = mailer_class
           @method_name = method_name
           @args = []
+          @at = nil
+          @queue = nil
           @job_matcher = ActiveJob::HaveEnqueuedJob.new(ActionMailer::DeliveryJob)
           set_expected_count(:exactly, 1)
         end
@@ -32,6 +34,18 @@ module RSpec
 
         def with(*args)
           @args = args
+          self
+        end
+
+        def at(send_time)
+          @at = send_time
+          @job_matcher.at(send_time)
+          self
+        end
+
+        def on_queue(queue)
+          @queue = queue
+          @job_matcher.on_queue(queue)
           self
         end
 
@@ -73,8 +87,10 @@ module RSpec
         def base_message
           "#{@mailer_class.name}.#{@method_name}".tap do |msg|
             msg << " #{expected_count_message}"
-            msg << " with #{@args}" if @args.any?
-            msg << ", but enqueued #{@job_matcher.matching_jobs.size}"
+            msg << " with #{@args}," if @args.any?
+            msg << " on queue #{@queue}," if @queue
+            msg << " at #{@at.inspect}," if @at
+            msg << " but enqueued #{@job_matcher.matching_jobs.size}"
           end
         end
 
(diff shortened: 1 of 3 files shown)


spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    ActiveJob::Base.queue_adapter = :test
+  end
+
+  describe "have_enqueued_mail" do
+    it "passes when a mailer method is called with deliver_later" do
+      expect {
+        TestMailer.test_email.deliver_later
+      }.to have_enqueued_mail(TestMailer, :test_email)
+    end
+  end
```
JonRowe: Can we get a spec for checking `with` too?

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -0,0 +1,87 @@
...
+
+module RSpec
+  module Rails
+    module Matchers
+      # @api private
+      #
+      # Matcher class for `have_enqueued_mail`. Should not be instantiated directly.
+      #
+      # @see RSpec::Rails::Matchers#have_enqueued_mail
+      class HaveEnqueuedMail < RSpec::Matchers::BuiltIn::BaseMatcher
```
JonRowe: I think yard wants an `@private` here, or at least its complaining about undocumented methods.

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
jdlubrano: Most definitely.  I left this PR in what I consider to be a very early stage.  My day job got in the way a bit this week.

jdlubrano:
@JonRowe After I started poking around the ActiveJob matcher I realized that a lot could be added to the have_enqueued_mail matcher which is admittedly barebones.  To answer your question, yes I do plan to mimic as much as I can with regards to the ActiveJob functionality.

jdlubrano:
@JonRowe 

How does this look to you now?  The only piece of functionality, I think, that I didn't implement for the `have_enqueued_mail` matcher was the ability to pass a block to `.with`.  Is that missing functionality a necessity?  I was kind of scratching my head wondering what the use case for passing a block to `.with` for `ActiveJobs` was in the first place.  I understand how it works from the documentation and the code.

Nevertheless, it didn't seem as easy to pass along only the mailer arguments to that `with` block given the implementation of the `RSpec::Rails::Matchers::HaveEnqueuedJob#check` method.  Before I potentially extracted something in the `ActiveJob` matchers, I wanted to get a better understanding of a real(ish) use case for passing a block `.with` in order to consider how such a case could apply to a `ActionMailer`.

JonRowe:
Very good! The `with` block allows you to express more complex argument matching than is possible directly, e.g.

```
expect { ... }.to have_enqueued_mail.with { |a| expect(a).to eq("a").or eq("b") }
```

I think its desirable, but its not a blocker. Let me know what you think.

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -0,0 +1,166 @@
...
+          msg_parts << "with #{mailer_args}" if mailer_args.any?
+          msg_parts << "on queue #{job[:queue]}" if job[:queue] && job[:queue] != 'mailers'
+          msg_parts << "at #{Time.at(job[:at])}" if job[:at]
+
+          "#{mailer_method} #{msg_parts.join(', ')}".strip
+        end
+
+        def mailer_job
+          ActionMailer::DeliveryJob
+        end
```
JonRowe: Not a blocker, but why not just use the constant name directly in places?

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -0,0 +1,246 @@
...
+    end
+
+    it "generates a failure message with unmatching enqueued mail jobs" do
+      send_time = Date.tomorrow.noon
+      queue = 'urgent_mail'
+
+      message = "expected to enqueue TestMailer.email_with_args exactly 1 time with [1, 2], but enqueued 0" + \
+                "\nQueued deliveries:" + \
+                "\n  TestMailer.test_email" + \
+                "\n  TestMailer.email_with_args with [3, 4], on queue #{queue}, at #{send_time}"
```
JonRowe: Not a blocker,  but it might be nice to show other jobs are not mentioned here.

jdlubrano: Okay, that's a good callout; I will add that!

lib/rspec/rails/matchers/have_enqueued_mail.rb
jdlubrano: I had the hardest time keeping the constant straight in my head.  I kept wanting to call it `ActionDelivery::MailerJob` or `ActionMailer::DeliverJob`.  So I guess out of frustration, I just made it a method 😅

jdlubrano:
@JonRowe I may have a way to replicate the `with` block capability that would not require too much overhead.  I will try that out and if it proves to be straightforward, I will include it.  I can probably get to it this weekend.

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
JonRowe: 👍

benoittgt:
Thanks a lot @jdlubrano

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -24,9 +24,12 @@ def description
           "enqueues #{@mailer_class.name}.#{@method_name}"
         end
 
-        def with(*args)
+        def with(*args, &block)
           @mail_args = args
-          super(*mailer_args)
+
+          super(*mailer_args) do |*job_args|
+            block.call(*(job_args - base_mailer_args)) if block.present?
```
JonRowe: This should just call super if block is nil, else pass this block

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -75,10 +76,14 @@ def mailer_args
                                mailer_method_arity
                              end
 
-            base_args + Array.new(number_of_args) { anything }
+            base_mailer_args + Array.new(number_of_args) { anything }
           end
         end
 
+        def base_mailer_args
+          [@mailer_class.name, @method_name.to_s, 'deliver_now']
```
JonRowe: The string should be frozen, and could be a constant within this class

jdlubrano:
@JonRowe 

I believe that I addressed the latest changes that you requested.  It looks like the builds are failing on Travis CI due to bundler issues, however.  I'm fairly certain those failures are not specific to this branch...

JonRowe:
Awaiting #2061 probably, will ask you to give it a poke when thats merged!

JonRowe:
#2061 is merged so a rebase would get this green hopefully.

jdlubrano:
@JonRowe 

Looks like the build passed!

PR #2057, opened 2018-12-26
Fix generate request spec name has `/`

koheisg:
## before

```
$ rails g rspec:request api/posts
Running via Spring preloader in process 2817
      create  spec/requests/api/api_posts_spec.rb
```

```
$ cat spec/requests/api/api_posts_spec.rb
require 'rails_helper'

RSpec.describe "Api::Posts", type: :request do
  describe "GET /api_posts" do
    it "works! (now write some real specs)" do
      get api_posts_index_path
      expect(response).to have_http_status(200)
    end
  end
end
```

## after

```
$ rails g rspec:request api/posts
Running via Spring preloader in process 36682
      create  spec/requests/api/posts_spec.rb
```

```
$ cat spec/requests/api/posts_spec.rb
require 'rails_helper'

RSpec.describe "Api::Posts", type: :request do
  describe "GET /api/posts" do
    it "works! (now write some real specs)" do
      get api_posts_index_path
      expect(response).to have_http_status(200)
    end
  end
end
```

### DIFF
--- a/lib/generators/rspec/integration/integration_generator.rb
+++ b/lib/generators/rspec/integration/integration_generator.rb
@@ -15,7 +15,7 @@ module Rspec
         return unless options[:request_specs]
 
         template 'request_spec.rb',
-                 File.join('spec/requests', class_path, "#{table_name}_spec.rb")
+                 File.join('spec/requests', "#{name.underscore.pluralize}_spec.rb")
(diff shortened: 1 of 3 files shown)


JonRowe:
Thank you! ❤️

PR #1999, opened 2018-06-16
Truncate method names in system tests

samphippen:
fixes #1976

@JonRowe @benoittgt @myronmarston

PR #2058, opened 2018-12-30
Truncate more method names in system tests

tkm-kj:
ref: https://github.com/rspec/rspec-rails/pull/1999/files

If system test fails, a screenshot is taken, but if the file name exceeds 255 characters(e.g. using turnip), the following error will be output.

```
File name too long @ rb_sysopen
```

The reason is the test fails, `failures_` is appended to first of file name.
So I truncated it to a certain length so that the length of the file name does not become long.

### DIFF
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -37,7 +37,7 @@ module RSpec
         @method_name ||= [
           self.class.name.underscore,
           RSpec.current_example.description.underscore
-        ].join("_").tr(CHARS_TO_TRANSLATE.join, "_")[0...251] + "_#{rand(1000)}"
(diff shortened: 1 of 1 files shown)


lib/rspec/rails/example/system_example_group.rb
```diff
@@ -37,7 +37,7 @@ def method_name
         @method_name ||= [
           self.class.name.underscore,
           RSpec.current_example.description.underscore
-        ].join("_").tr(CHARS_TO_TRANSLATE.join, "_")[0...251] + "_#{rand(1000)}"
+        ].join("_").tr(CHARS_TO_TRANSLATE.join, "_")[0...100] + "_#{rand(1000)}"
```
JonRowe: Can we have 200? It seems 50 chars would be enough for more added words, but 100 is going to be close to a normal spec files docstring length.

PR #2061, opened 2019-01-03
Updates from rspec-dev (2019-01-03)

JonRowe:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,7 @@ cache: bundler
 bundler_args: "--binstubs --path ../bundle --retry=3 --jobs=3"
 
 before_install:
-  - gem update --system 2.7.8
-  - gem install bundler
+  - script/update_rubygems_and_install_bundler
   - script/clone_all_rspec_repos
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ bundler_args: "--binstubs --path ../bundle --retry=3 --jobs=3"
 
 before_install:
   - script/update_rubygems_and_install_bundler
+  - script/downgrade_bundler_on_old_rails
   - script/clone_all_rspec_repos
 
 before_script:
--- a/script/downgrade_bundler_on_old_rails
+++ b/script/downgrade_bundler_on_old_rails
@@ -1,0 +1,19 @@
+#!/bin/bash
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
+# DO NOT modify it by hand as your changes will get lost the next time it is generated.
+
+set -e
+source script/functions.sh
+
+if ruby -e "exit(ENV['RAILS_VERSION'].to_f < 5)"; then
+  # On Rails versions less than 5, Bundler 2.0 is not supported
+  echo "Warning dowgrading to older version of Bundler"
+  gem uninstall -aIx bundler
+
+  # this only works on Ruby 2.3 which is luckily the version we need to fix
+  if is_ruby_23_plus; then
+    rvm @global do gem uninstall -aIx bundler
+  fi
+
+  gem install bundler -v '1.17.3'
+fi
--- a/script/update_rubygems_and_install_bundler
+++ b/script/update_rubygems_and_install_bundler
@@ -1,0 +1,15 @@
+#!/bin/bash
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
+# DO NOT modify it by hand as your changes will get lost the next time it is generated.
+
+set -e
+source script/functions.sh
+
+if is_ruby_23_plus; then
+  gem update --system
+  gem install bundler
+else
+  echo "Warning installing older versions of Rubygems / Bundler"
+  gem update --system '2.7.8'
+  gem install bundler -v '1.17.3'
+fi
--- a/.rubocop_rspec_base.yml
+++ b/.rubocop_rspec_base.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2018-12-30T16:45:57+00:00 from the rspec-dev repo.
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 # This file contains defaults for RSpec projects. Individual projects
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2018-12-30T16:45:57+00:00 from the rspec-dev repo.
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 version: "{build}"
--- a/script/clone_all_rspec_repos
+++ b/script/clone_all_rspec_repos
@@ -1,5 +1,5 @@
 #!/bin/bash
-# This file was generated on 2018-12-30T16:45:57+00:00 from the rspec-dev repo.
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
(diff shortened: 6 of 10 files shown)


script/update_rubygems_and_install_bundler
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
```
benoittgt: I don't see it on rspec-dev at the moment https://github.com/rspec/rspec-dev/tree/master/travis/script do we want to also do this change on other RSpec repos?

Also I see those functions but not where it is called? 🤔

script/update_rubygems_and_install_bundler
```diff
@@ -0,0 +1,15 @@
...
+# DO NOT modify it by hand as your changes will get lost the next time it is generated.
+
+set -e
+source script/functions.sh
+
+if is_ruby_23_plus; then
```
benoittgt: Also I see those functions but not where it is called? 🤔

script/update_rubygems_and_install_bundler
JonRowe: It's a WIP until I see it's working :p (also this is the worst repo to pick to review as its the most out of sync!)

script/update_rubygems_and_install_bundler
JonRowe: As this is rspec-rails it's `.travis.yml` is kept out of step, and I hadn't manually swapped it out, have now.

.travis.yml
```diff
@@ -23,8 +23,7 @@ cache: bundler
 bundler_args: "--binstubs --path ../bundle --retry=3 --jobs=3"
 
 before_install:
-  - gem update --system 2.7.8
-  - gem install bundler
+  - script/update_rubygems_and_install_bundler
```
benoittgt: 🎉

script/update_rubygems_and_install_bundler
benoittgt: Thanks Jon

benoittgt:
`ERROR:  Could not find a valid gem '1.17.3' (>= 0) in any repository`

But https://rubygems.org/gems/bundler/versions/1.17.3

JonRowe:
Yes, again, this wasn't the WIP repo, so it got updated last.

JonRowe:
Once I had rspec/rspec-support#362 working the changes got backported

benoittgt:
It seems the CI install multiple version of Bundler. I saw that your task install bundler then try to uninstall bundler if Rails version is under 5 then install bundler 1.17.3.

Reading : https://travis-ci.org/rspec/rspec-rails/jobs/475443754
- `$ export RAILS_VERSION='~> 4.2.0'`
- `rvm use 2.4.4 --install --binary --fuzzy` so we should use bundler `1.17.3`
- From the cache
   - `$ bundle --version`
      - Bundler version 1.16.2 1️⃣
  - `$ gem --version` 
      - 2.7.7
- `script/update_rubygems_and_install_bundler`
    - Successfully installed rubygems-update-3.0.2
    - Bundler 1.17.3 installed 2️⃣
    - RubyGems system software updated
    - Fetching bundler-2.0.1.gem
    - Successfully installed bundler-2.0.1 3️⃣
- `script/downgrade_bundler_on_old_rails` output nothing 🤔 

I count 3 versions before running RSpec test suite.


Why did you prefer to add a specific uninstall script rather than installing bundler at the right version first in `update_rubygems_and_install_bundler` script? Do you want to stay as sync as possible from rspec-dev?

script/downgrade_bundler_on_old_rails
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# This file was generated on 2019-01-03T20:34:23+00:00 from the rspec-dev repo.
+# DO NOT modify it by hand as your changes will get lost the next time it is generated.
+
+set -e
+source script/functions.sh
+
+if is_ruby_23_plus; then
+  if ruby -e "exit(!(ENV['RAILS_VERSION'].to_f < 5))"; then
```
benoittgt: I think there is a mistake. We don't need the bang.

With:
```sh
set -e

ruby -v
echo $RAILS_VERSION

function is_ruby_23_plus {
  if ruby -e "exit(RUBY_VERSION.to_f >= 2.3)"; then
    return 0
  else
    return 1
  fi
}

if is_ruby_23_plus; then
  if ruby -e "exit(ENV['RAILS_VERSION'].to_f < 5)"; then
    # On Rails versions less than 5, Bundler 2.0 is not supported
    echo "Warning dowgrading to older version of Bundler"
    gem uninstall bundler
    gem install bundler -v '1.17.3'
  else
    echo "Rails >= 5"
  fi
fi
```
output
```
ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin18]
~> 4.2.0
Warning dowgrading to older version of Bundler
```

script/downgrade_bundler_on_old_rails
```diff
@@ -0,0 +1,15 @@
...
+    # On Rails versions less than 5, Bundler 2.0 is not supported
+    echo "Warning dowgrading to older version of Bundler"
```
benoittgt: Oups *"downgrading"* 😉

script/downgrade_bundler_on_old_rails
JonRowe: Yes a typo from a half way though

script/downgrade_bundler_on_old_rails
JonRowe: Fixed

JonRowe:
@benoittgt I want to leave the current rspec-dev as is given rspec-rails is different, this is the easiest fix for older Rails, when rspec-rails 4 is release I expect this will go away quite quickly, as it drops older Rails in line with Rails

benoittgt:
The CI is failing on few versions. It seems we are getting bundler from the image/cache. https://travis-ci.org/rspec/rspec-rails/jobs/476497133

Maybe in downgrading bundler version we should uninstall all versions then install 1.7.3

benoittgt:
I just saw that you do it.. `gem uninstall -aIx bundler`

benoittgt:
🤔 
screenshot 2019-01-08 at 17 42 43

Maybe we should add logs to see `which` bundler is used?
Since rubygems 3.0.0 we can also do 

```sh
$ gem info -a bundler                                                                                                      [17:51:45]

*** LOCAL GEMS ***

bundler (1.17.3, 1.17.2)
    Authors: André Arko, Samuel Giddins, Colby Swandale, Hiroshi
    Shibata, David Rodríguez, Grey Baker, Stephanie Morillo, Chris
    Morris, James Wen, Tim Moore, André Medeiros, Jessica Lynn Suttles,
    Terence Lee, Carl Lerche, Yehuda Katz
    Homepage: http://bundler.io
    License: MIT
    Installed at (1.17.3, default): /Users/benoit.tigeot/.rvm/rubies/ruby-2.5.1/lib/ruby/gems/2.5.0
                 (1.17.2): /Users/benoit.tigeot/.rvm/gems/ruby-2.5.1

    The best way to manage your application's dependencies
```

benoittgt:
Thanks for the patch. I am wondering if it is related to:
> The version of Bundler in the BUNDLED WITH section is read by Bundler to determine which version of Bundler should run. Using the example lock above, Bundler 1 will be used, as you can see here: ([source](https://bundler.io/v2.0/guides/bundler_2_upgrade.html))

But we don't have Gemfile.lock..

Other scenario could be to declare the bundler version we want to use with `bundle _1.17.3_ install` but I hope your last patch will work.

JonRowe:
I believe it's because Travis uses rvm and it has bundler 2 on Ruby 2.3 by default as a globally installed gem, picking the bundler version manually should also work but I'd rather actually uninstall it if I can

benoittgt:
Ok it seems it is not working. Maybe we can try this ugly hack 😞 
https://github.com/travis-ci/travis-ci/issues/8717#issuecomment-366500378

PR #2069, opened 2019-01-15
Change negated active job matcher, fixes #1870

### LINKED ISSUE #1870: Negative `have_enqueued_job` does not work as expected with multiple jobs of same kind

Istanful:
# Change negated active job matcher, fixes #1870 
When using the active job matchers, they will default to expecting a job to be enqueued exactly once. This results in some non intuitional behaviour, when using the negated matcher.

Consider the following example:
```ruby
expect {
  heavy_lifting_job.perform_later
  heavy_lifting_job.perform_later
}.not_to have_enqueued_job
```

Some would read this as "When enqueuing two heavy lifting jobs I expect no jobs to have been enqueued". Since two jobs are actually enqueued it should fail, which it previously would not.

This pull request fixes this issue by setting the default expectation to use "at least once" when using the negated matcher.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -197,6 +197,12 @@ module RSpec
 
             check(in_block_jobs)
           end
+
+          def does_not_match?(proc)
+            set_expected_number(:at_least, 1)
+
+            !matches?(proc)
+          end
         end
 
         # @private
@@ -204,6 +210,12 @@ module RSpec
           def matches?(job)
             @job = job
             check(queue_adapter.enqueued_jobs)
+          end
+
+          def does_not_match?(proc)
+            set_expected_number(:at_least, 1)
+
+            !matches?(proc)
           end
         end
       end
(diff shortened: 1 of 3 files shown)


JonRowe:
Thanks!

PR #2072, opened 2019-01-19
Gemfile: Update Nokogiri, FFI deps

olleolleolle:
This PR onto 4.0-dev made it possible for me to install on Ruby 2.6.0.

(Are there any special reasons to avoid latest, for these two?)

- [Nokogiri Changelog](https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md#1101--2019-01-13)
- [FFI Changelog](https://github.com/ffi/ffi/blob/master/CHANGELOG.md#1100--2019-01-06)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,7 +25,7 @@ if RUBY_VERSION >= '2.4.0'
   gem 'json', '>= 2.0.2'
 end
 
-gem 'ffi', '~> 1.9.25'
+gem 'ffi', '~> 1.10.0'
 
 gem 'rake', '>= 10.0.0'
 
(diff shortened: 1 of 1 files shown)


samphippen:
Hi @olleolleolle,

we tend to keep things at the *lowest possible* compatible version of a gem. The reason for this is that we have reason to believe we're not developing against things that are too aggressively new. I'll add 2.6 to the travis.yml and see what we can do for you here.

olleolleolle:
Thanks for taking 2.6.0 into the matrix.

I'll be closing this issue.

JonRowe:
@olleolleolle These dependencies are typically in place to allow older versions of Ruby to install them, see the "work arounds" in our other Gemfiles. The Gemfile is typically work arounds for Travis versions etc, as the gemspec contains the real dependencies.

PR #2073, opened 2019-01-24
Remove Ruby 2.4.4 from the Rails master matrix

pixeltrix:
Rails 6.0 will require Ruby >= 2.5.0 - see rails/rails#34754.


samphippen:
yes.

PR #2075, opened 2019-01-30
generator system spec: closing 1933

benoittgt:
This PR is meant to close the https://github.com/rspec/rspec-rails/pull/1933

CI will probably fail this is a WIP PR but open for review.

For exemple why do I get:
`expected RuntimeError with "System Tests are available for Rails >= 5.1", got #<RuntimeError: System Tests are available for Rails >= 5.1> with backtrace:`

### DIFF
--- a/lib/generators/rspec/system/system_generator.rb
+++ b/lib/generators/rspec/system/system_generator.rb
@@ -1,24 +1,25 @@
 require 'generators/rspec'
 
-module Rspec
-  module Generators
-    # @private
-    class SystemGenerator < Base
-      class_option :system_specs, :type => :boolean, :default => true,  :desc => "Generate system specs"
+if ::Rails::VERSION::STRING >= '5.1'
+  module Rspec
+    module Generators
+      # @private
+      class SystemGenerator < Base
+        class_option :system_specs, :type => :boolean, :default => true,  :desc => "Generate system specs"
 
-      def generate_system_spec
-        return unless options[:system_specs]
-        raise "System Tests are available for Rails >= 5.1" if ::Rails::VERSION::STRING < '5.1'
+        def generate_system_spec
+          return unless options[:system_specs]
 
-        template template_name, File.join('spec/system', class_path, filename)
-      end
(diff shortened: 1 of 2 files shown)


lib/generators/rspec/system/system_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+
+module Rspec
+  module Generators
+    # @private
+    class SystemGenerator < Base
+      class_option :system_specs, :type => :boolean, :default => true,  :desc => "Generate system specs"
+
+      def generate_system_spec
+        return unless options[:system_specs]
+        raise "System Tests are available for Rails >= 5.1" if ::Rails::VERSION::STRING < '5.1'
```
benoittgt: Is it what we want? 🤔 

I look at other parts of RSpec code and didn't see that kind of behavior implemented.

spec/generators/rspec/system/system_generator_spec.rb
```diff
@@ -0,0 +1,47 @@
+# Generators are not automatically loaded by rails
+require 'generators/rspec/system/system_generator'
+require 'support/generators'
+
+RSpec.describe Rspec::Generators::SystemGenerator, :type => :generator do
+  setup_default_destination
+
+  describe "system specs" do
+    subject(:system_spec) { file("spec/system/posts_spec.rb") }
+    if ::Rails::VERSION::STRING < '5.1'
```
benoittgt: Do we want this as a gate or do we want to tag the example to don't be run under Rails 5.1?

lib/generators/rspec/system/system_generator.rb
JonRowe: The generator shouldn't be available on old rails...

benoittgt: So we should have a higher condition if not on > Rails 5.1?

JonRowe: I don't think we should even define it we can't run it.

benoittgt:
Thanks @JonRowe for the review

PR #2081, opened 2019-02-22
Implement feedbacks from #2071 for RSpec 4

benoittgt:


Gemfile
```diff
@@ -11,8 +11,9 @@ gem 'yard', '~> 0.8.7', :require => false
 
 ### deps for rdoc.info
 group :documentation do
-  gem 'redcarpet',     '2.3.0'
-  gem 'github-markup', '1.0.0'
+  gem 'redcarpet', '~> 3.4'
```
samphippen: please use exact pins. We want to pin to some low version, which is the lowest version that we support, such that as new versions get released, we know RSpec still works with old versions of those gems.

I'd say `gem 'redcarpet', '3.4.x'` where x is the highest currently released version in the series, and the same for all other gems in this file.

example_app_generator/generate_app.rb
```diff
@@ -23,6 +23,7 @@
   gsub_file "Gemfile", /.*debugger.*/, ''
   gsub_file "Gemfile", /.*byebug.*/, "gem 'byebug', '~> 9.0.6'"
   gsub_file "Gemfile", /.*puma.*/, ""
+  gsub_file "Gemfile", /.*sqlite3.*/, "gem 'sqlite3', '~> 1.3.6'"
```
samphippen: 👍

lib/generators/rspec/model/model_generator.rb
```diff
@@ -23,8 +23,7 @@ def create_model_spec
 
       def create_fixture_file
         return unless missing_fixture_replacement?
-
-        template 'fixtures.yml', File.join('spec/fixtures', "#{table_name}.yml")
+        template 'fixtures.yml', File.join('spec/fixtures', class_path, "#{(pluralize_table_names? ? plural_file_name : file_name)}.yml")
```
samphippen: 👍

samphippen:
@benoittgt small feedback about how the Gemfile is structured, otherwise this looks great, and thank you for your quick response!

PR #2088, opened 2019-02-27
AV::Base use #empty

ahorek:
reference: https://github.com/rails/rails/commit/e17fe52e0e0ef0842b6c409e1110a862c4e005bc#diff-0f66ef96762e67b6746a49de6182c1a4

fixes
```
  1) RSpec::Rails::HelperExampleGroup#helper includes ApplicationHelper
     Failure/Error: example.run

     RuntimeError:
       Warnings were generated: DEPRECATION WARNING: ActionView::Base instances should be constructed with a lookup context,
       assignments, and a controller.
```

### DIFF
--- a/spec/rspec/rails/example/helper_example_group_spec.rb
+++ b/spec/rspec/rails/example/helper_example_group_spec.rb
@@ -42,7 +42,11 @@ module RSpec::Rails
         group = RSpec::Core::ExampleGroup.describe do
           include HelperExampleGroup
           def _view
-            ActionView::Base.new
+            if ActionView::Base.respond_to?(:empty)
+              ActionView::Base.empty
+            else
+              ActionView::Base.new
+            end
           end
         end
         expect(group.new.helper).to be_kind_of(ApplicationHelper)


spec/rspec/rails/example/helper_example_group_spec.rb
```diff
@@ -42,7 +42,11 @@ class InternalClass
         group = RSpec::Core::ExampleGroup.describe do
           include HelperExampleGroup
           def _view
-            ActionView::Base.new
+            if ActionView::Base.respond_to?(:empty)
```
JonRowe: @benoittgt we typically define methods conditionally for this sort of thing, fancy refactoring it in the dev branch?

benoittgt: I am sorry @JonRowe I didn't see your message. What do you mean by "define methods conditionally" in that case?

Thanks

JonRowe:
```Ruby
if ActionView::Base.respond_to?(:empty)
  def _view
    ActionView::Base.empty
  end
else
  def _view
    ActionView::Base.new
  end
end
```

benoittgt: Ohhh ok. Thanks for your patience. 

I made the change in https://github.com/rspec/rspec-rails/pull/2071/commits/bf65b86556c2bdd152962ac5ad17c4dcbb74b534

PR #2092, opened 2019-02-27
represent_boolean_as_integer is deprecated

ahorek:
see https://github.com/rails/rails/commit/f59b08119bc0c01a00561d38279b124abc82561b#diff-0a1a3158ff1c2956f5e096a250c11a66

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -32,7 +32,7 @@ in_root do
     append_to_file('Gemfile', "gem 'rails-controller-testing', :git => 'https://github.com/rails/rails-controller-testing'\n")
   end
 
-  if Rails::VERSION::STRING >= '5.2.0'
+  if Rails::VERSION::STRING >= '5.2.0' && Rails::VERSION::STRING < '6'
     copy_file sqlite_initializer, 'config/initializers/sqlite3_fix.rb'
   end
 


JonRowe:
Hm, I'm not sure I know what that file did, @samphippen ?

samphippen:
I’d be willing to accept this if it came with a test for rails 6 specifically that demonstrated the behaviour.

eugeneius:
Rails 5.2 emits a deprecation warning if you don't set `represent_boolean_as_integer`; Rails 6 will emit a deprecation warning if you _do_ set it. I think the only reason this file exists is to suppress the warning, and the actual behaviour toggled by the setting is irrelevant to RSpec Rails.

This change will fix the failures that are currently blocking the Rails master build on the 4-0-dev branch, so I don't think it needs a test: https://travis-ci.org/rspec/rspec-rails/jobs/499526847#L2084

JonRowe:
@eugeneius it does, because it also affects older rails versions, at the very least something need gating

ahorek:
@JonRowe I will gladly add a test, but I'm not sure what exactly should be tested...?

this change is just about silencing a deprecation warning in the example app
`Rails.application.config.active_record.sqlite3.represent_boolean_as_integer = true`
on Rails 6 this file isn't generated, because it's already a default value
on prior Rails versions the file is still generated like before, so the behaviour wasn't actually changed

JonRowe:
Actually having reviewed the output a bit more closely I'm willing to accept this as is.

PR #2089, opened 2019-02-27
EmptyTemplateHandler fix arguments

ahorek:
fixes https://github.com/rspec/rspec-rails/issues/2086

Issue #2086, opened 2019-02-26
Changes In Rails 6 ActionView::Template Breaks EmptyTemplateHandler


erich-team:
Changes to the Template handler calls in Rails 6.0 beta breaks EmptyTemplateResolver

Specifically [this change](https://github.com/rails/rails/commit/28f88e0074f473f58c2d3fd54cb3daff81027c12#diff-562906215bf781788ec5aaf37e958cb8) which changes actionview/lib/action_view/template.rb, line 307, from this...

` code = @handler.call(LegacyTemplate.new(self, source))` 

... to this ...

`code = @handler.call(self, source)` 

... breaks rspec-rails, lib/rspec/rails/view_rendering.rb, line 105

`def self.call(_template)`

with the following error;

`     ActionView::Template::Error:
       wrong number of arguments (given 2, expected 1)
`

Ruby version: ruby 2.5.1p57
Rails version: 6.0.0.beta2
Rspec version: 3.8.2

JonRowe:
Any PR for this should target `4-0-dev`

jamesjefferies:
Hi @JonRowe - targeting ```4-0-dev``` (which currently has some test failures around ```Notifications```) sounds like a good plan - but this patch will need to sniff which version of ActionView it's running against right? Is there a simple example of where that is already being done?

benoittgt:
@jamesjefferies Why not be Rails version specific? For example https://github.com/rspec/rspec-rails/blob/8c6c9590b94916199950dc8a91a9741d3be30c7c/lib/rspec-rails.rb#L10

JonRowe:
TBH I'm not sure we need to gate this, our method defines one argument, it needs two on this version of Rails, seems like simply adding a second optional argument would fix it, and gate the spec instead

jamesjefferies:
@JonRowe ah yes, that sounds like the best way to sort this .. and yes! 👋

erich-team:
@JonRowe I toyed with that in a monkey-patch

`def self.call(_template, source)`

, but given the method doesn't use `_tempalte` as it is,  it seemed a reasonable future-proof solution,  Just a thought.

JonRowe:
It'd need to be `_source = nil` in order to remain backwards compatible.

JonRowe:
Fixed in #2089

PR #2095, opened 2019-02-28
Make EmptyTemplateHandler compatible with Rails 6

palkan:
Rails version: **6.0.0.beta2**

RSpec-Rails fails with:

```
ActionView::Template::Error:
       wrong number of arguments (given 2, expected 1)
```

It turned out that templates API has changed in Rails 6: https://github.com/rails/rails/pull/35119.


JonRowe:
Duplicate of #2086 #2089 and #2091, already merged into `4-0-dev`

PR #2100, opened 2019-03-15
Various fixes for Rails 6

sebjacobs:
Whilst investigating a Rails 6 upgrade (`6.0.0.beta3`) I discovered the following issues:

* The `ActionView::Template#formats` method has been deprecated [1]
* ~The Template handler API has changed [2]~ (already fixed on `4-0-dev`) 

This branch attempts to fix these issues whilst retaining support for Rails <= 5.2.

[1] https://www.github.com/rails/rails/commit/ca5e23ed4d8
[2] https://www.github.com/rails/rails/commit/28f88e0074

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -56,8 +56,18 @@ module RSpec
               template.identifier,
               EmptyTemplateHandler,
               :virtual_path => template.virtual_path,
-              :format => template.formats
+              :format => template_format(template)
             )
+          end
+        end
+
+        if ::Rails::VERSION::STRING >= '6'
+          def self.template_format(template)
+            template.format
+          end
+        else
+          def self.template_format(template)
+            template.formats
           end
         end
 


JonRowe:
Hi, thanks for this, Rails 6 support work is on going in `4-0-dev` so you need to target that, your second commit isn't necessary as that fix has already been done, although I don't think we've seen the formats issue before.

JonRowe:
I suggest you rebase if you can!

sebjacobs:
@JonRowe Argh ok sorry about that, but I hadn't spotted the `4-0-dev` branch. Thanks for taking a look!

I have rebased the branch against `4-0-dev`.

I have rebased the branch

sebjacobs:
Argh ok sorry about that I hadn't spotted that branch. I have rebased as requested.

Thanks for taking a look!

lib/rspec/rails/view_rendering.rb
```diff
@@ -56,7 +56,7 @@ def self.nullify_template_rendering(templates)
               template.identifier,
               EmptyTemplateHandler,
               :virtual_path => template.virtual_path,
-              :format => template.formats
+              :format => template.respond_to?(:format) ? template.format : template.formats
```
JonRowe: Great, is this a Rails specific or object specific? e.g. will we encounter both methods in one version of Rails? If so we prefer conditionally defining a method to extract this to save runtime conditionals.

sebjacobs: `template.formats` is deprecated in Rails 6, so we could rely on the Rails version instead, perhaps something like this?:

```
def self.template_format(format)
  if Rails::VERSION::STRING >= '6'
    template.format
  else
    template.formats
  end
end
```

JonRowe: Right so if this is Rails 6 only we'd typically do:
```
if Rails::VERSION::STRING >= '6'
  def self.template_format(format)
    template.format
  end
else
  def self.template_format(format)
    template.formats
  end
end
```

sebjacobs: That makes sense, thanks @JonRowe!

sebjacobs:
I have fixed up the suggested changes in (712c099)

JonRowe:
Thanks!

benoittgt:
Thanks a lot @sebjacobs

PR #2106, opened 2019-04-01
[ruby-head] Turn value into a String before regex match

olleolleolle:
This PR is tries to avoid a warning output in 2.7.0 (`ruby-head`), which causes a test failure in `rspec-core`:

> expected `"/home/travis/build/rspec/rspec-rails/Gemfile-rails-dependencies:49:
warning: deprecated Object#=~ is called on FalseClass; it always returns nil\n".empty?` to return true, got false

[Error in rspec-core Travis build](https://travis-ci.org/rspec/rspec-core/jobs/514297628#L3562)

This PR turns the value (`version`) being matched with `=~` into a String, first.

The change to ruby was introduced in https://svn.ruby-lang.org/cgi-bin/viewvc.cgi?revision=65989&view=revision


olleolleolle:
There's also the possibility of defaulting to `nil` instead of `''`: `NilClass#=~` _is_ defined.

JonRowe:
I'm happy with this as it is now it passes, thanks :)

PR #2107, opened 2019-04-14
CI: Drop sudo: false

olleolleolle:
- Removes an old setting from Travis. See https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: ruby
 
-sudo: false
-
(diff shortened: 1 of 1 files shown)


benoittgt:
I think we have to follow https://github.com/rubygems/rubygems/issues/2534

Do you think you can try?

benoittgt:
Thanks @olleolleolle

PR #2112, opened 2019-04-20
Fix builds by swapping to webdrivers instead of locking selenium-webdriver

twalpole:
`chromedriver-helper` was deprecated in favor of the `webdrivers` gem which includes handling of the `selenium-webdriver` deprecations.  This PR swaps to the `webdrivers` gem (as current Rails has) rather than locking `selenium-webdriver`

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -33,7 +33,7 @@ in_root do
   end
 
   if Rails::VERSION::STRING >= "5.1.0"
-    gsub_file "Gemfile", /.*selenium-webdriver.*/, "gem 'selenium-webdriver', '<= 3.14'"
+    gsub_file "Gemfile", /.*chromedriver-helper.*/, "gem 'webdrivers'"
   end
 
   if Rails::VERSION::STRING >= '5.2.0' && Rails::VERSION::STRING < '6'


benoittgt:
Source for `chromedriver-helper`: https://github.com/flavorjones/chromedriver-helper#notice-this-gem-is-out-of-support-as-of-2019-03-31
Similar change made on Rails recently https://github.com/rails/rails/commit/2512f24ef40f4c678c8dc364cdcaff09d4d5a65d

I ok with that change.

Maybe we could add it to #2071 @samphippen ?

example_app_generator/generate_app.rb
```diff
@@ -33,7 +33,7 @@
   end
 
   if Rails::VERSION::STRING >= "5.1.0"
-    gsub_file "Gemfile", /.*selenium-webdriver.*/, "gem 'selenium-webdriver', '<= 3.14'"
+    gsub_file "Gemfile", /.*chromedriver-helper.*/, "gem 'webdrivers'"
```
benoittgt: I think we should still keep specifying a version.

twalpole: Keep specifying a version of what? This just reverts the selenium-webdriver specification to whatever it was before https://github.com/rspec/rspec-rails/commit/e10484e3ea3352afeac5edfcc62ab456f357d53c which was done because of the deprecation notices emitted by the latest selenium-webdriver.  Beyond that `webdrivers` specifies it's compatibility with `selenium-webdriver` drivers and is added just like in Rails now (which is without any specific version specified).

benoittgt: I am ok with this then.

JonRowe:
Thanks!

PR #2117, opened 2019-05-04
add missing require to have_enqueued_mail

ignatiusreza:
Hiya, I'm having issue when trying out #2047 in the beta release.. the error raised was `You must pass an argument rather than a block to 'expect' to use the provided matcher (have enqueued mail ReservationMailer, :complete), or the matcher must implement 'supports_block_expectations?'.`

which seems to be caused by the matcher not being required.. hence, this PR fix it by adding the missing `require` call..

### DIFF
--- a/lib/rspec/rails/matchers.rb
+++ b/lib/rspec/rails/matchers.rb
@@ -23,6 +23,7 @@ require 'rspec/rails/matchers/have_http_status'
 
 if RSpec::Rails::FeatureCheck.has_active_job?
   require 'rspec/rails/matchers/active_job'
+  require 'rspec/rails/matchers/have_enqueued_mail'
 end
 
 if RSpec::Rails::FeatureCheck.has_action_cable_testing?
--- a/features/matchers/have_enqueued_mail_matcher.feature
+++ b/features/matchers/have_enqueued_mail_matcher.feature
@@ -1,0 +1,78 @@
+Feature: have_enqueued_mail matcher
+
+  The `have_enqueued_mail` (also aliased as `enqueue_mail`) matcher is used to check if given mailer was enqueued.
+
+  Background:
+    Given active job is available
+
+  @rails_post_5
+  Scenario: Checking mailer class and method name
+    Given a file named "spec/mailers/user_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "matches with enqueued mailer" do
+          ActiveJob::Base.queue_adapter = :test
+          expect {
+            NotificationsMailer.signup.deliver_later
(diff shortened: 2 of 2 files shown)


lib/rspec/rails/matchers.rb
```diff
@@ -22,4 +22,5 @@ module Matchers
 require 'rspec/rails/matchers/have_http_status'
 if RSpec::Rails::FeatureCheck.has_active_job?
   require 'rspec/rails/matchers/active_job'
+  require 'rspec/rails/matchers/have_enqueued_mail'
```
benoittgt: Don't we need a test?

ignatiusreza: Hi Benoit, thanks for the review!

I added a cucumber feature for `have_enqueued_mail` matcher, which would fail without this `require` line..

samphippen:
Hi there @ignatiusreza I ran our build against this PR and got the following failing features:

```
cucumber features/matchers/have_enqueued_mail_matcher.feature:9 # Scenario: Checking mailer class and method name
cucumber features/matchers/have_enqueued_mail_matcher.feature:26 # Scenario: Checking mailer enqueued time
cucumber features/matchers/have_enqueued_mail_matcher.feature:43 # Scenario: Checking mailer enqueued with no wait
cucumber features/matchers/have_enqueued_mail_matcher.feature:60 # Scenario: Using alias method
```

do I need to do something to get these green ?

ignatiusreza:
hi @samphippen, I rebased the branch to run this PR against CI, and just noticed that it's failing when `RAILS_VERSION` is not on `5.*` family.. sorry about that, i'll investigate further..

ignatiusreza:
I dig down a bit deeper, and found that the default mailer delivery job changed in rails 6 (see: https://github.com/rails/rails/pull/34591 & https://github.com/rails/rails/pull/34692).. in our test environment, `rspec` passes since it's still using `ActionMailer::DeliveryJob`, while `cucumber` failed since it is run against freshly generated rails 6 application and uses `ActionMailer::MailDeliveryJob`..

since this PR is about adding missing `require` line, should I open a different PR to update the job matching logic? or should I add it here? incidentally, I also opened a different PR for adding support for parameterized mailer in #2121.. so maybe it's best to consolidate the effort in that PR? wdyt?

JonRowe:
I'm happy with this if you are @samphippen

samphippen:
@ignatiusreza can you look at the build failures here and let us know what's up or if you need some help?

ignatiusreza:
@samphippen i found the issue and opened #2125 to address it.. i'll rebase this branch after #2125 is merged, and ci should pass..

ignatiusreza:
All green now!

samphippen:
LGTM

JonRowe:
This has also been released in 3.9.1 now.

PR #2118, opened 2019-05-04
allow partial args when using have_enqueued_mail

ignatiusreza:
this provide parity with `have_enqueued_job` with 0 arguments which would be useful especially in negated case where we want to make sure that absolutely no emails are enqueued


benoittgt:
Hello

Thanks a lot for this PR. It looks good to me.

JonRowe:
Thanks! :)

PR #2119, opened 2019-05-04
Add ActionMailbox spec helpers and test type

jamesdabbs-procore:
Adds the following helpers to example groups with `:type => :mailbox`

* [`process(mail_or_attributes)`](https://github.com/rspec/rspec-rails/pull/2119/files#diff-079fe3f19ddac323e429d712ead9c3efR67) - send mail directly to the mailbox under test for `process`ing.
* [`receive_inbound_email(mail_or_attributes)`](https://github.com/rspec/rspec-rails/pull/2119/files#diff-de58799c3a6e6969f826a234f594b58dR59) - matcher for asserting whether incoming email would route to the mailbox under test.
* [`have_been_delivered`](https://github.com/rspec/rspec-rails/pull/2119/files#diff-079fe3f19ddac323e429d712ead9c3efR50) - matcher for asserting whether an incoming email object was delivered.
* [`have_bounced`](https://github.com/rspec/rspec-rails/pull/2119/files#diff-079fe3f19ddac323e429d712ead9c3efR54) - matcher for asserting whether an incoming email object has bounced.
* [`have_failed`](https://github.com/rspec/rspec-rails/pull/2119/files#diff-079fe3f19ddac323e429d712ead9c3efR58) - matcher for asserting whether an incoming email object has failed.

Also adds an [ActionMailbox test generator](https://github.com/rspec/rspec-rails/pull/2119/files#diff-474e4701121ece1c71a1b90220f1ff8b).


lib/rspec/rails/example/mailbox_example_group.rb
```diff
@@ -0,0 +1,74 @@
+module RSpec
+  module Rails
+    # @api public
+    # Container module for mailbox spec functionality.
+    module MailboxExampleGroup
+      extend ActiveSupport::Concern
+
+      if RSpec::Rails::FeatureCheck.has_action_mailbox?
+        require 'action_mailbox/test_helper'
+        extend ::ActionMailbox::TestHelper
```
jamesdabbs-procore: I don't love this, but `ActionMailbox` appears to expect to operate on a particular named `ActionMailbox::InboundEmail < ActiveRecord::Base`, and I'd be worried that using some kind of double here might be brittle.

lib/rspec/rails/matchers/action_mailbox.rb
```diff
@@ -0,0 +1,64 @@
...
+        class ReceiveInboundEmail < Base
+          def initialize(message)
+            super()
+
+            @inbound_email = create_inbound_email(message)
+          end
+
+          def matches?(mailbox)
+            @mailbox  = mailbox
+            @receiver = ApplicationMailbox.router.send(:match_to_mailbox, inbound_email)
```
jamesdabbs-procore: `actionmailbox`'s public API does not expose a way to check which mailbox an email routes to that doesn't fully send the email to the mailbox for processing. `expect(mailbox).to receive(:receive)` also seems fraught because of the method name there.

I'll open up a PR upstream to expose something like this in the public API. It seems like a generally useful thing for testing, and I hope won't be too hard to land. If it does, I'll update the implementation here.

samphippen:
Thank you for this PR! It looks like overall the quality is pretty high :)

I'll review this soon

lib/rspec/rails/matchers/action_mailbox.rb
jamesdabbs: PR for that - https://github.com/rails/rails/pull/36181

lib/rspec/rails/example/mailbox_example_group.rb
benoittgt: Why don't you like it?

lib/rspec/rails/matchers/action_mailbox.rb
```diff
@@ -0,0 +1,64 @@
...
+
+            @receiver == @mailbox
+          end
+
+          def failure_message
+            "expected #{describe_inbound_email} to route to #{mailbox}".tap do |msg|
+              if receiver
```
benoittgt: Nice :)

benoittgt:
Thanks for your PR. I find it clear.

> Thoughts on the API exposed to tests? Anything you'd like to see added / changed?

It seems ok for me. 

> Does the feature toggling look reasonable? (Does travis not run on PRs into non-master branches?)

It is ok for me. For the CI yes. I have open a PR for that https://github.com/rspec/rspec-rails/pull/2120

>    Any preference for how to handle ActionMailbox::InboundEmail? What level of integration do we want to provide there? Personally, I'd prefer for matchers like process_inbound_email not to have to write to the database.

For your first question. I re-read [the doc](https://edgeguides.rubyonrails.org/action_mailbox_basics.html#testing-mailboxes) of Action MailBox but didn't find things we should implement now.

> I'd prefer for matchers like process_inbound_email not to have to write to the database.

Yes!

lib/generators/rspec/mailbox/mailbox_generator.rb
```diff
@@ -0,0 +1,12 @@
+require 'generators/rspec'
+
+module Rspec
+  module Generators
+    # @private
+    class MailboxGenerator < Base
+      def create_mailbox_spec
+        template 'mailbox_spec.rb.erb', File.join('spec/mailboxes', class_path, "#{file_name}_mailbox_spec.rb")
```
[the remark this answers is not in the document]
jamesdabbs: Updated this to be more consistent with the other generators that do have a line break here. Let me know if you meant for something like

```
        template(
          'mailbox_spec.rb.erb',
          File.join(
            'spec/mailboxes',
            class_path,
            "#{file_name}_mailbox_spec.rb"
          )
        )
```

instead though.

lib/rspec/rails/example/mailbox_example_group.rb
jamesdabbs: This is here mostly because [`ActionMailbox::InboundEmail.create_and_extract_message_id!` is doing a lot of work](https://github.com/rails/rails/blob/3d7391cb76758c82c8ea7759b6dd8d4402bc2346/actionmailbox/app/models/action_mailbox/inbound_email/message_id.rb#L16) that I didn't want to duplicate here, but this feels like I'm importing implementation details, not clean abstractions.

My inclination here is to do something like

* define a lightweight e.g. `RSpec::Rails::ActionMailbox::MockInboundEmail` that we use internally in `process_inbound_email`
* add shared examples for `::ActionMailbox::InboundEmail` and `MockInboundEmail` to make sure they are interchangeable
* consider using `MockInboundEmail` for the `process` helper (either by default or opt-in somehow?)
* work on landing `MockInboundEmail` upstream

How does that sound to y'all?

benoittgt: I love the idea of a `MockInboundEmail` but I think it can be another PR?

benoittgt:
The CI fail with missing documentation

```
Undocumented Objects:
(in file: lib/rspec/rails/example/mailbox_example_group.rb)
RSpec::Rails::MailboxExampleGroup.create_inbound_email
RSpec::Rails::MailboxExampleGroup#have_bounced
RSpec::Rails::MailboxExampleGroup#have_failed
Missing documentation coverage (currently at 97.84%)
The command "script/run_build 2>&1" exited with 1.
```

https://travis-ci.org/rspec/rspec-rails/jobs/530437674

@jamesdabbs Do you think you can open a PR to fix this? :)

benoittgt:
Ok so the CI is failing for something else now. https://travis-ci.org/rspec/rspec-rails/jobs/530476234

I merged https://github.com/rspec/rspec-rails/pull/2120 so you could open a PR from updated `4-0-dev` and CI should build your PR. It will make the fix easier to validate before merging it. ;)

spec/rspec/rails/example/mailbox_example_group_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      self.received += 1
+    end
+  end
+
+  self.received = 0
+end
+
+module RSpec
+  module Rails
+    describe MailboxExampleGroup, :skip => !RSpec::Rails::FeatureCheck.has_active_job? do
```
ignatiusreza: this should check for `has_action_mailbox?` instead of `has_active_job?`..

spec/rspec/rails/matchers/action_mailbox_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def self.router
+    Router.new
+  end
+end
+
+class Inbox < ApplicationMailbox; end
+class Otherbox < ApplicationMailbox; end
+
+RSpec.describe "ActionMailbox matchers", :skip => !RSpec::Rails::FeatureCheck.has_active_job? do
```
ignatiusreza: same as above

spec/rspec/rails/example/mailbox_example_group_spec.rb
jamesdabbs: Thank you! Opened in https://github.com/rspec/rspec-rails/pull/2123

PR #2121, opened 2019-05-07
add support for parameterized mailer when RAILS_VERSION > 5.1

ignatiusreza:
rails 5.1 introduced parameterized email via https://github.com/rails/rails/pull/27825.. this PR introduce support for using the recently added `have_enqueued_email` to recognize parameterized email.. 

PS: might conflict with #2118

PR #2125, opened 2019-05-13
add support for recognizing mailer many jobs:

ignatiusreza:
- parameterized mailer when RAILS_VERSION >= 5.1
- unified mailer job when RAILS_VERSION >= 6.0

per https://github.com/rspec/rspec-rails/pull/2117#issuecomment-491651319

this PR supersede #2121 with added support for recognizing `ActionMailer::MailDeliveryJob` freshly added into rails 6.0

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -97,7 +97,7 @@ module RSpec
 
           def check(jobs)
             @matching_jobs, @unmatching_jobs = jobs.partition do |job|
-              if arguments_match?(job) && other_attributes_match?(job)
+              if job_match?(job) && arguments_match?(job) && other_attributes_match?(job)
                 args = deserialize_arguments(job)
                 @block.call(*args)
                 true
@@ -134,6 +134,10 @@ module RSpec
             end
           end
 
+          def job_match?(job)
+            @job ? @job == job[:job] : true
+          end
+
           def arguments_match?(job)
             if @args.any?
               deserialized_args = deserialize_arguments(job)
@@ -151,7 +155,6 @@ module RSpec
             {}.tap do |attributes|
               attributes[:at]    = serialized_at if @at
               attributes[:queue] = @queue if @queue
-              attributes[:job]   = @job if @job
             end
           end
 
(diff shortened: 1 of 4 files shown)


ignatiusreza:
got lucky with rubocop new release, which break CI.. opened #2126 to fix CI.. will rebased this branch after CI is fixed..

benoittgt:
> this PR supersede #2121 

Can we close is then?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -134,6 +134,10 @@ def base_job_message(job)
             end
           end
 
+          def job_match?(job)
+            @job ? @job == job[:job] : true
```
benoittgt: In which case @job will be nil of false?

ignatiusreza: `@job` will be `nil` when matcher is called without explicit job, e.g.

```rb
expect {
  AJob.perform_later
}.to have_enqueued_job
```

benoittgt: ok

benoittgt: I just seem weird to have `job_match?(job)` that returns true when there is no job

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -63,20 +68,18 @@ def expected_count_message
           "#{message_expectation_modifier} #{@expected_number} #{@expected_number == 1 ? 'time' : 'times'}"
         end
 
-        def mailer_args
+        def job_match?(job)
+          MAILER_JOBS.any? { |mailer_job| send(mailer_job, job) }
```
JonRowe: I think I'd rather we just called `legacy_mail?(mailer_job, job) || parameterized_mail?(mailer_job, job) || unified_mail?(mailer_job, job)` rather than constantising it and iterating.

ignatiusreza: sure, updated.. I used constant because I thought it would be easier later IF we somehow need to add check for other job..

benoittgt:
@samphippen are you ok to merge this? 😊

PR #2126, opened 2019-05-13
address rubocop offenses:

ignatiusreza:
- fix warnings for wrong namespaces
- fix safe navigation offense
- turn off frozen string literal cop

rubocop released new [version](https://github.com/rubocop-hq/rubocop/releases/tag/v0.69.0), which break our CI..

PS: let me know if I should open this against master instead..

### DIFF
--- a/lib/rspec/rails/example/mailer_example_group.rb
+++ b/lib/rspec/rails/example/mailer_example_group.rb
@@ -22,7 +22,7 @@ if defined?(ActionMailer)
         included do
           include ::Rails.application.routes.url_helpers
           options = ::Rails.configuration.action_mailer.default_url_options
-          options.each { |key, value| default_url_options[key] = value } if options
+          options&.each { |key, value| default_url_options[key] = value }
         end
 
         # Class-level DSL for mailer specs.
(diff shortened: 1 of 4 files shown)


lib/rspec/rails/example/mailer_example_group.rb
```diff
@@ -22,7 +22,7 @@ module MailerExampleGroup
         included do
           include ::Rails.application.routes.url_helpers
           options = ::Rails.configuration.action_mailer.default_url_options
-          options.each { |key, value| default_url_options[key] = value } if options
+          options&.each { |key, value| default_url_options[key] = value }
```
JonRowe: What versions of Ruby does that operator support?

ignatiusreza: [>= 2.3](https://www.ruby-lang.org/en/news/2015/12/25/ruby-2-3-0-released/).. v4 will drop support for ruby below 2.3, so using lonely operator should be safe..

JonRowe:
Thanks :)

benoittgt:
Yeha! Thanks Jon and Ignatius

Issue #1993, opened 2018-05-08
Unable to set cookies to be passed in request specs


raldred:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.5.0
Rails version: 5.2.0
Rspec version: 3.7

## Observed behaviour

Cookies are not available in the controller

## Expected behaviour

Cookies should be available in the controller

spec
```
RSpec.describe "logout", type: :request do
  it "removes the cookie" do
    cookies['_cookiename'] = 'abc123'
    get "/logout"
    expect(response.cookies['_cookiename']).to_not be_present
  end
end
```

controller
```
def logout
  p cookies # <-- will output {} no cookies are available
  reset_session
  cookies.delete('_cookiename')
end
```

Works fine in controller specs

ebihara99999:
Request spec doesn't support testing cookies, but controller spec does. What's the problem with you using controller spec?

raldred:
No problem, like I said controller specs are fine, merely raising the gap in functionality.

mikegee:
Controller specs [are discouraged](https://github.com/rspec/rspec-rails#controller-specs) these days. I wouldn't expect to see features added.

raldred:
agree @mikegee, in a similar fashion to passing params and headers.
I suspect this is really down to a limitation of Rails' integration tests.

I wonder if you can do...

```
headers: {"HTTP_COOKIE" => "key=value; key2=value2;"}
```

JonRowe:
Most of the behaviour for these sets of specs is direct from Rails, can you see if it works in the minitest equivalent for me?

samphippen:
I actually just tried to reproduce this and failed.

@raldred 
Thanks for the issue. We need a little more detail to be able to reproduce this.

Could you please provide us with a rails app that we can clone that demonstrates the issue. Specifically it'd be great if

1)  you could `rails new` an application and commit
2) make all the changes necessary to reproduce the issue and commit

then, provide us with a description of how to clone your application and reproduce the issue.

Thanks :)

taki:
@salimane I reproduce this issue. Please check if you have the time.

## Steps to reproduce

This is  reproduced Rails application.

1. git clone `git@github.com:taki/rpsec-rails-does-not-set-cookie.git`
1. cd rpsec-rails-does-not-set-cookie
1. bundle install
1. bin/rails db:setup
1. bundle exec rspec spec/requests/sign_outs_controller_spec.rb

## Environment

[...]

PR #2127, opened 2019-05-13
Simplify the way we explain how to test cookies in controller

benoittgt:
The previous documentation was old and was not clear enough to help
developers to understand properly how to test cookies inside a
controller test. Since then it is much easier to test cookies if you
stick to `cookies` and don't use `response.cookies`.

You can use response.cookies when you don't change cookies inside your
test.

```ruby
def show
  cookies["user_name"] = nil
  head 200
end

RSpec.describe SignOutsController, type: :request do
  describe 'GET /signout' do
    it "clear cookie value" do
      get "/sign_out"

      expect(response.cookies["user_name"]).to eq(nil)
    end
  end
end
```

Use `cookies` to set and expect when you set cookies inside your test

```ruby
def show
  cookies.delete("user_name")
  head 200
end

RSpec.describe SignOutsController, type: :request do
  describe 'GET /signout' do
    it "clear cookie value" do
      cookies["user_name"] = "Sam"

      get "/sign_out"

      expect(cookies["user_name"]).to eq("")
    end
  end
end
```

This is the preferred way.

Related:
- https://github.com/rspec/rspec-rails/issues/1993


benoittgt:
Can we stick rubocop version...? It is failing way too often.

features/controller_specs/cookies.feature
```diff
@@ -0,0 +1,36 @@
+Feature: Cookies
+
+  There is different ways to test cookies in controller's tests be we recommend
+  to use `cookies` as below.
```
JonRowe: I think this needs rewording, theres a few grammatical issues. How about:
```
There are different ways to make assertions on cookies from controller specs, but we recommend using the `cookies` method as set out below.
```

features/controller_specs/cookies.feature
```diff
@@ -0,0 +1,36 @@
...
+
+  You can use strings or symbols to fetch or set your cookies because `cookies`
+  support indifferent access.
```
JonRowe: Swap out '`cookies` support' for 'the `cookies' method supports'

features/controller_specs/cookies.feature
benoittgt: Thanks. It is much better with your proposal. Changed with https://github.com/rspec/rspec-rails/pull/2127/commits/54104d8ea3ce6c4f75be17518f34593799930371

features/controller_specs/cookies.feature
benoittgt: Thanks. Done with https://github.com/rspec/rspec-rails/pull/2127/commits/54104d8ea3ce6c4f75be17518f34593799930371

JonRowe:
LGTM, merge on green!

PR #2130, opened 2019-05-18
Check for ActionCable const before adding matchers

mikker:
If your app hasn't requried `action_cable`, running `rspec` will raise `NameError: uninitialized constant RSpec::Rails::ChannelExampleGroup::ActionCable`. This fixes it.

### DIFF
--- a/lib/rspec/rails/example/channel_example_group.rb
+++ b/lib/rspec/rails/example/channel_example_group.rb
@@ -15,7 +15,7 @@ module RSpec
   end
 end
 
-if ::Rails::VERSION::MAJOR >= 6
+if ::Rails::VERSION::MAJOR >= 6 && defined?(ActionCable)
   module RSpec
     module Rails
       # @api public


lib/rspec/rails/example/channel_example_group.rb
```diff
@@ -15,7 +15,7 @@ module ClassMethods
   end
 end
 
-if ::Rails::VERSION::MAJOR >= 6
+if ::Rails::VERSION::MAJOR >= 6 && defined?(ActionCable)
```
benoittgt: Maybe we could use instead `RSpec::Rails::FeatureCheck.has_action_cable_testing?`

https://github.com/rspec/rspec-rails/pull/2071/files#diff-bf13f1e6c52828697cb1c321683f5360R36

WDYT @palkan?

palkan: > Maybe we could use instead RSpec::Rails::FeatureCheck.has_action_cable_testing?

I think, unifying checks is a good idea.
Though for some (maybe, historical) reason it hasn't been used in `rspec-rails` (I used Active Job integration as an example).

And we have `feature_check` required after example groups, probably, for a reason (I don't know about).

🤷‍♂️

JonRowe: 👍 for using a name feature flag

benoittgt:
Fixed with https://github.com/rspec/rspec-rails/pull/2071/commits/1eb171b68e3ae6ad3b605d55a224485ad68087da

@mikker could you try it?

mikker:
Works On My Machine™ 👍 🙏

PR #2093, opened 2019-02-27
jbuilder

ahorek:
use the master branch until https://github.com/rails/jbuilder/pull/453 is released

fixes
```
Action Mailer railtie hook in a non-development environment respects the setting from `show_previews`
     Failure/Error:
       expect(
         capture_exec(custom_env.merge('SHOW_PREVIEWS' => 'true'), exec_script)
       ).to eq("#{::Rails.root}/spec/mailers/previews")
     
       expected: "/home/travis/build/rspec/rspec-rails/tmp/example_app/spec/mailers/previews"
            got: #<struct CaptureExec io="DEPRECATION WARNING: Single arity template handlers are deprecated.  Templat....rb:37)\n/home/travis/build/rspec/rspec-rails/tmp/example_app/spec/mailers/previews", exit_status=0>
     
       (compared using ==)
     
       Diff:
       @@ -1,2 +1,2 @@
       -"/home/travis/build/rspec/rspec-rails/tmp/example_app/spec/mailers/previews"
       +"#<struct CaptureExec io=\"DEPRECATION WARNING: Single arity template handlers are deprecated.  Templat....rb:37)\\n/home/travis/build/rspec/rspec-rails/tmp/example_app/spec/mailers/previews\", exit_status=0>"
...
```

PR #2133, opened 2019-05-23
use jbuilder from rubygems

ahorek:
this reverts https://github.com/rspec/rspec-rails/pull/2093 because it's not necessary anymore


JonRowe:
Thanks!

PR #2145, opened 2019-07-08
Use trusty for TravisCI build

okuramasafumi:
RVM doesn't provide old Ruby binaries for newer Ubuntu distributions.
Until it's fixed, we need to use old Ubuntu distributions in CI.


JonRowe:
Thanks!

PR #2152, opened 2019-07-28
A fix for ArgumentError in Rails 6 controller spec

amatsuda:
With Rails 6 (rc2) pre and current stable rspec-rails (3.8.2), scaffolded controller spec fails with

> ArgumentError: wrong number of arguments (given 2, expected 1)

due to Action View handler API change at https://github.com/rails/rails/commit/28f88e0074f473f58c2d3fd54cb3daff81027c12.

This patch fixes the error by adding an optional parameter to `EmptyTemplateHandler.call` so that it works with both Rails < 6 and Rails >= 6.


JonRowe:
:wave: I've changed the base of this to the appropriate branch, Rails 6 support is in `4-0-dev` as we will be releasing a new version to support Rails 6 locked to the relevant supported versions of Rails. Can you check any conflicts and re/base merge as necessary to ensure only relevant changes are introduced? Thanks!

amatsuda:
@JonRowe Oh, I'm sorry I didn't know the way you use Git here...
I found that there already was an exactly the same commit in 4-0-stable branch (377ece399e01860210316de181be5f77cf261f2f), and it's been included in 4.0.0.beta2 release.

I'm sorry for bothering you with this. I should've tried a prerelease version before sending a patch. 😟

JonRowe:
No problem thanks for contributing!

PR #2157, opened 2019-08-17
have_enqueued_job supports time object

alpaca-tc:
Rails6 can pass Time as arguments to ActiveJob but `have_enqueued_job#with` doesn't support time object.
Deserialized time has 0 usec value, it will fail when compared with original time.

```ruby
time = Time.now
expect { job.perform_later(time) }.to have_enqueued_job.with(time) #=> failed
```

This PR fixes it by to serialize/deserialize arguments before comparing.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -140,8 +140,9 @@ module RSpec
 
           def arguments_match?(job)
             if @args.any?
+              args = serialize_and_deserialize_arguments(@args)
               deserialized_args = deserialize_arguments(job)
-              RSpec::Mocks::ArgumentListMatcher.new(*@args).args_match?(*deserialized_args)
+              RSpec::Mocks::ArgumentListMatcher.new(*args).args_match?(*deserialized_args)
             else
               true
             end
@@ -170,6 +171,13 @@ module RSpec
                                when :thrice then 3
                                else Integer(count)
                                end
+          end
+
+          def serialize_and_deserialize_arguments(args)
+            serialized = ::ActiveJob::Arguments.serialize(args)
+            ::ActiveJob::Arguments.deserialize(serialized)
+          rescue ::ActiveJob::SerializationError
+            args
           end
 
           def deserialize_arguments(job)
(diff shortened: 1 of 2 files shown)


JonRowe:
Can you retarget this to `4-0-dev` all Rails 6 patches are there /cc @samphippen

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -172,6 +173,13 @@ def set_expected_number(relativity, count)
                                end
           end
 
+          def serialize_and_deserialize_arguments(args)
+            serialized = ::ActiveJob::Arguments.serialize(args)
```
benoittgt: I am surprise we have to serialize then deserialize. I am wondering if it can be simpler?

https://github.com/rails/rails/blob/6-0-stable/activejob/lib/active_job/serializers/time_serializer.rb

alpaca-tc: I changed to use simple time rounder instead of serializer/deserializer.
It is copied from ActiveJob::TestHelper.

https://github.com/rails/rails/blob/c986369095ad89fd68fd84d21aa39006ea77904d/activejob/lib/active_job/test_helper.rb#L638

JonRowe: I prefer the original implementation, it's simpler as it replicates what Rails does without special casing, the new implementation may not catch other edgecases and we're just reinventing what Rails is doing.

JonRowe: I'd also like to see a green build, which is why I started #2158

alpaca-tc: I also like the original implementation, so I reverted. And, I deleted `ActiveJob::DeserializationError` because it never occurred.

benoittgt: Thanks for the change and sorry for the back and forth.

JonRowe:
Can you cherry-pick 4eeb9ac into this to see if you get a green build?

JonRowe:
I'm happy to merge this if you are @benoittgt I think the failures are the same as 4-0-dev normally right?

benoittgt:
Yes and yes Jon. I will try tonight to fix the build on 4-0-dev.

benoittgt:
Sorry last call. Can you rebase from 4-0-dev?

Thanks a lot

alpaca-tc:
@benoittgt All green now 💚 🎉

JonRowe:
Thanks! 💚

PR #2163, opened 2019-08-22
Update docs: :all => :context, :each => :example

giovannibonetti:
This documentation page is very important but the code is a bit outdated.
I just replaced `:all` by `:context`, and `:each` by `:example`


JonRowe:
Thanks!

PR #2173, opened 2019-09-23
Pin minitest on older Rubies

JonRowe:


JonRowe:
Hey @benoittgt care to re-review? I had to turn off part of the build to make this work but I think thats an ok trade off.

benoittgt:
LGTM Jon

PR #2174, opened 2019-09-23
Pin minitest on older Rubies 

JonRowe:


benoittgt:
Thanks Jon. Green build.

PR #2183, opened 2019-10-09
Delete unused script file

aried3r:
Rubyforge went offline in 2014. Didn't `[ci skip]` in case it's used somewhere after all.


JonRowe:
Thanks :)

PR #2198, opened 2019-10-31
Use latest rocket hash

hmtanbir:
Use latest rocket hash, removing old rocket hash from routing spec template.
Could you please review @brynary ?


pirj:
Thanks!
Are you up to do the same with the rest of templates?

hmtanbir:
I do not check yet, let me check.

PR #2206, opened 2019-11-10
Support argument matchers in have_enqueued_*

### LINKED ISSUE #2205: approximate at for activejob
I queue an email to be sent 24 hours after a certain controller action. When I try to test it, I currently have

```
expect { commit }.to have_enqueued_mail(...).at(Time.now + 24.hours)
```
The problem is that `Time.now + 24.hours` is not precisely equal to what it was during `commit`, when the job was created, so the test fails.

What I would hope to have is something like
```
expect { commit }.to have_enqueued_mail(...).approx_at(Time.now + 24.hours, err)
```
which passes if the given time is within `err` from the job time (perhaps with some default value for `err`).

As a more general alternative, it would perhaps be useful if one could easily override the class used by the matchers, say by a derived class, and then I could implement it myself (but I think the feature I suggest could be useful in general)

pirj:
Allows for:

```ruby
have_enqueued_job.at(a_value_within(5.seconds).of(future))
have_enqueued_mail.at(a_value_within(5.seconds).of(future))
```

Fixes #2205

`Time.at` seems to be quite fine as per https://github.com/rails/rails/blob/a707072/activejob/lib/active_job/test_helper.rb#L679

@mkamensky I can backport to 3.9 if you like, or appreciate if you do so if this one change gets through.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -97,7 +97,7 @@ module RSpec
 
           def check(jobs)
             @matching_jobs, @unmatching_jobs = jobs.partition do |job|
-              if job_match?(job) && arguments_match?(job) && other_attributes_match?(job)
+              if job_match?(job) && arguments_match?(job) && queue_match?(job) && at_match?(job)
                 args = deserialize_arguments(job)
                 @block.call(*args)
                 true
@@ -148,19 +148,18 @@ module RSpec
             end
           end
 
-          def other_attributes_match?(job)
-            serialized_attributes.all? { |key, value| value == job[key] }
+          def queue_match?(job)
+            return true unless @queue
+
+            @queue == job[:queue]
           end
 
-          def serialized_attributes
-            {}.tap do |attributes|
-              attributes[:at]    = serialized_at if @at
-              attributes[:queue] = @queue if @queue
-            end
-          end
+          def at_match?(job)
+            return true unless @at
+            return job[:at].nil? if @at == :no_wait
+            return false unless job[:at]
 
-          def serialized_at
-            @at == :no_wait ? nil : @at.to_f
+            values_match?(@at, Time.at(job[:at]))
           end
 
           def set_expected_number(relativity, count)
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,7 @@ Enhancements:
 * Improve path parsing in view specs render options. (John Hawthorn, #2115)
 * Add routing spec template as an option for generating controller specs.
   (David Revelo, #2134)
+* Add argument matcher support to `have_enqueued_*` matchers. (Phil Pirozhkov, #2206)
 
 Bug Fixes:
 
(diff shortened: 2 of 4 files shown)


glenpike:
Maybe a dumb question, but will this also support `have_been_enqueue_at`, e.g.

```
before do
  #in our before, we actually call AnotherJob.perform_now which calls the following:
  MyJob.set(wait: 7.days).perform_later('123')
end

it 'queues MyJob job' do
  expect(MyJob).to have_been_enqueued.exactly(:once).at(a_value_within(1.minute).of(7.days.from_now))
end
```

pirj:
@glenpike Absolutely.
Added a spec for this.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -148,19 +148,17 @@ def arguments_match?(job)
...
+            @queue == job[:queue]
           end
 
-          def serialized_at
-            @at == :no_wait ? nil : @at.to_f
+          def at_match?(job)
+            return true unless @at
+            return job[:at].nil? if @at == :no_wait
+
+            values_match?(@at, Time.at(job[:at]))
```
pirj: I imagine this will blow up with `TypeError: can't convert nil into an exact number` if we make an expectation of a job being scheduled in the future, but it's scheduled to be executed immediately.

Maybe a `job[:at] && values_match?(@at, Time.at(job[:at]))`?

JonRowe: Write a spec and find out? If it does add a `return false if job[:at].nil? && @at != nil`?

JonRowe: I'm assuming we're relying on `values_match?` here to expand the argument matcher?

pirj: Now it's back to the same behaviour as previously

pirj: > I'm assuming we're relying on values_match? here to expand the argument matcher?
Yep.

JonRowe: > Now it's back to the same behaviour as previously

No it's just defending against the `nil` condition...

pirj: What I mean by "back to the same behaviour" is that there's the same `expected to enqueue exactly 1 jobs, at .+ but enqueued 0` failure message in case the job is scheduled with no wait, and there's an expectation made with `at` qualifier.
Without this guard statement:
```ruby
return false unless job[:at]
```
it was erroring out with `TypeError` when trying to `Time.at` `nil` value that was in `job[:at]`.

However, I don't see that anything else needs to be done here, since `@at` at this moment is guaranteed to be neither `nil` (checked in the first guard statement) nor `:no_wait`, so it's either a date/time, either an argument matcher.

Do you have some case in mind that would misbehave?

JonRowe: Your guard condition is simplier :👍:

JonRowe:
Looks good, would prefer it squashed into one commit, and can you add a change log either before or after merging :)

pirj:
Changelog item added, squashed commits. Forgot to force-push, this pull request almost slipped my mind.

PR #2211, opened 2019-11-21
Support argument matchers in have_enqueued_*

### LINKED ISSUE #2205: approximate at for activejob

pirj:
Allows for:

```ruby
have_enqueued_job.at(a_value_within(5.seconds).of(future))
have_enqueued_mail.at(a_value_within(5.seconds).of(future))
```

Fixes #2205 for 3.9

`Time.at` seems to be quite fine as per https://github.com/rails/rails/blob/a707072/activejob/lib/active_job/test_helper.rb#L679

This is a backport of #2206 to 3-9-maintenance

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,14 @@
 ### Development
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.9.0...master)
 
+Enhancements:
+
+* Add argument matcher support to `have_enqueued_*` matchers. (Phil Pirozhkov, #2206)
+
 ### 3.9.0 / 2019-10-08
 [Full Changelog](http://github.com/rspec/rspec-rails/compare/v3.8.2...v3.9.0)
 
-Enhancements
+Enhancements:
 
 * Use `__dir__` instead of `__FILE__` in generated `rails_helper.rb` where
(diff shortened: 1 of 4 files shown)


JonRowe:
Hey @pirj our support policy for older versions of RSpec is generally not to back port enhancements, we will sometimes back port bug fixes if theres a good reason (across a major version at least) but as we prefer that the most recent up to date RSpec cater for as many people as possible and as all versions of supported Rails are supported by RSpec 4 I see no need to make an exception here.

pirj:
Thanks, Jon. That makes our lives easier 👍

PR #2212, opened 2019-11-22
Update rocket hash

hmtanbir:
Use latest rocket hash, removing old rocket hash from routing spec template.
Previous: #2198

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,7 +27,7 @@ FileName:
     # uses the gem name, things work without user configuration.
     - lib/rspec-rails.rb
 
-HandleExceptions:
+SuppressedException:
   Exclude:
     # RSpec is tightly coupled to capybara right now, this should be
     # re-evaluted in the future. For now we allow the empty rescue
@@ -44,13 +44,19 @@ PerlBackrefs:
     # we can't use named matches in 1.8.7
     - lib/generators/rspec/scaffold/scaffold_generator.rb
 
+HashSyntax:
+  Exclude:
+    # We probably can refactor the backref out, but for now excluding it since
+    # we can't use named matches in 1.8.7
+    - lib/generators/rspec/scaffold/scaffold_generator.rb
+
 Style/AccessModifierDeclarations:
   Enabled: false
 
 Naming/MemoizedInstanceVariableName:
   Enabled: false
 
-Naming/UncommunicativeMethodParamName:
+Naming/MethodParameterName:
   Enabled: false
 
 Metrics/PerceivedComplexity:
@@ -65,7 +71,7 @@ Lint/AssignmentInCondition:
 Lint/EmptyExpression:
   Enabled: false
 
-Layout/AlignHash:
+Layout/HashAlignment:
   Enabled: false
 
 Naming/RescuedExceptionsVariableName:
--- a/lib/generators/rspec/scaffold/scaffold_generator.rb
+++ b/lib/generators/rspec/scaffold/scaffold_generator.rb
@@ -7,17 +7,17 @@ module Rspec
     class ScaffoldGenerator < Base
       include ::Rails::Generators::ResourceHelpers
       source_paths << File.expand_path('../helper/templates', __dir__)
-      argument :attributes, :type => :array, :default => [], :banner => "field:type field:type"
+      argument :attributes, type: :array, default: [], banner: "field:type field:type"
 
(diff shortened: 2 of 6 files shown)


lib/generators/rspec/scaffold/templates/routing_spec.rb
```diff
@@ -5,41 +5,73 @@
   describe "routing" do
 <% unless options[:singleton] -%>
     it "routes to #index" do
-      expect(:get => "/<%= ns_table_name %>").to route_to("<%= ns_table_name %>#index")
+      if RUBY_VERSION > '1.9.3'
```
pirj: This check, and keeping the old syntax is unnecessary, edge [`rspec-rails` supports Rails 4.2+](https://github.com/rspec/rspec-rails/blob/master/rspec-rails.gemspec#L36), and they [only support Ruby 1.9.3+](https://github.com/rails/rails/blob/4-2-stable/rails.gemspec#L10).

lib/generators/rspec/scaffold/scaffold_generator.rb
```diff
@@ -7,17 +7,32 @@ module Generators
...
+        class_option :orm, desc: "ORM used to generate the controller"
+        class_option :template_engine, desc: "Template engine to generate view files"
+        class_option :singleton, type: :boolean, desc: "Supply to create a singleton controller"
+        class_option :api, type: :boolean, desc: "Skip specs unnecessary for API-only apps"
+
+        class_option :controller_specs, type: :boolean, default: true,  desc: "Generate controller specs"
+        class_option :view_specs,       type: :boolean, default: true,  desc: "Generate view specs"
+        class_option :helper_specs,     type: :boolean, default: true,  desc: "Generate helper specs"
+        class_option :routing_specs,    type: :boolean, default: true,  desc: "Generate routing specs"
+      else
```
pirj: I believe this branch can be dropped altogether.

hmtanbir: I'm wondering to see that when change has tag, the CI builds fails due to unknown hash subject.
I'm trying hard but could not understand why test is being failed.

pirj: Can you please point at a build with the failure you mention?

I can only see [RuboCop complaining](https://travis-ci.org/rspec/rspec-rails/jobs/615437211) at `Style/HashSyntax: Use hash rockets syntax`, feel free to update `.rubocop.yml`'s rule to prefer the new hash syntax and this offence will go away.

.rubocop.yml
```diff
@@ -44,13 +44,19 @@ PerlBackrefs:
...
+    # we can't use named matches in 1.8.7
+    - lib/generators/rspec/scaffold/scaffold_generator.rb
+
 Style/AccessModifierDeclarations:
   Enabled: false
 
 Naming/MemoizedInstanceVariableName:
   Enabled: false
 
-Naming/UncommunicativeMethodParamName:
```
pirj: Which version of RuboCop do you have bundled? I guess 0.77.

Version [0.74 was supposed to be pinned](https://github.com/rspec/rspec-rails/blob/master/Gemfile#L70) 
still has [`UncommunicativeMethodParamName`](https://github.com/rubocop-hq/rubocop/blob/96b090f1a2b04fca2bf57b370f461ee5f1a4bd91/lib/rubocop/cop/naming/uncommunicative_method_param_name.rb#L46) ([also](https://github.com/rubocop-hq/rubocop/blob/96b090f1a2b04fca2bf57b370f461ee5f1a4bd91/config/default.yml#L2038)) and does not yet have `MethodParamName`.

Do you mind to change:
```diff
- gem 'rubocop', "~> 0.74"
+ gem 'rubocop', '0.74'
```
or
```diff
+ gem 'rubocop', "~> 0.74.0"
```
to avoid RuboCop errors due to an older version being bundled for other contributors?

pirj: In this case, so many changes to `.rubocop*` files won't be required.

hmtanbir: Thanks @pirj , I will do update, thanks for reviewing.

pirj: I'll handle this in #2216

JonRowe:
Can you rebase this? We updated master to have the latest rubocop.

hmtanbir:
OK, I will rebase.

PR #2215, opened 2019-12-02
Fix `WrongScopeError` being thrown on Rails master:

Edouard-chin:
Fix `WrongScopeError` being thrown on Rails master:

- ### Problem

  Rails made a change in rails/rails@d4367eb72601f6e5bba3206443e9d141e9753af8
  and TestFixtures don't make use of `method_name` anymore, but simply
  `name`.

  `name` on RSpec raises a `WrongScopeError` when called within an
  example.

  This patch overrides `name` to return the example name instead.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -39,6 +39,12 @@ module RSpec
             end
           end
 
+          if ::Rails.version.to_f >= 6.1
+            def name
+              @example
+            end
+          end
+
           fixtures RSpec.configuration.global_fixtures if RSpec.configuration.global_fixtures
         end
       end
--- a/spec/rspec/rails/fixture_support_spec.rb
+++ b/spec/rspec/rails/fixture_support_spec.rb
@@ -13,5 +13,16 @@ module RSpec::Rails
         expect(group).to respond_to(:fixture_path=)
       end
     end
+
+    it "will allow #setup_fixture to run successfully", if: Rails.version.to_f > 6.0 do
+
+      group = RSpec::Core::ExampleGroup.describe do
+        include FixtureSupport
+
+        self.use_transactional_tests = false
+      end
+
+      expect { group.new.setup_fixtures }.to_not raise_error
+    end
   end
 end
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -196,7 +196,9 @@ RSpec.describe "Configuration" do
 
   it "metadata `type: :request` sets up request example groups" do
     a_rails_app = double("Rails application")
-    the_rails_module = Module.new
+    the_rails_module = Module.new {
+      def self.version; end;
+    }
     allow(the_rails_module).to receive(:application) { a_rails_app }
     version = ::Rails::VERSION
     stub_const "Rails", the_rails_module
@@ -230,7 +232,9 @@ RSpec.describe "Configuration" do
 
   it "metadata `type: :feature` sets up feature example groups" do
     a_rails_app = double("Rails application")
-    the_rails_module = Module.new
+    the_rails_module = Module.new {
+      def self.version; end;
+    }
     allow(the_rails_module).to receive(:application) { a_rails_app }
     version = ::Rails::VERSION
     stub_const "Rails", the_rails_module


jasnow:
Appears the fix triggered an existing rubocop error.

lib/rspec/rails/fixture_support.rb
```diff
@@ -50,6 +50,10 @@ def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
             end
           end
 
+          def name
```
JonRowe: If this is only needed on master, we should surround it with a version check, e.g:
```suggestion
          if Rails.version.to_f > 6
            def name
              @example
            end
          end
```

lib/rspec/rails/fixture_support.rb
```diff
@@ -50,6 +50,10 @@ def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
...
+            @example
+          end
```
JonRowe: If this is only needed on master, we should surround it with a version check, e.g:
```suggestion
          if Rails.version.to_f > 6
            def name
              @example
            end
          end
```

JonRowe: If this is only needed on master, we should surround it with a version check, e.g:
```suggestion
          if Rails.version.to_f > 6
            def name
              @example
            end
          end
```

Edouard-chin:
Thanks for the quick review. I made the suggested changes

lib/rspec/rails/fixture_support.rb
```diff
@@ -50,6 +50,12 @@ def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
             end
           end
 
+          if ::ActiveRecord.version >= Gem::Version.new('6.1.0')
```
JonRowe: Sorry you can't use `Gem::Version`, as we don't have a dependency on rubygems being available. A simple
```suggestion
          if Rails.version.to_f > 6.0
```
is preferred.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,15 @@ module RSpec::Rails
         expect(group).to respond_to(:fixture_path=)
       end
     end
+
+    it "doesn't raise a WrongScopeError on Rails 6.1" do
+      allow(ActiveRecord).to receive(:version) { Gem::Version.new('6.1.0') }
```
JonRowe: Proposed change removes this necessity

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,15 @@ module RSpec::Rails
...
+
+      group = RSpec::Core::ExampleGroup.describe do
+        include FixtureSupport
+      end
+
+      expect { group.new.name }.to_not raise_error
```
JonRowe: This doesn't really prove the issue, what code triggers the name usage?

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,15 @@ module RSpec::Rails
         expect(group).to respond_to(:fixture_path=)
       end
     end
+
+    it "doesn't raise a WrongScopeError on Rails 6.1" do
```
JonRowe: If the spec only needs to run on a version of Rails it should be flagged as such as we have CI running against multiple Rails and Ruby versions.

JonRowe:
If you rebase this the rubocop build should be fixed

jasnow:
@Edouard-chin - Looking forward to this fixed version.

lib/rspec/rails/fixture_support.rb
Edouard-chin: Rubygem is part of the standard library since ruby 1.9. I can use `Rails.version.to_f` if you really want to but I don't see any real reason

spec/rspec/rails/fixture_support_spec.rb
Edouard-chin: Issue that triggered the problem is when enrolling fixture transaction https://github.com/rails/rails/blob/b4ab1f19d8bcc4639c7379b560f000f01753b5b0/activerecord/lib/active_record/test_fixtures.rb#L101

This proves the issue as I just want to make sure `name` can be called without raising an error.
If you feel more confortable I can write an integration test and actually load fixtures, but this test should remain as I want to make sure `name` can be called within a spec

lib/rspec/rails/fixture_support.rb
JonRowe: We don't load any part of the standard library unless 100% necessary as it causes environment poisoning (where someone might not have required the right file but we do, tests pass, but it fails outside of tests). There is also a speed benefit to not using rubygems which we ourselves utilise in our tests by disabling it (which gives us the speed benefit and tests we're not poisoning the env).

JonRowe: e.g Its simpler to not use it so please don't :)

spec/rspec/rails/fixture_support_spec.rb
JonRowe: This should be demonstrated by loading a fixtures, its an implementation detail that `name` is available as we never call it.

lib/rspec/rails/fixture_support.rb
Edouard-chin: Ok I'll use `Rails.version.to_f`, thanks for quick answer :)

Not trying to argue on such a small detail but `Rails.version` already loads rubygem 🤷‍♂ . But anyway having the syntax you proposed will make things consistent since that's what it's used in this library.

lib/rspec/rails/fixture_support.rb
```diff
@@ -50,6 +50,12 @@ def self.proxy_method_warning_if_called_in_before_context_scope(method_name)
             end
           end
 
+          if ::Rails.version.to_f >= 6.1
```
benoittgt: I see the error on the CI. `Rails.version` is not good enough?

jasnow:
Rails::VERSION::STRING.to_f

lib/rspec/rails/fixture_support.rb
pirj: It seems that in this context `Rails` will be resolved to `RSpec::Rails`.

Edouard-chin: nop, it's just 2 tests stubbing the `Rails` module for I don't know what reason https://github.com/rspec/rspec-rails/pull/2215/files#diff-1279af43b24337b9b8a76fc3b3287f9aR199. Build should go green

Edouard-chin:
Not too sure what's up with the jruby builds, otherwise it's 🍏

pirj:
https://github.com/flori/json/issues/392

benoittgt:
Thanks @pirj. I just reran the 2 builds, the issue on json seems to be fixed now.

benoittgt:
@Edouard-chin thanks a lot for the last commit. Do you think you can handle @JonRowe review? 

Thanks :)

Edouard-chin:
> Do you think you can handle @JonRowe review?

Sure happy to, but which comment did I not address ?  If you are referring to https://github.com/rspec/rspec-rails/pull/2215#discussion_r355616564 I added a test to load fixtures

benoittgt:
Sorry @Edouard-chin. I just noticed. @JonRowe are you ok with the last changes?

Cheers :)

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,27 @@ module RSpec::Rails
...
+
+    it "doesn't raise a WrongScopeError" do
+      skip if Rails.version.to_f < 6.1
+
+      group = RSpec::Core::ExampleGroup.describe do
+        include FixtureSupport
+      end
+
+      expect { group.new.name }.to_not raise_error
+    end
```
JonRowe: This test can be removed, the other is fine.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,27 @@ module RSpec::Rails
...
+
+    it "setup_fixture successfuly" do
```
JonRowe:
```suggestion
    it "will allow #setup_fixture to run successfully" do
```

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,27 @@ module RSpec::Rails
...
+      skip if Rails.version.to_f < 6.1
```
JonRowe:
```suggestion
```

Edouard-chin:
Thanks for the review. I addressed all the comment. I also got rid of the test you mentioned, though I disagree that calling `name` is an implementation detail.

`name` is the standard way in Rails to retrieve the name of a test/spec. By adding a test where we explicitly call `name` we make sure that RSpec-rails is compatible with how Rails deal with test names.

JonRowe:
> Thanks for the review. I addressed all the comment. I also got rid of the test you mentioned, though I disagree that calling name is an implementation detail.

> `name` is the standard way in Rails to retrieve the name of a test/spec. By adding a test where we explicitly call `name` we make sure that RSpec-rails is compatible with how Rails deal with test names. Anyway just my cents, it's your project :)

What we care about is that this can be used with Rails, testing `name` does not guarantee that, it just guarantees we respond to `name` as we never call `name` ourself it is pointless. What we needed (and you have provided) is an integration test showing that we work with Rails whatever Rails calls, which happens to be `name` in `master` but previously has not been.

jasnow:
Any update on this PR?

JonRowe:
It needs to be rebased to go green

Edouard-chin:
Rebased and CI is 🍏 . Thanks for your patience

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -13,5 +13,16 @@ module RSpec::Rails
         expect(group).to respond_to(:fixture_path=)
       end
     end
+
+    it "will allow #setup_fixture to run successfully", if: Rails.version.to_f > 6.0 do
+
```
JonRowe:
```suggestion

```

Issue #2220, opened 2019-12-05
`rails g rspec:install` doesn't add `config.generators.test_framework` setting


rlue:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.7.0preview1
Rails version: 6.0.1
RSpec version: 3.9

## Observed behaviour

First...

```
$ rails new foo
# add rspec-rails to Gemfile
$ bundle
$ rails g rspec:install
```

then...

```
$ rails g scaffold posts
      ...
      invoke    test_unit
      create      test/controllers/posts_controller_test.rb
      create      test/system/posts_test.rb
      ...
```

## Expected behaviour

```
$ rails g scaffold posts
      ...
      invoke    rspec
      create      spec/controllers/messages_controller_spec.rb
      create      spec/views/messages/edit.html.erb_spec.rb
      create      spec/views/messages/index.html.erb_spec.rb
      create      spec/views/messages/new.html.erb_spec.rb
      create      spec/views/messages/show.html.erb_spec.rb
      create      spec/routing/messages_routing_spec.rb
      ...
```

Obviously this is achieved by adding the following line to the Rails application config:

```
config.generators.test_framework :rspec
```

but I am curious—is does the RSpec team consider this too much hand-holding? In my opinion, if it's not going to be provided automatically by the `rspec:install` generator, then it should at least be in the README.

pirj:
I believe it's [here](https://github.com/rspec/rspec-rails/blob/be3ae66f25863e05336d7fb54b363a833179aa4c/lib/rspec-rails.rb#L21).

Can you reproduce this with `rspec-rails` version `4.0.0.beta3`?

rlue:
Wow, that was fast!

It appears that in order to take advantage of this, I need to `require 'rspec-rails'` in `/config/application.rb`. Is this the correct approach? Is there another way to do it?

pirj:
As far as I remember, putting this in `Gemfile` is sufficient:
```ruby
group :development, :test do
  gem 'rspec-rails', '~> 4.0' # in case of pre-release '4.0.0.beta3'
end
```

rlue:
My goodness, you're right. I think when I tried it yesterday, maybe I hadn't run `spring stop`, so the generators were running based on the outdated application config? Thanks again for your insight.

rlue:
Okay, one thing I just noticed is that the default generators still create controller specs, even though they are discouraged as of RSpec 3.5. 

[...]

PR #2222, opened 2019-12-07
Favor request specs over controller specs when generating a controller

### LINKED ISSUE #2056: Generate request spec by default when generating controller

Ruby version: 2.4.1
Rails version: 5.2.2
Rspec version: 3.8
```sh
- rspec-core 3.8.0
- rspec-expectations 3.8.2
- rspec-mocks 3.8.0
- rspec-rails 3.8.1
- rspec-support 3.8.0
```

## Observed behavior
When generating Rails controller, it will automatically generate corresponding controller spec.

## Expected behavior
Generate corresponding request spec.

> The official recommendation of the Rails team and the RSpec core team is to write request specs instead.

Is it possible for the user to configure RSpec-rails so that it generates request spec by default when generating controller test?

klyonrad:
Closes #2056

Open questions in my mind (and why I consider this a work in progress):

1. Should I take care of the scaffolding generators as well?
2. Should the request spec templates have more content?

EDIT:

Regarding 2: Yes, they should
Regarding 1: I created a separate issue for this. See #2230


lib/generators/rspec/controller/templates/request_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'rails_helper'
+
+RSpec.describe "<%= class_name.pluralize %>", <%= type_metatag(:request) %> do
```
klyonrad: Copied over from `lib/generators/rspec/integration/templates/request_spec.rb`

I didn't try "very hard" to use the same template file because I am not sure if that is even wished for

JonRowe: I'd recommend writing a spec for get name etc that returns success or something similarly basic.

spec/generators/rspec/controller/controller_generator_spec.rb
```diff
@@ -5,23 +5,29 @@
...
-        run_generator %w(posts)
-      end
+  describe 'request specs' do
+    subject { file('spec/requests/posts_request_spec.rb') }
 
-      describe 'the spec' do
-        it { is_expected.to exist }
-        it { is_expected.to contain(/require 'rails_helper'/) }
-        it { is_expected.to contain(/^RSpec.describe PostsController, #{type_metatag(:controller)}/) }
+    if Rails.version.to_f >= 5.0
```
JonRowe: We're dropping official support for versions below 5 so you can exclude this in favour of the else

pirj:
Up for discussion, should we omit generation of controller specs altogether? Brought up [here](https://github.com/rspec/rspec-rails/issues/2220#issuecomment-562911237).
```ruby
generators.test_framework :rspec, controller_specs: false
```
[Related code](https://github.com/rspec/rspec-rails/blob/be3ae66f25863e05336d7fb54b363a833179aa4c/lib/rspec-rails.rb#L21).

JonRowe:
> Up for discussion, should we omit generation of controller specs altogether? Brought up here.

No changing the default is sufficient

spec/generators/rspec/controller/controller_generator_spec.rb
klyonrad: I was guided by [your comment from the issue](https://github.com/rspec/rspec-rails/issues/2056#issuecomment-450561088) 😉

> I'd support a PR that made the behaviour configurable, and changed the default the behaviour of the generator for Rails 5.1, (maybe 5) but older Rails expect controller specs

Anyway, the simpler the better :)

lib/generators/rspec/controller/templates/request_spec.rb
klyonrad: Yeah, figuring the details for the templates out was/is a bit difficult for me. I am sure, I would figure it out - but can you clarify the specs a bit maybe? I mean, what does the rspec-rails team think should be the convention for request specs that test a controller

benoittgt: Does the request spec guide help you? https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec

JonRowe: > I mean, what does the rspec-rails team think should be the convention for request specs that test a controller

I don't use generators myself as I find that the similar lines tend to just be the naming only. There isn't that much of a semantic difference between a controller spec and a request spec, its the level of stack integration that differs. Personally I would be happy with:

```
describe "GET /<%= name.underscore.pluralize %>" do
  it "works! (now write some real specs)" do
    get "/<%= name.underscore.pluralize %>"
    expect(response).to be_success
  end
end
```

However finesse wise I think that `name.underscore.pluralize` is wrong here, because it needs to be name spaced (or not) appropriately.

JonRowe: Let me know if you want any more feedback

benoittgt:
For the issue on master branch. I will try to fix it with. https://github.com/rspec/rspec-rails/pull/2231

JonRowe:
Can you rebase this? Master is fixed.

JonRowe:
Thanks for your patience here!

PR #2223, opened 2019-12-07
Make version check work if version is not from ENV

mvz:
This fixes an issue where setting the Rails version in the `.rails-version` file would cause an "undefined method \`gsub' for nil:NilClass" error, because `gsub` was called on the `nil` value of the `RAILS_VERSION` env var.


JonRowe:
Thanks!

PR #2225, opened 2019-12-08
[experimental] Drop support of Ruby < 2.2 & Rails < 5.2

pirj:
As per [Rails' support policy](https://guides.rubyonrails.org/maintenance_policy.html), Rails dropped all support for 5.0/5.1, even for severe security issues.

[This comment](https://github.com/rspec/rspec-rails/pull/2071#issuecomment-466551497) mentioned that we planned to keep support for 5.0.
[Another comment](https://github.com/rspec/rspec-rails/pull/2071#issuecomment-469073857) mentioning plans to drop 4.2 support.

From [Versioning Strategy RFC](https://github.com/rspec/rspec-rails/blame/master/rfcs/versioning-strategy.md#L60):

> Release a major with any Rails Major, removing support for any newly unsupported versions of Rails, and adding support for the new major.

This pull request has some rough edges, but I can iron it out if you feel like dropping support for < 5.2 is the way to go.


pirj:
The only failure is YARD on JRuby mistakenly takes `{lib}` in
```ruby
       gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => 'master'
```
for a reference.

I'm getting a weird `nokogiri` issue locally, running `rake` fails to find 1.8.5. Update to `nokogiri` 1.9.1 (that requires Ruby 2.1+) fixes the issue.

pirj:
Questions/TODO:

Is it possible to get rid of [this](https://github.com/rspec/rspec-rails/blob/master/example_app_generator/generate_action_mailer_specs.rb#L5)?
```ruby
# We need to copy this method from Thor for older Rails versions
def comment_lines(path, flag, *args)
```

Evaluate numerous `respond_to?` calls, e.g.:
```ruby
          ::ActiveRecord::Migration.respond_to?(:check_pending!)
```
```ruby
I18n.enforce_available_locales = true if I18n.respond_to?(:enforce_available_locales)
```

Simplify `FixtureFileUploadSupport` that has some trick to support Rails 3.

JonRowe:
I seem to recall the original discussion was to drop support for 4.x in 4, but not 5.x, leaving that for later, I know thats not what the RFC says, but basically kicking out all of the previously supported rails versions at once is a bit much.

JonRowe:
See #2071 for reference

PR #2226, opened 2019-12-10
Drop code related to Rails < 5.0 & Ruby < 2.3.0

pirj:
Another cleanup attempt (former: #2225).

Note: in the [Changelog, we state that we drop support for Ruby versions below 2.3](https://github.com/rspec/rspec-rails/blame/afa6382067969f93effed4d1bd2aeb1f433fc536/Changelog.md#L20), even though Rails 5.2 supports Ruby 2.2.2. It's a matter of getting just a couple lines back to keep 2.2.2 support.

I didn't dare to refactor [`FixtureFileUploadSupport`](https://github.com/rspec/rspec-rails/blob/7870656722c84f7651d9e4b4dc66caaf9142c8c4/lib/rspec/rails/fixture_file_upload_support.rb#L27).


pirj:
I can reproduce `yard doc --no-cache` failure locally:
```
[warn]: ...{lib}...
[warn]: In file `README.md':17: Cannot resolve link to lib from text:
```

It seems that JRuby's doc generation is awfully broken:
image

For the reference, same generated on `master` and MRI 2.5.5:
image

Still not pixel perfect (list numbering is off, code block layout) match to [what's published](http://rspec.info/documentation/3.9/rspec-rails/file.README.html):
image

Maybe it's acceptable to disable doc checks on JRuby?

Gemfile
```diff
@@ -43,7 +43,7 @@ else
 end
 
 if RUBY_VERSION >= '2.4.0'
-  gem 'json', '>= 2.0.2'
+  gem 'json', '>= 2.0.2', '< 2.3.0'
```
pirj: Temporary, due to https://github.com/flori/json/issues/392

pirj:
@JonRowe It's a big one, but mostly removals of old Rubies/Rails 3/4 support. Appreciate if you glance over.

benoittgt:
For Rails 4.2 did you read this comment https://github.com/rspec/rspec-rails/pull/2071#issuecomment-466551497
I am wondering how much we should clean code related to Rails 4.2

Thanks for the work.

JonRowe:
Pretty much all of this was good to go @pirj I'd reopen it as is mostly.

> At 4.0 release time, 4.2 will be removed from the build matrix, but not the gem spec

This means we can remove all of our spec / feature tests for Rails 4.2

> 4.2 will not be officially supported in RSpec Rails 4, and bugs against that version will not be fixed

This means that we should leave the checks for 4.2 methods to check they are supported before calling in lib, and the same for files. However anything that is old Ruby is ok to be dropped. We're hard supporting 5.0 and soft supporting 4.2 but not old Ruby, thats gone.

> A warning will be issued at require time, if a user is using Rails 4.2, indicating to them that while the version is unlikely to be supported, bug fixes will not be issued against 4.2

This needs to be added somewhere.

benoittgt:
Sorry @pirj I was maby not clear enough. But I don't this PR should be closed. :)

JonRowe:
@pirj let me know when this is ready to review again :)

pirj:
@JonRowe @benoittgt Please take another look.

I broke four examples on 4.2, `mailer_generator_spec.rb` and `configuration_spec.rb`, by removing `< 5.0` checks in those specs. The code should work fine. If you comment them out
```ruby
RAILS_VERSION=4.2.11 rake
```
expectedly get stuck at the point where `@rails_post_5` tags were.

example_app_generator/generate_app.rb
```diff
@@ -63,21 +58,7 @@
...
-    |  gem 'rake', '< 12.0.0' # rake 12 requires Ruby 2.0.0 or later
-    |else
-    |  gem 'rake', '< 11.0.0' # rake 11 requires Ruby 1.9.3 or later
-    |end
-    |
-    |# Version 3 of mime-types 3 requires Ruby 2.0
-    |if RUBY_VERSION < '2.0.0'
-    |  gem 'mime-types', '< 3'
-    |end
+    |gem 'rake', '>= 10.0.0'
```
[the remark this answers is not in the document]
pirj: Latest 13.0.1 depends on Ruby >= 2.2. Did I understand you correctly?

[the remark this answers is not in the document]
JonRowe: Yeah thats fine

pirj:
@JonRowe @benoittgt Green!

script/predicate_functions.sh
```diff
@@ -73,22 +69,14 @@ function rspec_rails_compatible {
...
-
 function additional_specs_available {
   type run_additional_specs > /dev/null 2>&1
   return $?
 }
 
 function documentation_enforced {
   if [ -x ./bin/yard ]; then
-    if is_mri_2plus; then
+    if is_mri; then
```
JonRowe: This file is managed by rspec-dev, I'd revert these unless you have a good reason?

JonRowe:
One change really, only due to our process, currently the dev build should apply cleanly to rspec-rails and I'd like to keep it that way for centrally managed files

script/predicate_functions.sh
pirj: Sure, makes total sense 👍 
I haven't thought those will be overwritten with the next update from rspec-dev anyway.

pirj:
All green. Reverted changes to predicate functions. Squashed commits.

PR #2227, opened 2019-12-10
Fix Rails version env var format in Travis config

pirj:
[`downgrade_bundler_on_old_rails`](https://github.com/rspec/rspec-rails/blob/d401812/script/downgrade_bundler_on_old_rails#L8) has a check for Rails < 5, and it false triggered for builds where RAILS_VERSION had two digits.

Example builds: [before](https://travis-ci.org/rspec/rspec-rails/jobs/620832476), [after](https://travis-ci.org/rspec/rspec-rails/jobs/623433099).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,13 +41,13 @@ matrix:
     - rvm: jruby-head
       jdk: oraclejdk11
       env:
-        - RAILS_VERSION='~>6.0'
+        - RAILS_VERSION='~> 6.0.0'
         - JRUBY_OPT=--dev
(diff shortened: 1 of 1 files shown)


JonRowe:
👍 Whilst this does have different semantics, I'm ok with putting this back to a full version.

pirj:
Yep, I though that this will miss 6.0 when 6.1 is out, but pretty sure we'll ajdust the config by that time.

JonRowe:
We'd have to add 6.0 anyway, so its no harm

PR #2228, opened 2019-12-11
ActiveJob performed matchers

pirj:
* `have_been_performed` is like `have_been_enqueued`, but for performed
  jobs
* `have_performed_job` is like `have_enqueued_job`, but for performed
  jobs

It's somewhat unusual to use `perform_enqueued_jobs`, but it seems nice if 
rspec-rails supports it!

Original work by @isaacseymour in #1785

Closes #1785


pirj:
Haven't realized push to pr branch worked.

PR #2233, opened 2019-12-16
Fix RUBYOPT usages

pirj:
There's a warning regarding an unknown argument to `--disable`, and it seems that `--disable=gem` is not entirely correct, should be `--disable=gems`.
[Documentation](https://docs.ruby-lang.org/ja/latest/doc/spec=2frubycmd.html) (in Japanese).

[Example output](https://travis-ci.org/rspec/rspec-rails/jobs/625710994):
```
ruby: warning: unknown argument for --disable: `gem:'
ruby: warning: features are [gems, did_you_mean, rubyopt, frozen_string_literal].
ruby: warning: unknown argument for --enable: `rubygems'
ruby: warning: features are [gems, did_you_mean, rubyopt, frozen_string_literal].
```

### DIFF
--- a/example_app_generator/travis_retry_bundle_install.sh
+++ b/example_app_generator/travis_retry_bundle_install.sh
@@ -5,7 +5,7 @@ source FUNCTIONS_SCRIPT_FILE
 
 echo "Starting bundle install using shared bundle path"
 if is_mri_192_plus; then
-  travis_retry eval "RUBYOPT=$RUBYOPT:' --enable rubygems' bundle install --gemfile ./Gemfile --path REPLACE_BUNDLE_PATH --retry=3 --jobs=3"
+  travis_retry eval "RUBYOPT=$RUBYOPT' --enable gems' bundle install --gemfile ./Gemfile --path REPLACE_BUNDLE_PATH --retry=3 --jobs=3"
 else
   travis_retry eval "bundle install --gemfile ./Gemfile --path REPLACE_BUNDLE_PATH --retry=3 --jobs=3"
 fi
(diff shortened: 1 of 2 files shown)


benoittgt:
Don't think all of those changes should be done on rspec-dev?

JonRowe:
No I don't think these are necessary, unsetting rubyopts would have no affect if we were wrong about the rubyopts option we used. So this might be updated on later rubies but must be still valid...

pirj:
It seems that Ruby is tolerable, and both `gem`/`gems` can be used. Using `gem:` was a mistake though.

JonRowe:
The `:` is the joiner for environment variables... `$A = $A:"more"` is the bash equivalent of `a += "
more"`

pirj:
It worked differently for some reason:
```sh
$ A='1'
$ B=$A:' 2'
$ echo $B
1: 2
```

JonRowe:
My bad, I think it's actually just a $PATH thing

benoittgt:
@pirj do you think you could rebase from master?

pirj:
Done, rebased.

benoittgt:
Thanks. This should normally help us for the next step https://github.com/bundler/bundler/pull/7505

I am happy to merge this to rspec-dev (https://github.com/rspec/rspec-rails/pull/2239)

JonRowe:
If we don't think this will change anything on the build, (I don't, but I'm willing to be convinced) I'd like to hold off making any changes with this until we have a green build.

pirj:
Also green.

JonRowe:
Does this make a difference? I'm inclined to leave it be otherwise?

pirj:
Not a big deal, just two warnings less in the build. Can be addressed later.

PR #2239, opened 2019-12-18
Updates from rspec-dev (2019-12-18)

JonRowe:


JonRowe:
Closed because this needs tweaking to make it not conflict

JonRowe:
@benoittgt @pirj I think this stands a good chance of going green (its late here so I'm off), its the bare minimum with @benoittgt's rubygems unset and @mvz's bundler patch for Aruba, plus ignoring 4.2.0 (because of bundler 1.x required and we're dropping support anyway) and ignoring a warning on jruby only. 🤞

features/support/env.rb
```diff
@@ -6,9 +6,30 @@ def run_command(cmd, timeout = nil)
     exec_cmd = cmd =~ /^rspec/ ? "bin/#{cmd}" : cmd
     # Ensure bundler env vars are unset
     unset_bundler_env_vars
-    # Ensure the correct Gemfile is found
+    # Ensure the correct Gemfile and binstubs are found
     in_current_directory do
-      super(exec_cmd, timeout)
+      with_unbundled_env do
```
mvz: I _think_ this call to `with_unbundled_env` is not needed.

JonRowe:
I merged #2244 in preference as it does less.

features/support/env.rb
JonRowe: Yep, #2244 did not need it.

PR #2244, opened 2019-12-20
Update travis build scripts 2019 12 20 for master

JonRowe:
@benoittgt @pirj this seems to be the minimum required changes to get our build green so I'm proposing we use this...


Gemfile
```diff
@@ -42,21 +42,12 @@ else
   gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
 end
 
-if RUBY_VERSION >= '2.4.0'
-  gem 'json', '>= 2.0.2'
-end
```
JonRowe: Technically these made no difference for this build.

spec/sanity_check_spec.rb
```diff
@@ -34,7 +34,7 @@ def with_clean_env
     end
   end
 
-  it "passes when libraries are required" do
+  it "passes when libraries are required", unless: RSpec::Support::Ruby.jruby? do
```
benoittgt: Do you think we should revert that change when gem/json-jruby warning will be fixed?

benoittgt:
We have many open PR for the same thing (https://github.com/rspec/rspec-rails/pull/2239). I am happy to merge this one.

spec/sanity_check_spec.rb
JonRowe: Yes probably.

PR #2246, opened 2019-12-22
Updated travis build scripts (from rspec-dev)

pirj:
Removed `yes |` gem trick, as new RubyGems version is out.

Drop `rspec-support` support check.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2019-12-18T14:01:39+00:00 from the rspec-dev repo.
+# This file was generated on 2019-12-22T22:08:33+03:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 version: "{build}"
@@ -33,6 +33,9 @@ test_script:
 
 environment:
   matrix:
+    - ruby_version: 200
+    - ruby_version: 21
+    - ruby_version: 22
     - ruby_version: 23-x64
     - ruby_version: 24-x64
     - ruby_version: 25-x64
--- a/script/functions.sh
+++ b/script/functions.sh
@@ -1,4 +1,4 @@
-# This file was generated on 2019-12-18T14:01:39+00:00 from the rspec-dev repo.
+# This file was generated on 2019-12-22T22:08:33+03:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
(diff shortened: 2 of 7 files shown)


JonRowe:
Not compatible

pirj:
It seems it is compatible, those jobs that were run before branch deletion succeeded.

appveyor.yml
```diff
@@ -33,6 +33,9 @@ test_script:
 
 environment:
   matrix:
+    - ruby_version: 200
+    - ruby_version: 21
+    - ruby_version: 22
```
JonRowe: These should not be added.

script/functions.sh
```diff
@@ -13,7 +13,7 @@ MAINTENANCE_BRANCH=`cat maintenance-branch`
 
 # Don't allow rubygems to pollute what's loaded. Also, things boot faster
 # without the extra load time of rubygems. Only works on MRI Ruby 1.9+
-if is_mri; then
+if is_mri_192_plus; then
```
JonRowe: In theory this is ok, but makes no sense for rspec-rails as its always 192 plus.

script/functions.sh
```diff
@@ -47,7 +47,12 @@ function run_cukes {
 
     echo "${PWD}/bin/cucumber"
 
-    if is_jruby; then
+    if is_mri_192; then
+      # For some reason we get SystemStackError on 1.9.2 when using
+      # the bin/cucumber approach below. That approach is faster
+      # (as it avoids the bundler tax), so we use it on rubies where we can.
+      bundle exec cucumber --strict
+    elif is_jruby; then
```
JonRowe: This has been added negating the `else` branch that is actually what we want.

script/functions.sh
```diff
@@ -158,8 +163,6 @@ function check_documentation_coverage {
     end
   "
 
-  echo "bin/yard doc --no-cache"
-
```
JonRowe: Removes the doc check?

script/functions.sh
```diff
@@ -184,6 +187,9 @@ function run_all_spec_suites {
   fold "rspec-core specs" run_spec_suite_for "rspec-core"
   fold "rspec-expectations specs" run_spec_suite_for "rspec-expectations"
   fold "rspec-mocks specs" run_spec_suite_for "rspec-mocks"
-  fold "rspec-rails specs" run_spec_suite_for "rspec-rails"
+  if rspec_rails_compatible; then
+    fold "rspec-rails specs" run_spec_suite_for "rspec-rails"
+  fi
+
```
JonRowe: Again this is technically ok but makes no sense.

script/update_rubygems_and_install_bundler
```diff
@@ -1,9 +1,15 @@
...
-yes | gem update --system
-yes | gem install bundler
+if is_ruby_23_plus; then
+  gem update --system
+  gem install bundler
+else
+  echo "Warning installing older versions of Rubygems / Bundler"
+  gem update --system '2.7.8'
+  gem install bundler -v '1.17.3'
+fi
```
JonRowe: We're always ruby 2.3 plus.

PR #2256, opened 2020-01-08
Fix Mocha compatibility

### LINKED ISSUE #2252: v4.0.0.beta3 regression when using mocha

Ruby version: 2.6.5p114
Rails version: 6.0.2.1
RSpec version: 4.0.0.beta2

The application in question uses the mocha gem together with rspec-rails

## Observed behaviour

WIth 4.0.0.beta3 if I use something like this:
`Node.any_instance.stubs(:compiled_parameters).raises ParameterConflictError`
I get the following error:
`The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported`

## Expected behaviour

With 4.0.0.beta2 the same tests run flawlessly.

I bisected the issue and git pointed to 7ff7c3e291fcce6a4706bc45314ab2a50f72ffa9 as the first problematic commit.

The only added code outside of tests is `require 'rspec/rails/matchers/have_enqueued_mail'` which in turn does `require "rspec/mocks"`. This leads to the test now jumping into `rspec::mocks` (https://github.com/rspec/rspec-mocks/blob/master/lib/rspec/mocks/syntax.rb#L78) instead of using the method from the mocha gem.

pirj:
RSpec Mocks happened to load its configuration, use the default syntax and define `any_instance` on `Class` before Mocha had a chance to in case `rspec/rails` was required in `spec/spec_helper.rb` before `rspec/core` and configuring Mocha as `mock_with`.

This is not something typical, as usually `rails_helper.rb` requires `rspec-rails`, and `spec_helper.rb` requires `rspec-core` and configures `mock_with`, but not something that is a complete no-no.

fixes #2252

- [x] add changelog

### DIFF
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -1,4 +1,7 @@
-require "rspec/mocks"
+# We require the minimum amount of rspec-mocks possible to avoid
+# conflicts with other mocking frameworks.
+# See: https://github.com/rspec/rspec-rails/issues/2252
+require "rspec/mocks/argument_matchers"
 require "rspec/rails/matchers/active_job"
 
 module RSpec
@@ -11,7 +14,7 @@ module RSpec
       class HaveEnqueuedMail < ActiveJob::HaveEnqueuedJob
         MAILER_JOB_METHOD = 'deliver_now'.freeze
 
-        include RSpec::Mocks::ExampleMethods
+        include RSpec::Mocks::ArgumentMatchers
 
         def initialize(mailer_class, method_name)
           super(nil)
--- a/Changelog.md
+++ b/Changelog.md
@@ -38,6 +38,7 @@ Bug Fixes:
   (Takumi Shotoku, #2188)
 * Prevent a `WrongScopeError` being thrown during loading fixtures on Rails
   6.1 development version. (Edouard Chin, #2215)
+* Fix Mocha mocking support with `should`. (Phil Pirozhkov, #2256)
 
 Breaking Changes:
 


pirj:
@ZeroPointEnergy highly appreciate if you test this fix out with `puppet-dashboard`.

benoittgt:
Thanks for the patch. Do you think we can test it in rspec-rails?

Waiting for @ZeroPointEnergy feedback.

pirj:
> Do you think we can test it in rspec-rails?
I'm open to ideas how to do that.

ZeroPointEnergy:
@pirj is it possible that the PR is not on the 4.x development branch? It threw some dependency errors when I tried to use the branch directly.

However I simply cherry-picked ac886dc8db3cb6bbf50805a0cc4c56bcc1ce6742 on top of v4.0.0.beta3 and can confirm that the patch indeed fixes the issue in puppet-dashboard! :+1:

Thanks a lot for your work.

JonRowe:
Honestly I'm ok with just adding a comment on top explaining why the minimalist approach is taken. To add a test you'd have to assert a bunch of stuff about poisoning the environment which is a fair amount of effort to setup...

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -1,4 +1,4 @@
-require "rspec/mocks"
+require "rspec/mocks/argument_matchers"
```
JonRowe:
```suggestion
# We require the minimum amount of rspec-mocks possible to avoid conflicts with over mocking frameworks.
# See: https://github.com/rspec/rspec-rails/issues/2252
require "rspec/mocks/argument_matchers"
```

pirj:
Looks green.

pirj:
Rushed it a bit. @benoittgt do you have any objections on merging as is? I can do a follow-up in that case.

benoittgt:
No no it's perfect! Thanks a lot 😊

JonRowe:
Released as `v.4.0.0.beta4`

PR #2258, opened 2020-01-09
Drop some legacy code

pirj:
We don't support Ruby 1.x anymore, so those tricks are no longer necessary.

### DIFF
--- a/features/controller_specs/anonymous_controller.feature
+++ b/features/controller_specs/anonymous_controller.feature
@@ -51,7 +51,7 @@ Feature: anonymous controller
       RSpec.describe ApplicationController, :type => :controller do
         controller do
           def index
-            raise ApplicationController::AccessDenied
+            raise AccessDenied
           end
         end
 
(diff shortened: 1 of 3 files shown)


features/controller_specs/anonymous_controller.feature
```diff
@@ -51,7 +51,7 @@ Feature: anonymous controller
       RSpec.describe ApplicationController, :type => :controller do
         controller do
           def index
-            raise ApplicationController::AccessDenied
+            raise AccessDenied
```
JonRowe: Fairly certain this is just a scoping issue not a 1.x thing?

pirj: Indeed. Decided to keep things as - it's better to be on the safe side.

JonRowe: Are you keeping this branch or can it be deleted?

pirj: Deleted 👍

PR #2259, opened 2020-01-10
fix docs matcher links

mkamensky:
Fix links to matchers in the job specs description


benoittgt:
Tested and it's good to me:
Capture d’écran 2020-01-10 à 21 01 24

Maybe we could switch to a list with `-` to make it more readable but that's not a blocker

PR #2261, opened 2020-01-11
Get rid of redundant requires

pirj:
We already tell RSpec to require `spec/spec_helper.rb` in `.rspec` with `--require spec_helper`.

### DIFF
--- a/spec/rspec/rails/assertion_adapter_spec.rb
+++ b/spec/rspec/rails/assertion_adapter_spec.rb
@@ -1,5 +1,3 @@
-require "spec_helper"
-
 describe RSpec::Rails::MinitestAssertionAdapter do
   include RSpec::Rails::MinitestAssertionAdapter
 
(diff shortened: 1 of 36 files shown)


features/directory_structure.feature
```diff
@@ -151,8 +151,6 @@ Feature: Directory Structure
   Scenario: Non-rails related specs do not require `:type` metadata by default
     Given a file named "spec/ledger/entry_spec.rb" with:
       """ruby
-      require "spec_helper"
```
JonRowe: I'd prefer leaving this as it makes the documentation more explicit.

JonRowe:
Also I noticed we have several spec files with bare `describe` top level statements, fancy changing those to `RSpec.describe` in another PR if you're cleaning things up?

pirj:
> bare describe top level statements, fancy changing those to RSpec.describe in another PR if you're cleaning things up?
Absolutely 👍 Noticed that as well.

Issue #2043, opened 2018-11-27
Headers instead of sessions on scaffolding for API apps


exocode:
I have a question regarding to scaffolding on API only apps. 

Each time I scaffold an spec (controllers for example) it writes

`session: valid_session`

instead of 

`headers: valid_headers`

```
  describe "GET #show" do
    it "returns a success response" do
      user = User.create! valid_attributes
      get :show, params: {id: user.to_param}, session: valid_session
      expect(response).to be_successful
    end
  end
```

Wouldn't it be better to scaffold with `headers` instead of `sessions`?
The most APIs are using JWT or other tokens to authenticate.
Aren't Sessions mostly obsolete in APIs?

```
  describe "GET #show" do
    it "returns a success response" do
      user = User.create! valid_attributes
      get :show, params: {id: user.to_param}, headers: valid_headers
      expect(response).to be_successful
    end
  end
```

JonRowe:
The scaffolding was based on the old controller specs, it simply wasn't something considered. Only hesitation is that switching would have to consider the difference in how headers are handled across rails versions, this would need to be tested as working on all versions of Rails and not just as "output" is correct.

benoittgt:
@exocode Do you want to work on a PR to validate the behavior on Rails versions supported by rspec-rails?

exocode:
@benoittgt I would give it a try :-)

klyonrad:
I was wondering about this when I worked on my PR. I am not sure if I understand the initial problem of this issue correctly, but in my opinion this is less of an issue now because since https://github.com/rspec/rspec-rails/pull/2288 scaffolding with the `--api` option does actually generate **request** specs that generate a `valid_headers` method.

JonRowe:
Closing since I think this is solved now.

PR #2262, opened 2020-01-11
Generate request specs when scaffold generator is used

### LINKED ISSUE #2230: Generate request spec by default when generating scaffolds
Similar problem like in #2056. Upon using `rails generate scaffold` it would be good if controller tests would not get generated and instead request specs are generated.

klyonrad:
Closes #2230

Ouf, this is quite some work... 😁 

## TODO

- [x] Make the generated specs actually work
- [x] `show` and `/edit` action
- [x] Implement differences for `--api` flag
- [x] Fix the [double request spec problem](https://github.com/rspec/rspec-rails/issues/2230#issuecomment-573320813)
- [ ] Add appendix with example results into PR body
- [x] Consistent generation of paths/urls?

# Open questions

## Integration / smoke test coverage

As far as I understand the smoke tests were testing the controller specs because they were generated by default. So making them not the default reduces the test coverage a bit. Should we take a go at explicitly generating the controller specs in the smoke app too?

## Title of the generated request spec

Top title of the generated specs is debatable (See also [Comment](https://github.com/rspec/rspec-rails/pull/2262#discussion_r365566595))

References

- The [documentation](https://relishapp.com/rspec/rspec-rails/v/3-9/docs/request-specs/request-spec) mentions something "ResourceName management"
- Corresponding generated file from rails project (https://github.com/rails/rails/blob/5572df92be8c529909277ee8e4ac5b9da26b893a/railties/lib/rails/generators/test_unit/scaffold/templates/functional_test.rb.tt#L4)
- Rails guide on the matter https://guides.rubyonrails.org/testing.html#functional-tests-for-your-controllers

## Valid_session

The controller specs generated a `:valid_session` method. I don't know what the equivalent for request specs should be called. The same thing?

## valid_headers

Should I take a go at fixing the issue from https://github.com/rspec/rspec-rails/issues/2043 for scaffolding?

## Naming of the describe blocks

The controller scaffold specs were grouped with `PATCH #update` etc. The current draft has `PATCH /update`. This does not fit so well for request specs anymore. Method documentation style  `#update` does not fit because we use actual urls and don't call methods, `/update` does not fit because that is not really the URL.

We could write `PATCH /index_url` and `GET /resource_name/:id`. This fits as long as users stick to the simple rails conventions

# Appendix

TODO


lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      it "creates a new <%= class_name %>" do
+        expect {
+          post <%= "#{index_helper}_path" %>, params: { <%= ns_file_name %>: valid_attributes }
+        }.to change(<%= class_name %>, :count).by(1)
+      end
+
+      it "redirects to the created <%= ns_file_name %>" do
+        expect {
+          post <%= "#{index_helper}_path" %>, params: { <%= ns_file_name %>: valid_attributes }
+        }.to redirect_to("<%= ns_table_name %>/#{<%= class_name %>.last}")
```
klyonrad: Not so sure which style should be the convention. Should the specs use the paths for the URLs (at this point routing specs become a bit pointless) or use the rails helpers?

lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+# It assumes that the implementation code is generated by the rails scaffold
+# generator. If you are using any extension libraries to generate different
+# controller code, this generated spec may or may not pass.
+#
+# It only uses APIs available in rails and/or rspec-rails. There are a number
+# of tools you can use to make these specs even more expressive, but we're
+# sticking to rails and rspec-rails APIs to keep things simple and stable.
+
+<% module_namespacing do -%>
+RSpec.describe <%= controller_class_name %>Controller, <%= type_metatag(:request) %> do
```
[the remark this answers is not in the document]
JonRowe: Request specs don't test controllers, they test specific requests. They route to a certain path and something happens, it could be the controller you intend, it might not. It could be terminated in a rack middleware, it could be routed to a different controller.

JonRowe:
Ping, how are you getting on with this?

klyonrad:
> Ping, how are you getting on with this?

Still have in this in the back on my mind, but didn't have time for it this week

lib/generators/rspec/scaffold/templates/request_spec.rb
klyonrad: I got guided by `rails/generators/test_unit/scaffold/templates/functional_test.rb.tt`

klyonrad:
@JonRowe Tests should be green now, the API request spec template is still missing but please have a look already, since I will copy a lot from the current template from the draft now

example_app_generator/generate_stuff.rb
```diff
@@ -137,8 +137,8 @@ def using_source_path(path)
           'config.warnings = false'
 gsub_file '.rspec', '--warnings', ''
 
-# Remove skips so we can test controller specs work
-gsub_file 'spec/controllers/gadgets_controller_spec.rb',
+# Remove skips so we can test that request specs work
+gsub_file 'spec/requests/gadgets_request_spec.rb',
```
benoittgt: Could you explain why we have to change that?

example_app_generator/run_specs.rb
```diff
@@ -7,7 +7,6 @@
 run('bin/rake --backtrace spec:requests') || abort
 run('bin/rake --backtrace spec:models') || abort
 run('bin/rake --backtrace spec:views') || abort
-run('bin/rake --backtrace spec:controllers') || abort
```
benoittgt: Could you explain why we have to change that?

example_app_generator/generate_stuff.rb
klyonrad: The file that was attempted to edit does not exist anymore. That's what I meant with my question in the PR body about *Integration / smoke test coverage*

example_app_generator/run_specs.rb
klyonrad: I suspected that the smoke app does not generate controller specs anymore because the default of the ScaffoldGenerator changed

JonRowe: Rather than removing these, you need to add a command to generate some controller tests for this to run.

example_app_generator/generate_stuff.rb
JonRowe: You need to ensure that the smoke coverage covers both scenarios.

.rubocop.yml
```diff
@@ -104,6 +104,10 @@ Metrics/BlockLength:
 # Warns when the class is excessively long.
 Metrics/ClassLength:
   Max: 100
+  Exclude:
+    # Splitting the Generator classes would make matters more confusing
+    # than them being too big
+    - lib/generators/rspec/**/*
```
JonRowe: Add the relevant magic comment to the generator that fails rather than doing this, or add the specific files, its possible to extract things without class splitting.

example_app_generator/generate_stuff.rb
klyonrad: Done.

Also added generation with using the `--api` flag

example_app_generator/run_specs.rb
klyonrad: Done.

example_app_generator/generate_stuff.rb
```diff
@@ -89,10 +89,12 @@ def using_source_path(path)
 
 generate('model thing name:string')
 generate('helper things')
-generate('scaffold widget name:string category:string instock:boolean foo_id:integer bar_id:integer --force')
+generate('scaffold product name:string category:string instock:boolean foo_id:integer bar_id:integer --force')
```
klyonrad: At this point the project stops being fun 😠 

Apparently there other tests assumed that the `Widget` constant is created and I cannot figure how to make the testing of IntegrationGenerator and RequestGenerator work, sometimes the `<%= index_helper %>_path` works, sometimes or the `_url` helper.

I still think that generated request specs is important and very useful but for working on this **as a lone wolf I am out**. So please help, take over oder schedule chatting/pair programming (personally; I am more available in February).

---

Log of the smoke tests:

```
Failures:

  1) Using rspec-mocks with models supports stubbing attribute methods on models
     Failure/Error: a_widget = Widget.new
     
     NameError:
       uninitialized constant Widget
     # ./spec/features/model_mocks_integration_spec.rb:10:in `block (2 levels) in <top (required)>'

  2) Using rspec-mocks with models supports stubbing class methods on models
     Failure/Error: allow(Widget).to receive(:all).and_return(:any_stub)
     
     NameError:
       uninitialized constant Widget
     # ./spec/features/model_mocks_integration_spec.rb:5:in `block (2 levels) in <top (required)>'

  3) Widgets GET /widgets works! (now write some real specs)
     Failure/Error: get widgets_index_path
     
     NameError:
       undefined local variable or method `widgets_index_path' for #<RSpec::ExampleGroups::Widgets::GETWidgets:0x00007ff1f27043d8>
     # ./spec/requests/widgets_spec.rb:6:in `block (3 levels) in <top (required)>'
```

lib/generators/rspec/scaffold/scaffold_generator.rb
```diff
@@ -60,8 +78,6 @@ def generate_routing_spec
         template 'routing_spec.rb', template_file
       end
 
-      hook_for :integration_tool, as: :integration
```
klyonrad: (Check also my commit message on this change)

I had to dig extremely dig into the git history of the project, tried to figure the purpose and the consequences of this setting (the generator from `railties` does not have this). But I don't understand it. But is the thing that causes the `IntegrationGenerator` to run and also generate a `request_spec` file (figured that out through debugging and lookig at the backtrace).

Changing this seems to cause *something*, at least that's what the test suite is telling me

See also 98b886b5528aa2a60c42d80b83a0973f24bfcdb7

example_app_generator/generate_stuff.rb
pirj: Have you had a chance to take another look? I can lend a hand.

klyonrad: Yeah, I'll write you an email to chat up or something

JonRowe:
Thanks for your hard work getting this started @klyonrad after playing whack-a-mole on the tests I got this working as #2288, closing this in favour of #2288

PR #2264, opened 2020-01-12
Streamline RSpec config

pirj:
Disable monkey patching and fix specs related to this.

### DIFF
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -60,12 +60,11 @@ module RSpec::Rails
       let(:controller) { double('controller') }
       let(:example) { group.new }
       let(:routes) do
-        routes = nil
         with_isolated_stderr do
           routes = ActionDispatch::Routing::RouteSet.new
           routes.draw { resources :foos }
+          routes
         end
-        routes
       end
 
       before do
@@ -84,7 +83,6 @@ module RSpec::Rails
 
       it "calls NamedRouteCollection#route_defined? when it checks that given route is defined or not" do
         expect(routes.named_routes).to receive(:route_defined?).and_return(true)
-        expect(routes.named_routes).not_to receive(:helpers)
 
         example.foos_url
       end
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -3,7 +3,7 @@ class ::ApplicationController
 end
 
 module RSpec::Rails
-  describe ControllerExampleGroup do
+  RSpec.describe ControllerExampleGroup do
     it_behaves_like "an rspec-rails example group mixin", :controller,
       './spec/controllers/', '.\\spec\\controllers\\'
 
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -56,7 +56,7 @@ module RSpec::Rails
         end
 
         it 'includes the application helper' do
-          group = RSpec::Core::Example.describe 'bars/new.html.erb'
+          group = RSpec::Core::ExampleGroup.describe 'bars/new.html.erb'
           expect(group).to receive(:helper).with(ApplicationHelper)
           group.class_exec do
             include ViewExampleGroup
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -8,7 +8,9 @@ module RSpec::Rails
       module ::Namespaced; module ThingsHelper; end; end
 
       it 'includes the helper with the same name' do
-        group = RSpec::Core::ExampleGroup.describe 'things/show.html.erb'
+        group = RSpec::Core::ExampleGroup.describe('things/show.html.erb') do
+          def self.helper(*); end # Stub method
+        end
         expect(group).to receive(:helper).with(ThingsHelper)
         group.class_exec do
           include ViewExampleGroup
@@ -16,7 +18,9 @@ module RSpec::Rails
       end
 
       it 'includes the namespaced helper with the same name' do
-        group = RSpec::Core::ExampleGroup.describe 'namespaced/things/show.html.erb'
+        group = RSpec::Core::ExampleGroup.describe('namespaced/things/show.html.erb') do
+          def self.helper(*); end # Stub method
+        end
         expect(group).to receive(:helper).with(Namespaced::ThingsHelper)
         group.class_exec do
           include ViewExampleGroup
@@ -56,7 +60,9 @@ module RSpec::Rails
         end
 
         it 'includes the application helper' do
-          group = RSpec::Core::ExampleGroup.describe 'bars/new.html.erb'
+          group = RSpec::Core::ExampleGroup.describe('bars/new.html.erb') do
+            def self.helper(*); end # Stub method
+          end
           expect(group).to receive(:helper).with(ApplicationHelper)
           group.class_exec do
             include ViewExampleGroup
@@ -124,6 +130,7 @@ module RSpec::Rails
             end
           end
           include local
+          def _default_file_to_render; end # Stub method
           include ViewExampleGroup::ExampleMethods
         end.new
       end
@@ -174,7 +181,7 @@ module RSpec::Rails
         Class.new do
           include ViewExampleGroup::ExampleMethods
           def controller
-            @controller ||= Object.new
+            @controller ||= OpenStruct.new(params: nil)
           end
         end.new
       end
@@ -189,8 +196,10 @@ module RSpec::Rails
       let(:view_spec) do
         Class.new do
           include ViewExampleGroup::ExampleMethods
+          def _default_file_to_render; end
         end.new
       end
+
       context "with a common _default_file_to_render" do
         it "it returns the directory" do
           allow(view_spec).to receive(:_default_file_to_render).
@@ -213,6 +222,7 @@ module RSpec::Rails
     describe "#view" do
       let(:view_spec) do
         Class.new do
+          def _view; end # Stub method
           include ViewExampleGroup::ExampleMethods
         end.new
       end
@@ -227,13 +237,13 @@ module RSpec::Rails
         with_isolated_config do
           run_count = 0
           RSpec.configuration.before(:each, :type => :view) do
-            allow(view).to receive(:a_stubbed_helper) { :value }
+            allow(view).to receive(:render) { :value }
             run_count += 1
           end
           group = RSpec::Core::ExampleGroup.describe 'a view', :type => :view do
             specify { true }
           end
-          group.run NullObject.new
+          group.run
           expect(run_count).to eq 1
         end
       end
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -1,5 +1,5 @@
 module RSpec::Rails
-  describe ViewExampleGroup do
+  RSpec.describe ViewExampleGroup do
     it_behaves_like "an rspec-rails example group mixin", :view,
       './spec/views/', '.\\spec\\views\\'
 
--- a/spec/rspec/rails/matchers/have_rendered_spec.rb
+++ b/spec/rspec/rails/matchers/have_rendered_spec.rb
@@ -4,6 +4,7 @@
     let(:response) { ActionDispatch::TestResponse.new }
 
     context "given a hash" do
+      def assert_template(*); end
       it "delegates to assert_template" do
         expect(self).to receive(:assert_template).with({:this => "hash"}, "this message")
         expect("response").to send(template_expectation, {:this => "hash"}, "this message")
@@ -11,6 +12,7 @@
     end
 
     context "given a string" do
+      def assert_template(*); end
       it "delegates to assert_template" do
         expect(self).to receive(:assert_template).with("this string", "this message")
         expect("response").to send(template_expectation, "this string", "this message")
@@ -18,6 +20,7 @@
     end
 
     context "given a symbol" do
+      def assert_template(*); end
       it "converts to_s and delegates to assert_template" do
         expect(self).to receive(:assert_template).with("template_name", "this message")
         expect("response").to send(template_expectation, :template_name, "this message")
@@ -26,8 +29,8 @@
 
     context "with should" do
       context "when assert_template passes" do
+        def assert_template(*); end
         it "passes" do
-          def assert_template(*); end
           expect do
             expect(response).to send(template_expectation, "template_name")
           end.to_not raise_exception
--- a/spec/rspec/rails/matchers/have_rendered_spec.rb
+++ b/spec/rspec/rails/matchers/have_rendered_spec.rb
@@ -1,5 +1,5 @@
 %w[have_rendered render_template].each do |template_expectation|
-  describe template_expectation do
+  RSpec.describe template_expectation do
     include RSpec::Rails::Matchers::RenderTemplate
     let(:response) { ActionDispatch::TestResponse.new }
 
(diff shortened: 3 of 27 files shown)


spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -58,7 +58,7 @@ module ::ApplicationHelper; end
         end
 
         it 'includes the application helper' do
-          group = RSpec::Core::Example.describe 'bars/new.html.erb'
+          group = RSpec::Core::ExampleGroup.describe 'bars/new.html.erb'
```
pirj: This is interesting.

spec/spec_helper.rb
```diff
@@ -25,17 +26,34 @@ def self.run_all(reporter=nil)
...
 end
 
+# See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
+  config.expect_with :rspec do |expectations|
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
+  end
+
+  config.mock_with :rspec do |mocks|
+    # mocks.verify_partial_doubles = true
```
pirj: A lot of specs fail. Not sure if it's worth it considering we're only using `instance_double` once.

pirj: Complete nonsense ☝️

pirj: A lot of specs fail, e.g.:
```
  17) RSpec::Rails::SetupAndTeardownAdapter::setup registers teardown hooks in the order setup is received
      Failure/Error: expect(klass).to receive(:after).ordered { |&block| expect(block.call).to eq "foo" }
        #<Class:0x00007fdb8f30ec70> does not implement: after
      # ./spec/rspec/rails/setup_and_teardown_adapter_spec.rb:40:in `block (3 levels) in <top (required)>'

  18) render_template given a symbol converts to_s and delegates to assert_template
      Failure/Error: expect(self).to receive(:assert_template).with("template_name", "this message")
        #<RSpec::ExampleGroups::RenderTemplate::GivenASymbol "converts to_s and delegates to assert_template" (./spec/rspec/rails/matchers/have_rendered_spec.rb:23)> does not implement: assert_template
```

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: I mean... I'm not sure theres a difference :joy:

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -19,7 +19,7 @@ def to_global_id(options = {})
   end
 end
 
-RSpec.describe "have_broadcasted_to matchers", :skip => !RSpec::Rails::FeatureCheck.has_action_cable_testing? do
+RSpec.describe "have_broadcasted_to matchers", :if => RSpec::Rails::FeatureCheck.has_action_cable_testing? do
```
JonRowe: This behaves differently, `if` means it won't have the `skipped` message?

spec/spec_helper.rb
JonRowe: There is a method that can surpress verification for specific spec's that need it.

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -84,7 +83,6 @@ def my_helper
 
       it "calls NamedRouteCollection#route_defined? when it checks that given route is defined or not" do
         expect(routes.named_routes).to receive(:route_defined?).and_return(true)
-        expect(routes.named_routes).not_to receive(:helpers)
```
pirj: Frankly, I'm not sure what it has to do with helpers.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -225,6 +235,7 @@ def controller
 
       it 'is accessible to hooks' do
         with_isolated_config do
+          view = double("view")
```
pirj: No idea why it was passing before.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -124,6 +130,7 @@ def _assigns
             end
           end
           include local
```
pirj: Quite surprised, but defining those methods direcly in the class breaks the spec.

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: Check the original history with blame, I imagine this is a regression check of some kind

spec/rspec/rails/matchers/have_rendered_spec.rb
```diff
@@ -1,8 +1,10 @@
 %w[have_rendered render_template].each do |template_expectation|
-  describe template_expectation do
+  RSpec.describe template_expectation do
     include RSpec::Rails::Matchers::RenderTemplate
     let(:response) { ActionDispatch::TestResponse.new }
 
+    def assert_template(*); end # Stub method
```
JonRowe: Why did you move this?

pirj: Not only that example that I've moved it from need `assert_template` to be defined on `self`.
I was on the fence of keeping it there for clarity. Should I?

spec/rspec/rails/example/controller_example_group_spec.rb
pirj: This needs some deeper research on how it works with 4.2 and later.
I'll comment this out instead and add a todo, there might be a potential to get rid of some legacy in `controller_example_group.rb`.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: I'm not sure what this refers to

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -189,8 +196,10 @@ def controller
       let(:view_spec) do
         Class.new do
           include ViewExampleGroup::ExampleMethods
+          def _default_file_to_render; end
```
JonRowe: What needs this?

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: Is `view` not also a method?

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -213,6 +222,7 @@ def controller
     describe "#view" do
       let(:view_spec) do
         Class.new do
+          def _view; end # Stub method
```
JonRowe: Whats this for?

spec/rspec/rails/example/view_example_group_spec.rb
pirj: I've tried to:
```ruby
Class.new do
  def received
    @received ||= []
  end
  # ...
```
but this doesn't work.

spec/rspec/rails/example/view_example_group_spec.rb
pirj: [This](https://github.com/rspec/rspec-rails/pull/2264/files/87aee32e014aed4c818156a9c473c5de8682b8cc#diff-0cdf276129b0b83c6314283584c4df6eR205) and [this](https://github.com/rspec/rspec-rails/pull/2264/files/87aee32e014aed4c818156a9c473c5de8682b8cc#diff-0cdf276129b0b83c6314283584c4df6eR214).

spec/rspec/rails/example/view_example_group_spec.rb
pirj: It's to satisfy [this](https://github.com/rspec/rspec-rails/pull/2264/files/87aee32e014aed4c818156a9c473c5de8682b8cc#diff-0cdf276129b0b83c6314283584c4df6eR232).

spec/rspec/rails/example/view_example_group_spec.rb
pirj: Doesn't seem so, at least not it the bare context of `ExampleGroup.describe`.
This is needed [here](https://github.com/rspec/rspec-rails/pull/2264/files/87aee32e014aed4c818156a9c473c5de8682b8cc#diff-0cdf276129b0b83c6314283584c4df6eR241).
Picked it up [here](https://github.com/rspec/rspec-rails/pull/2264/files/87aee32e014aed4c818156a9c473c5de8682b8cc#diff-0cdf276129b0b83c6314283584c4df6eR231).
I suspect that `with_isolated_config` was swallowing `NoMethodError` previously.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: Doesn't work as in?

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: I'd not comment it out, but feel free to leave a todo to investigate.

spec/rspec/rails/matchers/have_rendered_spec.rb
JonRowe: It makes it available to more specs, which is not necessarily desired.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: Aside, can you quote code as well as link, Github makes it really hard to navigate with just links...

JonRowe: Ok I see it now, it's just a bit weird because its not in the same context.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: This link doesn't seem to work

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: Nope nor these links

JonRowe: `with_isolated_config` does not swallow errors, it just runs an `ensure`

JonRowe: It's defined here: `lib/rspec/rails/example/view_example_group.rb:81` so this shouldn't be removed...

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -225,6 +235,7 @@ def controller
 
       it 'is accessible to hooks' do
         with_isolated_config do
+          view = double("view")
           run_count = 0
           RSpec.configuration.before(:each, :type => :view) do
             allow(view).to receive(:a_stubbed_helper) { :value }
```
JonRowe: This runs in a `ViewExampleGroup` where `view` is a method.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: No I meant how doesn't it work, how does:
```
Class.new do
  def received
    @received ||= []
  end
end
```
Not have the same effect as above? What is the error it causes.

pirj: It's this:
```
  4) RSpec::Rails::ViewExampleGroup#render given no input sends render(:template => (described file)) to the view
     Failure/Error: expect(view_spec.received.first).to eq([{:template => "widgets/new"},{}, nil])

       expected: [{:template=>"widgets/new"}, {}, nil]
            got: [{}, {}, nil]

       (compared using ==)

       Diff:
       @@ -1,2 +1,2 @@
       -[{:template=>"widgets/new"}, {}, nil]
       +[{}, {}, nil]
```

It does not make sense to me, I don't see any `prepend` coming into play.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -213,6 +222,7 @@ def controller
     describe "#view" do
       let(:view_spec) do
```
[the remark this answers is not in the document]
JonRowe: Nah

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: You need to keep `with_isolated_config`, if your changes here are causing an error you need to ensure the right methods are defined, doing `double` here is just wrong..

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: It must be something similar, as this exists as is I suggest leaving it.

spec/rspec/rails/example/view_example_group_spec.rb
pirj: A call to `allow(view).to receive(:a_stubbed_helper) { :value }` results in:
```
RSpec::Mocks::MockExpectationError: #<#<Class:0x00007fa59165da18>:0x00007fa59165ceb0 @_config={}, ...> does not implement: a_stubbed_helper
```
that for some reason doesn't bubble up.
The same happens if you move stubbing directly to the example:
```ruby
          group = RSpec::Core::ExampleGroup.describe 'a view', :type => :view do
            specify {
              allow(view).to receive(:a_stubbed_helper) { :value }
              run_count += 1
            }
          end
```
Example returns prematurely, `run_count` is not incremented, but there's no exception from mocks. Is it a bug?

Changed to stub an existing `render` method instead.

Removed `NullObject.new` from `group.run`'s arguments, it doesn't seem to affect this in any way.

pirj:
Rebase fixed, all green.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -227,13 +237,13 @@ def controller
...
           RSpec.configuration.before(:each, :type => :view) do
-            allow(view).to receive(:a_stubbed_helper) { :value }
+            allow(view).to receive(:render) { :value }
             run_count += 1
           end
           group = RSpec::Core::ExampleGroup.describe 'a view', :type => :view do
             specify { true }
           end
-          group.run NullObject.new
+          group.run
```
JonRowe: Doesn't this cause the groups output to be output to the real formatter?

pirj: There's no difference in the output.

JonRowe: I was surprised by that assertion, as the point of this object is to prevent the real reporter being used, but its failing locally for me? I want to investigate this further as its not possible to prove either way on CI without changing the format...

What I expect this to do is to add the spec you've specified above into the reporter, so if you run this spec on its own you'd see "2 examples passed" for this one example.

pirj: If I run it focused (`rspec spec/rspec/rails/example/view_example_group_spec.rb:236`), it fails locally. However, the spec when run altogether passes.
This weird behavior remains even with just one example left, and if you remove `module` namespace, and `with_isolated_`, and nesting contexts. I've put it on my list to investigate.

pirj: Maybe it doesn't add to the main reporter because of https://github.com/rspec/rspec-rails/pull/2264/files#diff-93830fa29d616f7c87903d08b5b1b29aR55 ?
```ruby
    real_world = RSpec.world
    example.run
  end
```

pirj: I'm not sure if `with_isolated_config` is even necessary in specs with that hook.

JonRowe: So I got to the bottom of why this doesn't run when focused. If you're interested its an artefact in our filtering, `with_isolated_config` does not remove filters from this spec, as they exist in the world and the world is not isolated. That is something I might consider fixing at some point but is waaaaaaaaay out of scope for this.

I also confirmed this removal is ok, we now use a `NullReporter` by default, a change which post-dates this spec I suspect.

PR #2266, opened 2020-01-21
Add configuration to surpress active_record checks

### LINKED ISSUE #1876: Add a global switch for ActiveRecord

JonRowe:
Alternative to #2038 fixes #1876. 

I'm currently missing a reliable test for this, the approach in #2038 is one I do not want to take, we've never configured a database and our specs have worked, so if this switch is required when no database is present there has to be a way to trigger that error without configuring a database?

Backtrace [from stackoverflow](https://stackoverflow.com/questions/46204197/testing-a-databaseless-rails-5-application-with-rspec-rails):
```
 ActiveRecord::ConnectionNotEstablished:
   No connection pool with id primary found.
 # activerecord-5.0.1/lib/active_record/connection_adapters/abstract/connection_pool.rb:882:in `retrieve_connection'
 # activerecord-5.0.1/lib/active_record/connection_handling.rb:128:in `retrieve_connection'
 # activerecord-5.0.1/lib/active_record/connection_handling.rb:91:in `connection'
 # activerecord-5.0.1/lib/active_record/fixtures.rb:516:in `create_fixtures'
 # activerecord-5.0.1/lib/active_record/fixtures.rb:1015:in `load_fixtures'
 # activerecord-5.0.1/lib/active_record/fixtures.rb:988:in `setup_fixtures'
 # activerecord-5.0.1/lib/active_record/fixtures.rb:852:in `before_setup'
 # rspec-rails-3.6.1/lib/rspec/rails/adapters.rb:126:in `block (2 levels) in <module:MinitestLifecycleAdapter>'
 # rspec-core-3.6.0/lib/rspec/core/example.rb:447:in `instance_exec'
```


lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -42,6 +42,13 @@
   # instead of true.
   config.use_transactional_fixtures = true
 
+  # You can uncomment this line to turn off active_record support entirely.
+  # config.use_active_record = false
+
+<% else -%>
+  # Remove this line to enable support for active_record
```
pirj: WDYT of mentioning `config.use_transactional_fixtures` and `config.fixture_path` here to get full power out of ActiveRecord?

JonRowe: Yeah maybe, mostly just thinking aloud with this, it fails a spec which I'm not sure is this PR or would happen for both...

lib/rspec/rails/fixture_support.rb
```diff
@@ -10,6 +10,10 @@ module FixtureSupport
         include ActiveRecord::TestFixtures
 
         included do
+          setup if RSpec.configuration.use_active_record?
```
pirj: This was calling another `setup`, specificly the one from `lib/rspec/rails/adapters.rb`.
The following:
```
included do
  next unless RSpec.configuration.use_active_record?

  self.fixture_path = RSpec.configuration.fixture_path
  # the rest of setup
```

works just fine.

pirj:
@JonRowe this seems trivial and mostly working, I can take it from here if you like.

lib/generators/rspec/install/templates/spec/rails_helper.rb
JonRowe: Edited.

JonRowe:
@pirj Happy for you to take it from here, I might have more time to work on it later.

pirj:
Sounds good 👍

lib/rspec/rails/fixture_support.rb
```diff
@@ -44,8 +56,6 @@ def name
               @example
```
pirj: @Edouard-chin Do you mind to approve if I didn't break [your fix](https://github.com/rspec/rspec-rails/pull/2215)?

spec/generators/rspec/install/install_generator_spec.rb
```diff
@@ -59,19 +67,24 @@ def filter_rails_from_backtrace
...
+    specify "with ActiveRecord" do
       run_generator
-      expect(rails_helper).to have_a_fixture_path
+      expect(rails_helper).to have_active_record_enabled
+      expect(rails_helper).not_to have_active_record_disabled
     end
 
     specify "with default fixture path" do
       run_generator
-      expect(rails_helper).to have_transactional_fixtures_enabled?
```
pirj: Those two were mixed up.

pirj:
What tests did you have in mind to cover this change, @JonRowe ?

JonRowe:
> It's so tempting to remove this.

I mean I guess we could, but maybe in `rspec-rails` 5 is better, and we deprecate it for now...

> What tests did you have in mind to cover this change, @JonRowe ?

We need to disable the db and check that it still works

lib/rspec/rails/fixture_support.rb
Edouard-chin: Looks good to me 👍 . Thanks for the ping

benoittgt:
> I mean I guess we could, but maybe in rspec-rails 5 is better, and we deprecate it for now...

+1

JonRowe:
This passes the build now, for tests I think the best step is to add something to the `no_active_record` smoke app that tests that nothing happens with fixtures when this is turned off

pirj:
That's what I'm on right now. Trying to reproduce the original failure from [this app](https://github.com/JoostVanAverbeke/devise_remote_client/tree/usage_of_rspec-rails_3_3_2) to mimic it properly in the test.

example_app_generator/generate_stuff.rb
```diff
@@ -44,6 +44,7 @@ def setup_tasks
 
     def final_tasks
       copy_file 'spec/verify_no_active_record_spec.rb'
+      copy_file 'spec/verify_no_fixture_setup_spec.rb'
```
JonRowe: ?

example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+# so that it's reachable from RSpec::Rails::FixtureSupport.
+# NOTE: this has to be defined before requiring `rails_helper`.
+module RSpec
+  module Rails
+    module ActiveRecord
+      module TestFixtures
+      end
+    end
+  end
+end
```
pirj: It's like that since the original problem was that `active_record` was defined, just wasn't used and DB wasn't set up.

I've double-checked that this test does blow up if `config.use_active_record = true` inside `no_active_record` app.

pirj: I tried hard, but couldn't make [this example repo](https://github.com/JoostVanAverbeke/devise_remote_client/tree/usage_of_rspec-rails_3_3_2) work to reproduce the original issue. Mostly due to bundler 1/2 problem, and then dependencies hell with Rails version which I had to update to 5.2, and in the end with missing `rake-test`'s methods, probably due to version mismatch.

example_app_generator/generate_stuff.rb
pirj: Added another spec to the smoke app. Should it be done differently?

JonRowe: Sorry no, I read this as being identical several times, must be getting later :joy:

pirj: I was confused as well when I've finished typing its name 😆

JonRowe:
Fixes #1876

JonRowe:
@pirj I rewrote the commits for this and re-pushed, once green I'm going to merge if you're happy?

pirj:
@Jack12816 @mrgordon @JoostVanAverbeke does this work for you?

mrgordon:
Thanks!

PR #2267, opened 2020-01-22
Use latest Rails by default

pirj:
Latest Rails includes ActionCable tests, that are otherwise skipped when testing the compatibility of other RSpec projects, e.g. `rspec-expectation` on CI.

Related pull request that was expected to fail, but didn't https://github.com/rspec/rspec-expectations/pull/1157

Originally found here https://github.com/rspec/rspec-expectations/pull/1155#issuecomment-573457208

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,7 @@ when /stable/
   MAJOR = 6
   MINOR = 0
 when nil, false, ""
-  MAJOR = 5
+  MAJOR = 6
   MINOR = 0
 else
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -28,7 +28,7 @@ when /stable/
   MAJOR = 6
   MINOR = 0
 when nil, false, ""
```
JonRowe: Slight improvement, you could merge this with above.

pirj:
This can be merged after https://github.com/rspec/rspec-dev/pull/235 and its child pull requests.

Issue #2276, opened 2020-02-09
4.0.0beta4 reports RSpec::Core::ExampleGroup::WrongScopeError Exception: `default_url_options` is not available on an example group


samstickland:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.6.4
Rails version: 6.0.2.1
RSpec version: 4.0.0.beta4

## Observed behaviour

rspec-rails is unable to set `default_url_options[:host]` to  `::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST` as `default_url_options` is not defined.

Here's the output from byebug inserted just before the exception is raised:

```
[727, 736] in /Users/samuelstickland/.gem/ruby/2.6.4/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb
   727:
   728:       def self.method_missing(name, *args)
   729:         if method_defined?(name)
   730:           debugger
   731:
=> 732:           raise WrongScopeError,
   733:                 "`#{name}` is not available on an example group (e.g. a " \
   734:                 "`describe` or `context` block). It is only available from " \
   735:                 "within individual examples (e.g. `it` blocks) or from " \
   736:                 "constructs that run in the scope of an example (e.g. " \
(byebug) name
:default_url_options
(byebug) up

[13, 22] in /Users/samuelstickland/.gem/ruby/2.6.4/gems/rspec-rails-4.0.0.beta4/lib/rspec/rails/example/feature_example_group.rb
   13:         app = ::Rails.application
   14:         if app.respond_to?(:routes)
   15:           include app.routes.url_helpers     if app.routes.respond_to?(:url_helpers)
   16:           include app.routes.mounted_helpers if app.routes.respond_to?(:mounted_helpers)
   17:
=> 18:           default_url_options[:host] ||= ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
   19:         end
   20:       end
   21:
   22:       # Shim to check for presence of Capybara. Will delegate if present, raise
(byebug) default_url_options
*** RSpec::Core::ExampleGroup::WrongScopeError Exception: `default_url_options` is not available on an example group (e.g. a `describe` or `context` block). It is only available from within individual examples (e.g. `it` blocks) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc).

nil
(byebug) ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
"www.example.com"
```

## Expected behaviour

[...]

PR #2277, opened 2020-02-10
Restore respond_to? check for default_url_options

eugeneius:
Fixes https://github.com/rspec/rspec-rails/issues/2276.
Reverts part of https://github.com/rspec/rspec-rails/pull/2226.

Feature example groups usually have a `default_url_options` class attribute, but it's not added if the method is already defined:

https://github.com/rails/rails/blob/v6.0.2.1/actionpack/lib/action_dispatch/routing/url_for.rb#L92

Request example groups have a `default_url_options` instance method:

https://github.com/rails/rails/blob/v6.0.2.1/actionpack/lib/action_dispatch/testing/integration.rb#L388

This means that when a feature example group is defined inside a request example group, it will only have the instance method, and trying to call `default_url_options` on the example group will fail.

### DIFF
--- a/lib/rspec/rails/example/feature_example_group.rb
+++ b/lib/rspec/rails/example/feature_example_group.rb
@@ -15,7 +15,9 @@ module RSpec
           include app.routes.url_helpers     if app.routes.respond_to?(:url_helpers)
           include app.routes.mounted_helpers if app.routes.respond_to?(:mounted_helpers)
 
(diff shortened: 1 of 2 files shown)


JonRowe:
I've merged this because I think its the easiest solution, we could probably find a better way of setting his, I ruled out `app` directly as I think it will leak but for now we'll get this into the next beta.

nbulaj:
@eugeneius thanks for the fix, I hope it's also would be in 4.0 release because beta4 breaks CI pipeline for me

PR #2281, opened 2020-03-04
Relax maximum capybara version constraint

### LINKED ISSUE #1095: Hard bump the capybara version
If we still have a dependency on capybara, we should hard bump the version on 4.0. Additionally, reach out to the capybara team early on this process to better coordinate any changes that need to be made, and sync the releases between the two gems.

Related to: #1091, #1094

pirj:
fixes #1095

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -48,7 +48,7 @@ gem 'rake', '~> 12'
 
 gem 'mime-types', "~> 3"
 
-gem 'capybara', '~> 2.13', require: false
+gem 'capybara', '>= 2.13', '< 4.0', require: false
 
 if MAJOR < 6
   gem 'nokogiri', '1.9.1'
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ you can run the specs and Cucumber features, or submit a pull request.
 ### Recommended third-party extensions
 
 * [FactoryBot](https://github.com/thoughtbot/factory_bot)
-* [Capybara](https://github.com/jnicklas/capybara)
+* [Capybara](https://github.com/teamcapybara/capybara)
   (Included by default in Rails 5.1+.
   Note that [additional configuration is required][] to use the Capybara DSL
   anywhere other than system specs and feature specs.)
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -48,7 +48,7 @@ gem 'rake', '~> 12'
 
 gem 'mime-types', "~> 3"
 
-gem 'capybara', '~> 2.13', require: false
+gem 'capybara', '>= 2.13', '< 4.0', require: false
```
pirj: Checked locally, `rake` is green with the latest Capybara 3.x.

Changes:
1. Capybara 3 changes the default server to Puma ([doc](https://github.com/teamcapybara/capybara/blob/master/UPGRADING.md#server))
This is pretty much up to the users, `rspec-rails` don't depend on either.

2. Finders ([doc](https://github.com/teamcapybara/capybara/blob/master/UPGRADING.md#finders))
Up to the users as well, we don't wrap `first`/`all`.

3. Node ([doc](https://github.com/teamcapybara/capybara/blob/master/UPGRADING.md#node))
We don't wrap those methods, so it's up to the users.

Things might go wrong for users that would `bundle update rspec-rails` without the `--conservative` flag, they would have to deal with those update issues. However, this will only happen when Rails hits 6.1 (or when Rails will decide to bump Capybara to 3.x), and Rails' 6.0 -> 6.1 dos should pretty much cover it.

README.md
```diff
@@ -366,7 +366,7 @@ you can run the specs and Cucumber features, or submit a pull request.
 ### Recommended third-party extensions
 
 * [FactoryBot](https://github.com/thoughtbot/factory_bot)
-* [Capybara](https://github.com/jnicklas/capybara)
+* [Capybara](https://github.com/teamcapybara/capybara)
```
JonRowe: 👍

Gemfile
JonRowe: I'd still prefer `'~> 2.13', '~> 3.0'` its semantically the same but a bit clearer in my opinion. Unless its not valid for some reason?

pirj: Bundler seems to combine the options passed with "and", so this fails with:
```
Could not find gem 'capybara (~> 2.13, ~> 3.0)' in any of the gem sources listed in your Gemfile.
```

JonRowe: Fair enough!

PR #2283, opened 2020-03-04
Support symbol argument in #on_queue

nsommer:
Discovered today that expectations fail when using `#on_queue` with symbol args instead of strings, which surprised me, especially given that `queue_as` in active job classes is usually used with a symbol, too. Not sure if you have any philosophy around this, let me know if I can improve anything.


JonRowe:
Thanks

JonRowe:
Changelog in 9ab269052873c241a31404bc3bf8e672f9cf9586 I messed up the commit message :joy:

PR #2286, opened 2020-03-08
Avoid installing documentation gems on TravisCI build

benoittgt:
Probably only relish is needed

Related:
- https://github.com/rspec/rspec-rails/commit/e9949e7b9321d1062ed358202654afa754bd27e8
- https://github.com/rspec/rspec-expectations/pull/1152#issuecomment-596233253


benoittgt:
This is not needed. I will find a different way.

PR #2288, opened 2020-03-12
Generate request specs when scaffold generator is used

JonRowe:
There are some important things to note when generating things, one, your project config will override defaults, even if thats non sensical (e.g. a project can generate controller and request specs here). Two, there are things Rails will generate out of your control.

I took a more minimal approach using rails helpers only and a bit of gsubing to get them to somewhere we'd like, and then had to play whack-a-mole on the tests. Apparently the as scaffolded api views don't work, I guess cause we use sequel lite?

*edit* still playing whack-a-mole


JonRowe:
Fixes #2230

pirj:
Changelog entry?

JonRowe:
Added after the fact to prevent more builds :joy:

Issue #2290, opened 2020-03-17
Possible regression 4.0.0.beta4 to 4.0.0.rc1: List of ActionMailer deliveries is not cleared between specs


CUnknown:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.7.0
Rails version: 6.0.2
RSpec version: 4.0.0.rc1

The following is included for these mailer tests:

RSpec.configure do |config|
  config.include ActionMailer::TestHelper
end

## Observed behaviour

When running specs, we check the to adresses that are set for the ActionMailer::Base.deliveries, like this:
```
expect(ActionMailer::Base.deliveries.map(&:to).flatten.include?(leo.email)).to eq true
expect(ActionMailer::Base.deliveries.map(&:to).flatten.sort).to eq([droid.email, leo.email])
```
This works for version `4.0.0.beta3` and `4.0.0.beta4`, but for `4.0.0.rc1`, this will break.
It seems that the deliveries array is not cleared between specs. The list of email addresses that are stored in the TO of the mailer, will get larger after each spec. Therefore, running each spec seperately works, but when the fourth or fifth spec is ran, the list of adresses is way bigger than expected:

```
     expected: ["droid@cm.nl", "leo@cm.nl"]
            got: ["droid@cm.nl", "droid@cm.nl", "leo@cm.nl", "leo@cm.nl", "leo@hoefsmit.nl"]
```

## Expected behaviour
The ActionMailer instance is reset between each spec and the TO, CC and BCC attributes of ActionMailer::Base.deliveries should be cleared between each spec.

## Can you provide an example app?
I have isolated the issue into an example app:
https://github.com/CUnknown/rspec-rails-bug-example

benoittgt:
At the moment I see nothing relevant here https://github.com/rspec/rspec-rails/compare/v4.0.0.beta3...v4.0.0.rc1

Did you try `v4.0.0.beta3`?

> I think that would be possible, please let me know if that is really neccesarry!

Yes please

CUnknown:
Yes, I tried v4.0.0.beta3 and that works fine.

JonRowe:
Have you tried `v4.0.0.beta4`? (The narrower the scope the better :joy:)

CUnknown:
Ah, I'm sorry, I wasn't aware of a beta 4. `v4.0.0.beta4` seems to work fine too!

pirj:
[...]

PR #2292, opened 2020-03-18
Let the user choose active job queue in system test

benoittgt:
This reverts commit 3e4c770f529237dbaa5877ff80aa0d0b4594b0d1.

Do not remove ::ActionDispatch::IntegrationTest::Behavior because this
lead to unwanted behavior like uncleaned queue for
ActionMailer::Base.deliveries between tests.

As suggested in https://github.com/rails/rails/issues/37270 there is
many other way to implement this. At the moment this line seems to be
enough for our case.

Fix: https://github.com/rspec/rspec-rails/issues/2290

Related:
  - https://github.com/rails/rails/issues/37270

### DIFF
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -75,6 +75,7 @@ module RSpec
         original_after_teardown =
           ::ActionDispatch::SystemTesting::TestHelpers::SetupAndTeardown.instance_method(:after_teardown)
 
+        other.include ::ActionDispatch::IntegrationTest::Behavior
         other.include ::ActionDispatch::SystemTesting::TestHelpers::SetupAndTeardown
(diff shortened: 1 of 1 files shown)


pirj:
Looks good. JRuby failure seems unrelated (`rubocop`).
@CUnknown does this fix work for you?

pirj:
Filed https://github.com/jruby/jruby/issues/6125

lib/rspec/rails/example/system_example_group.rb
```diff
@@ -102,6 +103,7 @@ def driven_by(*args, &blk)
 
         before do
           @routes = ::Rails.application.routes
+          ActiveJob::Base.disable_test_adapter
```
JonRowe: I'm really unsure about this, why are we doing it? We need to be able to change adapters sure, but surely the default should be the test adapter...

PR #2293, opened 2020-03-20
Clear test mailbox from ActionMailer::Base between each example

benoittgt:
We notice in https://github.com/rspec/rspec-rails/issues/2290 that
ActionMailer::Base.deliveries mailbox is not cleaned between example.

Fix: https://github.com/rspec/rspec-rails/issues/2290

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -18,6 +18,7 @@ Bug Fixes:
   (Jonathan Rochkind, #2242)
 * `rails generate generator` command now creates related spec file (Joel Azemar, #2217)
 * Relax upper `capybara` version constraint to allow for Capybara 3.x (Phil Pirozhkov, #2281)
+* Leans ActionMailer test mailbox after each example (Benoit Tigeot, #2293)
 
 ### 4.0.0.beta4
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v4.0.0.beta3...v4.0.0.beta4)
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -258,25 +258,28 @@ RSpec.describe "Configuration" do
     end
 
     describe 'leans test mailbox after each example' do
-      class BaseMailer < ActionMailer::Base
-        default from: 'from@example.com'
+      let(:base_mailer) do
+        Class.new(ActionMailer::Base) do
+          default from: 'from@example.com'
 
-        def welcome(to:)
-          mail(to: to, subject: 'subject', body: render(inline: "Hello", layout: false))
+          def welcome(to:)
+            mail(to: to, subject: 'subject', body: render(inline: "Hello", layout: false))
+          end
         end
       end
+
       before do
         ActionMailer::Base.delivery_method = :test
       end
 
       it 'send to email@' do
-        BaseMailer.welcome(to: 'email@example.com').deliver_now
+        base_mailer.welcome(to: 'email@example.com').deliver_now
 
         expect(ActionMailer::Base.deliveries.map(&:to).flatten.sort).to eq(['email@example.com'])
       end
 
       it 'send to email_2@' do
-        BaseMailer.welcome(to: 'email_2@example.com').deliver_now
+        base_mailer.welcome(to: 'email_2@example.com').deliver_now
 
         expect(ActionMailer::Base.deliveries.map(&:to).flatten.sort).to eq(['email_2@example.com'])
       end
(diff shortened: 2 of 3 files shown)


spec/rspec/rails/configuration_spec.rb
```diff
@@ -256,6 +256,31 @@ def self.application; end
       expect(group.mailer_class).to be(a_mailer_class)
       expect(group.new).to be_a(RSpec::Rails::MailerExampleGroup)
     end
+
+    describe 'cleans test mailbox between each example in all rspec-rails example' do
```
benoittgt: Not sure about this type of test in this file but I didn't wanted a feature test that will be visible in relish doc. Maybe you have a better idea?

JonRowe: 👍

spec/rspec/rails/configuration_spec.rb
```diff
@@ -256,6 +256,31 @@ def self.application; end
...
+      class BaseMailer < ActionMailer::Base
```
JonRowe: Can this be an anonymous class in a let?

spec/rspec/rails/configuration_spec.rb
```diff
@@ -256,6 +256,31 @@ def self.application; end
...
+        BaseMailer.welcome(to: 'email@example.com').deliver_now
+
+        expect(ActionMailer::Base.deliveries.map(&:to).flatten.sort).to eq(['email@example.com'])
+      end
+
+      it 'send to email_2@' do
+        BaseMailer.welcome(to: 'email_2@example.com').deliver_now
+
+        expect(ActionMailer::Base.deliveries.map(&:to).flatten.sort).to eq(['email_2@example.com'])
+      end
```
JonRowe: take it this fails without the change? if so see above then good to go!

spec/rspec/rails/configuration_spec.rb
benoittgt: yep! Good idea

spec/rspec/rails/configuration_spec.rb
benoittgt: Thanks for the quick review.

I need both test to have one that fail. Maybe I could rewrite them to be more explicit but I thought it would make the test more complicated to understand. Like having explicit `example.run`. I don't know.

JonRowe: I'm just checking that they fail without the change to `lib/rspec/rails/configuration.rb`

benoittgt: yes they are. :)

spec/rspec/rails/configuration_spec.rb
pirj: Not only in `rspec-rails` example. If there's one test in a focused run that would require `rails_helper`, we'll be cleaning deliveries after each example, Rails or not.
Maybe just "leans test mailbox after each example"?

pirj:
Does this fix the problem you're experiencing @CUnknown?

pirj:
JRuby failures are unrelated, fixed on JRuby `master`.
Cleared Travis cache for the branch and PR, restarted JRuby jobs.

benoittgt:
> Does this fix the problem you're experiencing @CUnknown?

It has been tested against the repo provided + the test I wrote with the mailer service.

spec/rspec/rails/configuration_spec.rb
benoittgt: Could you explain you prefer?

Done with https://github.com/rspec/rspec-rails/pull/2293/commits/368484b0fe4195edc6fe84dbd95719de3ce0eca4

spec/rspec/rails/configuration_spec.rb
benoittgt: https://github.com/rspec/rspec-rails/pull/2293/commits/f9d2dc8dd029077981cff6e21bde0efb1c343a03

pirj:
[`jruby-head`](http://rubies.travis-ci.org/ubuntu/14.04/x86_64/jruby-head) on Travis seems to be 10h old (`74.22 MiB, 2020-03-20 11:41:48 UTC`), while [the fix is from 3h ago](https://github.com/jruby/jruby/commit/dbad978b80eadd415d2a54bb9e8e6076b7cfb53d). I don't see a good reason to wait for it to update.

@benoittgt Changelog?

benoittgt:
🙏 Thank you!

Changelog.md
```diff
@@ -18,6 +18,7 @@ Bug Fixes:
   (Jonathan Rochkind, #2242)
 * `rails generate generator` command now creates related spec file (Joel Azemar, #2217)
 * Relax upper `capybara` version constraint to allow for Capybara 3.x (Phil Pirozhkov, #2281)
+* Leans ActionMailer test mailbox after each example (Benoit Tigeot, #2293)
```
pirj: *C*leans :D

PR #2295, opened 2020-03-21
Reword mailer specs doc strings for ensuring deliveries are cleared

JonRowe:


benoittgt:
Thanks. Much better.

spec/rspec/rails/configuration_spec.rb
```diff
@@ -257,8 +257,8 @@ def self.application; end
       expect(group.new).to be_a(RSpec::Rails::MailerExampleGroup)
     end
 
-    describe 'leans test mailbox after each example' do
-      let(:base_mailer) do
+    describe 'clears ActionMailer::Base::Deliveries after each example' do
```
pirj: Sorry for the late appearance. Do you think it makes sense to add `order: :defined` here?

benoittgt: Their is no need. You need at least two example, order does not mater. :)

JonRowe: Yep they are order independant, they just need different emails.

pirj: 👍

PR #2296, opened 2020-03-23
Pin JRuby to latest stable

JonRowe:


JonRowe:
No effect.

pirj:
```
       -:io => (be blank),
       +:io => "Warning: the --client flag is deprecated and has no effect most JVMs\n",
```

What if we remove `--client` [here](https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/script/functions.sh#L46)?

PR #2305, opened 2020-03-30
It supports have_enqueud_mail matcher sub class of DeliveryJob.

yalab:
Hi.

I Fixed a bug.

Thanks.

### DIFF
--- a/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
+++ b/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
@@ -22,11 +22,11 @@ if RSpec::Rails::FeatureCheck.has_active_job?
       def email_with_args(arg1, arg2); end
     end
 
-    class SubClassDeliveryJob < ActionMailer::DeliveryJob
+    class DeliveryJobSubClass < ActionMailer::DeliveryJob
     end
 
-    class SubClassUnifiedMailer < ActionMailer::Base
-      self.delivery_job = SubClassDeliveryJob
+    class UnifiedMailerWithDeliveryJobSubClass < ActionMailer::Base
+      self.delivery_job = DeliveryJobSubClass
 
       def test_email; end
     end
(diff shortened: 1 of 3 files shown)


spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -21,6 +21,15 @@ class UnifiedMailer < ActionMailer::Base
       def test_email; end
       def email_with_args(arg1, arg2); end
     end
+
+    class SubClassDeliveryJob < ActionMailer::DeliveryJob
```
JonRowe:
```suggestion
    class DeliveryJobSubClass < ActionMailer::DeliveryJob
```

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -21,6 +21,15 @@ class UnifiedMailer < ActionMailer::Base
...
+    end
+
+    class SubClassUnifiedMailer < ActionMailer::Base
+      self.delivery_job = SubClassDeliveryJob
```
JonRowe:
```suggestion
      self.delivery_job = DeliveryJobSubClass
```

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -21,6 +21,15 @@ class UnifiedMailer < ActionMailer::Base
       def test_email; end
       def email_with_args(arg1, arg2); end
     end
+
+    class SubClassDeliveryJob < ActionMailer::DeliveryJob
+    end
+
+    class SubClassUnifiedMailer < ActionMailer::Base
```
JonRowe:
```suggestion
    class UnifiedMailerWithDeliveryJobSubClass < ActionMailer::Base
```

yalab:
> an interesting use of the <= operator, is that documented anywhere?

See `ri 'Class#<=' on your terminal.

```
= Class#<=

(from ruby core)
=== Implementation from Module
------------------------------------------------------------------------
  mod <= other   ->  true, false, or nil

------------------------------------------------------------------------

Returns true if mod is a subclass of other or is the
same as other. Returns nil if there's no relationship between
the two. (Think of the relationship in terms of the class definition:
"class A < B" implies "A < B".)
```

PR #2307, opened 2020-03-30
Updates from rspec-dev (2020-03-30)

benoittgt:


pirj:
I think this does not apply to `rspec-rails` anymore.

Issue #2265, opened 2020-01-15
Release rspec-rails 4


benoittgt:
Hello

This is an issue to follow the release of rspec-rails 4.

As mentioned in https://github.com/rspec/rspec-rails/pull/2242#issuecomment-574546068 we have pending PR/issue that needs review or work. 

I think we should tag them with [v4.0](https://github.com/rspec/rspec-rails/pulls?utf8=%E2%9C%93&q=is%3Aopen+sort%3Aupdated-desc+label%3Av4.0+) or set a milestone to help us to work on.

At the moment I don't know where I could help.

Thanks 🌝

JonRowe:
I think its just #2242 and #1938 that are really holding me off releasing

klyonrad:
I think that [this behaviour (scaffolding created controller specs and empty request spec](https://github.com/rspec/rspec-rails/issues/2230#issuecomment-573320813) that I introduced with irritate some users.

micodel:
Why update the README to state 4.0 before its release?

JonRowe:
Readme refers to the current branch and commit, the relevant branches refer to their relevant versions.

micodel:
I'm sure you are more experienced than me 😅, but what about having some sort of versioned 4.0 branch with a living README, and letting master (aka the main page README) reflect the latest stable release? I have seen this implemented in other gems and its a nice touch

JonRowe:
Thats really not how Github works, the default branch is the target of all PRs and is thus expected to be the core development branch, you can change what that branch is but it is always the default landing branch. Thus you should **always** assume the README.md is for the development branch and check the releasing strategy accordingly.

JonRowe:
We could not have the current version being worked on in the readme but we've been on the verge of releasing 4.0.0 for months so we updated it when we merged that branch ready to release and it just hasn't happened yet :joy:

benoittgt:
I think we can already release the 4.0. Then iterate for the upcoming features.

klyonrad:
As mentioned, I introduced a bug with the *generate request specs* PR

> [this behaviour (scaffolding creates controller specs and empty request spec](https://github.com/rspec/rspec-rails/issues/2230#issuecomment-573320813)

So maybe we should revert it, release and then iterate

JonRowe:
So we need to release those remaining issues in a major release, so if not `4.0.0` then they'll have to wait until `5.0.0` however I really don't want to ship controller specs generated in `4.0.0`

[...]

PR #2217, opened 2019-12-03
Fix generator spec generator

joel:
# Context

Rails offer the possibility to create our own set of generators [[1]](https://guides.rubyonrails.org/generators.html#customizing-your-workflow). While executing the following command 

```
bin/rails generate generator my_awesome_generator
```

the test framework is called

See [hook_for :test_framework](https://github.com/rails/rails/blob/v6.0.1/railties/lib/rails/generators/rails/generator/generator_generator.rb#L15)

RSpec must give the expected generator

# Problem 

The given command get this error 

```
bin/rails generate generator foo
      create  lib/generators/foo
      create  lib/generators/foo/foo_generator.rb
      create  lib/generators/foo/USAGE
      create  lib/generators/foo/templates
       error  rspec [not found]
``` 

# History

@yhirano55 provided an implementation https://github.com/rspec/rspec-rails/pull/1847 Jul 19, 2017, which I don't know why it wasn't integrated in the first place, @penelopezone?

@tragiclifestories opened an issue about it https://github.com/rspec/rspec-rails/issues/2084 on  Feb 26, 2019.

@ConSou implemented a version https://github.com/rspec/rspec-rails/pull/2085 the same day and it was merged a couple of days after by @JonRowe on Mar 1, 2019.

However, as @kdiogenes commented a couple of days ago here https://github.com/rspec/rspec-rails/pull/2085#discussion_r348795229 this doesn't work. 

# Environment

* Rails 6.0.1
* Rspec Rails (3.9.0)

- [1] https://guides.rubyonrails.org/generators.html#customizing-your-workflow

This PR fix the issue.

```
bin/rails generate generator --help
Running via Spring preloader in process 26116
Usage:
  rails generate generator NAME [options]

Options:
      [--skip-namespace], [--no-skip-namespace]  # Skip namespace (affects only isolated applications)
      [--namespace], [--no-namespace]            # Namespace generator under lib/generators/name
                                                 # Default: true
  -t, [--test-framework=NAME]                    # Test framework to be invoked
                                                 # Default: rspec

Rspec options:
  [--generator-specs], [--no-generator-specs]  # Generate generator specs

Runtime options:
  -f, [--force]                    # Overwrite files that already exist
  -p, [--pretend], [--no-pretend]  # Run but do not make any changes
  -q, [--quiet], [--no-quiet]      # Suppress status output
  -s, [--skip], [--no-skip]        # Skip files that already exist

Description:
    Stubs out a new generator at lib/generators. Pass the generator name as an argument,
[...]

PR #2324, opened 2020-04-27
Update documentation about avaliable generators in v4

choosen:
Add missing entries to the list in generator documentation.

Channel generator was introduced with version rspec 4, so patch for previous pages needs to be a little bit changed

Related with #2265 

* [x] double checked

### DIFF
--- a/features/Generators.md
+++ b/features/Generators.md
@@ -25,8 +25,5 @@ The same generator pattern is available for all specs:
 * feature
 * job
 * channel
-* generator
 * mailbox
-* plugin
-* policy
 * system
--- a/features/Generators.md
+++ b/features/Generators.md
@@ -20,10 +20,11 @@ The same generator pattern is available for all specs:
 * helper
 * view
 * mailer
-* observer
 * integration
 * feature
 * job
 * channel
+* generator
 * mailbox
+* request
 * system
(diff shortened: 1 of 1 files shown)


klyonrad:
Do we care about updating the list in readme.md as well?

features/Generators.md
```diff
@@ -25,8 +25,5 @@ The same generator pattern is available for all specs:
 * feature
 * job
 * channel
-* generator
```
pirj: I think that a generator generator is Rails-related, I remember a recent fix for its specs https://github.com/rspec/rspec-rails/pull/2217/files#diff-3fb1bfb25a4bda08b0ad2295bb53297aR3

choosen: as I can see here: https://github.com/rspec/rspec-rails/tree/ca988b0697524717e41ce71663869708273b1966/lib/generators/rspec
Generators are aimed only for rails : )

So generate generator can be also included here.
I got a plugin by mistake from test unit 🤦 

I'm syncing all 15 generators with this list

choosen:
@klyonrad  
From readme:
```
# List all RSpec generators
$ rails generate --help | grep rspec
```
I think it is not so important for readme

features/Generators.md
choosen: policy generator came from pundit gem

free to merge @pirj

JonRowe:
Thanks~

benoittgt:
Shouldn't we backport it to 4-0-maintenance?

JonRowe:
Ah yep, thanks for the reminder :joy:

PR #2325, opened 2020-05-04
fix-finding-extension

masarakki:
6.x: https://github.com/rails/rails/blob/v6.0.3/actionview/lib/action_view/template/resolver.rb#L349-L351
5.x: https://github.com/rails/rails/blob/v5.2.4.3/actionview/lib/action_view/template/resolver.rb#L368-L374
4.x: https://github.com/rails/rails/blob/v4.2.11.3/actionview/lib/action_view/template/resolver.rb#L347-L349

Resolver search files and resolve most matched file with **symbolized** string,
so `_default_render_options` must return symbolized values.

### DIFF
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -150,10 +150,12 @@ module RSpec
           match = path_regex.match(_default_file_to_render)
 
           render_options = {template: match[:template]}
-          render_options[:handlers] = [match[:handler]] if match[:handler]
+
+          # remove stringified parts when dropping rails-4.x support
+          render_options[:handlers] = [match[:handler], match[:handler].to_sym] if match[:handler]
           render_options[:formats] = [match[:format].to_sym] if match[:format]
-          render_options[:locales] = [match[:locale]] if match[:locale]
-          render_options[:variants] = [match[:variant]] if match[:variant]
+          render_options[:locales] = [match[:locale], match[:locale].to_sym] if match[:locale]
+          render_options[:variants] = [match[:variant], match[:variant].to_sym] if match[:variant]
 
           render_options
         end
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -148,19 +148,19 @@ module RSpec::Rails
         it "converts the filename components into render options" do
           allow(view_spec).to receive(:_default_file_to_render) { "widgets/new.en.html.erb" }
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], formats: [:html], handlers: ['erb']}, {}, nil])
+          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en', :en], formats: [:html], handlers: ['erb', :erb]}, {}, nil])
         end
 
         it "converts the filename with variant into render options" do
           allow(view_spec).to receive(:_default_file_to_render) { "widgets/new.en.html+fancy.erb" }
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], formats: [:html], handlers: ['erb'], variants: ['fancy']}, {}, nil])
+          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en', :en], formats: [:html], handlers: ['erb', :erb], variants: ['fancy', :fancy]}, {}, nil])
         end
 
         it "converts the filename without format into render options" do
           allow(view_spec).to receive(:_default_file_to_render) { "widgets/new.en.erb" }
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], handlers: ['erb']}, {}, nil])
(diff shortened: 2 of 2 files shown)


JonRowe:
:wave: Thanks for contributing this! Seems like we need an integration test to check this as the current spec is obviously not enough to catch this, I believe 5.2 does string interpolation so this might be ok for all versions we support but I'm not sure.

masarakki:
> I believe 5.2 does string interpolation 

oh...
I think it may ok with like this code

```
render_options[:variants] = [match[:variant], match[:variant].to_sym] if match[:variant]
```

pirj:
We still soft-support 4.2, so intentionally breaking it would be irresponsible. Your code seems to be quite ok to support both. Can you please check if your fix works for your project?

May I ask you to add a note to remove stringified parts once we completely drop 4.2?

masarakki:
updated.

masarakki:
some error occurred in travis, but I can't detect what's wrong from error message.

pirj:
That's weird, I can't reproduce locally, can you by running `cucumber features/system_specs/system_specs.feature`? Ruby 2.5.1 + Rails 6.0, all green.

masarakki:
all pass on my local machine.
ruby 2.6.6 + rails 6.0.3

benoittgt:
The CI error happens in another build too. https://github.com/rspec/rspec-rails/pull/2328#issue-415156360

I am gonna try to look at it, this afternoon.

benoittgt:
Ok I don't know how is giving a status 9. I wrote a quick and dirty script

```sh
bin/rake clobber:app
bin/rake generate:app

bin/rake generate:stuff


mkdir -p tmp/example_app/spec/system/
tee -a tmp/example_app/spec/system/widget_system_spec.rb << END
require "rails_helper"
RSpec.describe "Widget management", :type => :system do
  before do
    driven_by(:selenium_chrome_headless)
  end
  it "enables me to create widgets" do
    visit "/widgets/new"
    fill_in "Name", :with => "My Widget"
    click_button "Create Widget"
    expect(page).to have_text("Widget was successfully created.")
  end
end
END

cd tmp/example_app/
bundle install
bundle exec rspec --backtrace spec/system/widget_system_spec.rb
echo $?

cd ../../

bin/cucumber features/system_specs/system_specs.feature
```

Capture d’écran 2020-05-08 à 20 56 17

The highlighted line show a status 0... then  we a have a status 9... I don't get it. I am in favor of removing this status expectation. IMHO we should care only if the status is 0 or 1 but not something else.

benoittgt:
@mvz Do you have an idea for my last comment? 
I dig into `aruba`'s code to try to spot why we have a status 9 but didn't find an answer.

JonRowe:
Feel free to cherry-pick 1f723f5 from #2332 into your branch which ignores the exit status 9 issue

JonRowe:
You can also rebase this off master now

mvz:
> @mvz Do you have an idea for my last comment?
> I dig into `aruba`'s code to try to spot why we have a status 9 but didn't find an answer.

If aruba is doing its job properly this means rspec exits with status 9. I'm attempting to reproduce this locally to see if I can debug this.

pirj:
Rebased on master.

pirj:
@masarakki Sorry for the trouble with CI.
Are you up to add an integration test somewhere in features/view_specs/view_spec.feature that fails without your change?

lib/rspec/rails/example/view_example_group.rb
```diff
@@ -150,10 +150,12 @@ def _default_render_options
           match = path_regex.match(_default_file_to_render)
 
           render_options = {template: match[:template]}
-          render_options[:handlers] = [match[:handler]] if match[:handler]
+
+          # remove stringified parts when dropping rails-4.x support
+          render_options[:handlers] = [match[:handler], match[:handler].to_sym] if match[:handler]
```
JonRowe, line 155: We don't want to match both, we just want to match the right type for each rails version.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -148,19 +148,19 @@ def _default_file_to_render; end # Stub method
         it "converts the filename components into render options" do
           allow(view_spec).to receive(:_default_file_to_render) { "widgets/new.en.html.erb" }
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], formats: [:html], handlers: ['erb']}, {}, nil])
+          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en', :en], formats: [:html], handlers: ['erb', :erb]}, {}, nil])
```
JonRowe, line 151: The problem with these specs are they are making exact expectations on value we know we're sending, we need an integration test for this that shows its a problem that they are not symbols.

pirj:
@masarakki Sorry for pushing. Will you have a moment to wrap this up?

masarakki:
It is easy to match only right type for rails version, but it is very hard to test it.
We should test which file rendered, not what parameters are.

leboshi:
FYI, this PR is moot as of 3ccd1f.

pirj:
Should we close this, @masarakki ? Does it resolve your case?

masarakki:
this problem fixed by #2521 .
thanks!

PR #2329, opened 2020-05-09
Define url_options as per system test case does

### LINKED ISSUE #2327: System specs broken with Rails 6.0.1+

Ruby version: 2.6.5p114
Rails version: Rails 6.0.1.rc1
RSpec version: RSpec 3.9 (rspec-rails 4.0.0)

## Observed behaviour

Starting with Rails 6.0.1.rc1 urls in system specs are being generated by Rails url helpers for host `www.example.com` instead of `http://127.0.0.1` as was the case in Rails 6.0.0 and before. Due to this change, all system specs fail with timeout errors.

failures_r_spec_example_groups_website_works_939

Specifically [this line and commit](https://github.com/rails/rails/commit/ea77dbffdc81d6e40bbde155961ab57f1450d1e4#diff-f26f82ebcdedae7b27f9fe5151096128L10) introduced this change in RSpec system specs behavior.

## Expected behaviour

Urls should be generated with host set to `http://127.0.0.1`.

https://github.com/nnc/rspec_system_specs_bug

Test app includes both a system test and a system spec, just to show that system tests continue working fine before and after the offending commit.

This issue is the same in all Rails versions after 6.0.1.rc1 so I didn't include those. But there is a commit in test app with Rails 6.0.0 (last Rails release before https://github.com/rails/rails/commit/ea77dbffdc81d6e40bbde155961ab57f1450d1e4) where system spec is passing.

JonRowe:
Rails switched to manually defining `url_options` when they switched away from inheriting from integration test case.

See:
https://github.com/rails/rails/commit/ea77dbffdc81d6e40bbde155961ab57f1450d1e4#diff-ff5f366a442b8761d35f94807700fa77R162

This should fix #2327


lib/rspec/rails/example/system_example_group.rb
```diff
@@ -100,6 +100,10 @@ def driven_by(driver, **driver_options, &blk)
           @driver = ::ActionDispatch::SystemTestCase.driven_by(driver, **driver_options, &blk).tap(&:use)
         end
 
+        def url_options
```
nnc, line 103: @JonRowe this method is never invoked by `DISABLE_SPRING=1 bundle exec rspec spec/system/websites_spec.rb` on the example app in #2327

benoittgt:
I don't have anymore timeout with the PR, but I had no timeout with `rspec-rails` master too. 
https://github.com/benoittgt/rspec_system_specs_bug

I think the problem is related to this:
Capture d’écran 2020-05-09 à 18 10 32

If you do :
```diff
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,9 @@ ruby '2.6.6'

 # Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
 gem 'rails', '6.0.0.rc1'
+gem 'actionpack', '6.0.0.rc1'
+gem 'activesupport', '6.0.0.rc1'
+gem 'railties', '6.0.0.rc1'
```

You do no have the timeout anymore. 🤔

nnc:
@benoittgt Rails 6.0.0.rc1 doesn't have the timeout issue because the problem was first introduced in Rails 6.0.1.rc1. Everything before that, including 6.0.0.rc1 is OK for me as well.

benoittgt:
> @benoittgt Rails 6.0.0.rc1 doesn't have the timeout issue because the problem was first introduced in Rails 6.0.1.rc1. Everything before that, including 6.0.0.rc1 is OK for me as well.

Sorry... 🤦

features/system_specs/system_specs.feature
```diff
@@ -106,4 +106,4 @@ Feature: System spec
         When I run `rspec spec/system/widget_system_spec.rb`
         Then the output should contain "1 example, 0 failures"
         And the output should not contain "starting Puma"
-        And the exit status should be 0
+        And the exit status should be 0 or I debug
```
benoittgt, line 109: If you understand why, I am interested. I wrote this https://github.com/rspec/rspec-rails/pull/2325#issuecomment-625967122

PR #2332, opened 2020-05-12
Add back Ruby 2.2 support for Rails 5

JonRowe:


JonRowe:
@benoittgt I've temporarily suppressed our phantom error as I'd like to do a release to resolve this issue in particular

benoittgt:
No problem Jon.

benoittgt:
If you need some help on this @JonRowe

JonRowe:
I've just been playing whackamole with the Travis build @benoittgt :joy: I think it'll pass now, care to re-review?

.travis.yml
```diff
@@ -77,6 +77,11 @@ matrix:
       env: RAILS_VERSION='~> 5.2.0'
     - rvm: 2.3.8
       env: RAILS_VERSION='~> 5.2.0'
+    - rvm: 2.2.10
+      env: RAILS_VERSION='~> 5.2.0'
+      allow_failure: true
+    - rvm: 2.2.10
+      env: RAILS_VERSION='5-2-stable'
```
benoittgt, line 84: Why do we want to try stable and `~> 5.2.0`?

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -234,17 +236,29 @@ def _view; end # Stub method
...
             allow(view).to receive(:render) { :value }
             run_count += 1
           end
           group = RSpec::Core::ExampleGroup.describe 'a view', type: :view do
-            specify { true }
+            specify { expect(true).to eq true }
           end
-          group.run
+          group.run(failure_reporter)
+          expect(failure_reporter.exceptions).to eq []
```
benoittgt, line 261: I do not understand why we need the `FailureReporter` anymore?

.travis.yml
JonRowe: `5-2-stable` is `2.2.10` compatible, it contains an unreleased bugfix for Ruby 2.2, `~> 5.2.0` is included but will fail, eventually when patched the pending spec will fail this build and we can remove the `5-2-stable` the same as the other versions.

spec/rspec/rails/example/view_example_group_spec.rb
JonRowe: Its included for future debugging, we've had this spec fail before and its really hard to see why without this pattern, I'm going to use it in other places for the same reason.

.travis.yml
benoittgt: Ok. Thanks!

spec/rspec/rails/example/view_example_group_spec.rb
benoittgt: Nice! Maybe we could add a little bit of documentation on the top of file or somewhere?

PR #2335, opened 2020-05-15
Test Aruba failure

mvz:
This is just an attempt to figure out what's going on ...

### DIFF
--- a/features/system_specs/system_specs.feature
+++ b/features/system_specs/system_specs.feature
@@ -106,4 +106,4 @@ Feature: System spec
         When I run `rspec spec/system/widget_system_spec.rb`
         Then the output should contain "1 example, 0 failures"
         And the output should not contain "starting Puma"
-        And the exit status should be 0 (ignoring CI failure)
+        And the exit status should be 0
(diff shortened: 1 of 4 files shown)


JonRowe:
Thanks for the help debugging this @mvz, any tips on how to find out what its happening underneath are appreciated!

For what its worth my working theory is that perhaps travis is `kill -9`ing the selenium / chrome driver process as we exit and this is bubbling up to us, I can say that the rspec scenario *is* passing, so this status is coming from an interaction of the processes involved... RSpec itself would not return `9` unless configured to do so which it is not in this suite.

benoittgt:
I found a way to fix it. https://github.com/rspec/rspec-rails/commit/b628615073f6514ebf655cc74724de74292f5a72

But I am not able to explain why we had a status 9 when using cucumber and not when running the test directly. I am not sure we need to spend more time on this... even if I am curious about the answer. :)

mvz:
That's great news @benoittgt.

features/system_specs/system_specs.feature
```diff
@@ -82,6 +82,7 @@ Feature: System spec
       When I run `rspec spec/system/some_job_system_spec.rb`
       Then the example should pass
 
+    @announce-output
```
benoittgt, line 85: This is a very good tips. Thanks a lot.

PR #2336, opened 2020-05-16
Prevent travis build from failing to install properly Google chrome

benoittgt:
Hello

I was upset with the failing build with selenium and wrong exit status. I compared the last green version or our build for a specific rails version against one failing build. I _diffed_ the travis output for hours.. then spotted this difference:

Capture d’écran 2020-05-16 à 10 27 00

`chromedriver` was not properly installed. I still don't know what is an exit status 9, and from who. But I think that's good enough.

To fix this error. Simply add `dpkg` package when we start our travis build.

Related: https://github.com/travis-ci/travis-ci/issues/9361

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ addons:
   chrome: stable
   apt:
     packages:
+      - dpkg
       - chromium-chromedriver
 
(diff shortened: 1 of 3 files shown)


benoittgt:
Added to `4-0-maintenance`

PR #2340, opened 2020-05-20
Install chrome driver for current chrome version

JonRowe:


JonRowe:
Sadly xenial (or maybe just travis) has upgraded chrome but not chromedriver, seems safer to try to download the correct version of chromedriver for the version of chrome we have, but it wasn't super easy to work out how :P

PR #2342, opened 2020-05-20
Drop outdated documentation

pirj:
Mostly related to either RSpec Rails 3 or even earlier.


features/GettingStarted.md
```diff
@@ -1,19 +1,19 @@
...
+    $ cd example_app
 
 ### Add rspec-rails to the Gemfile
 
-    $ echo 'gem "rspec-rails", :group => [:development, :test]' >> Gemfile
+    $ echo 'gem "rspec-rails", group: [:development, :test]' >> Gemfile
 
 ### Install the bundle
 
-    $ bundle install
```
pirj, line 16: `install` is not enough as it doesn't add to `Gemfile.lock`.

features/GettingStarted.md
```diff
@@ -29,15 +29,15 @@ the files in the `spec` directory to RSpec.
 
 ### Run migrations
 
-    $ rake db:migrate && rake db:test:prepare
+    $ rails db:migrate && rails db:test:prepare
 
 ### Run RSpec
 
     $ rake spec
```
pirj, line 36: Never using this, not sure if it can/should be `rails spec` now or `rake` is preferrable just like with running other common tasks.

features/GettingStarted.md
```diff
@@ -29,15 +29,15 @@ the files in the `spec` directory to RSpec.
...
 
 or
 
-    $ rspec spec --format documentation
```
pirj, line 40: Doc format is the default when running just one file.
`spec` is the default directory.

features/RailsVersions.md
```diff
@@ -8,3 +9,5 @@
     4.2           | >= 3.1
     5.0           | >= 3.5
     5.1           | >= 3.6
+    5.2           | >= 3.8
```
pirj, line 12: It might be that Rails 5.2 was supported in 3.6/3.7, but in 3.8 there were some fixes to improve Rails 5.2 support.

features/upgrade/README.md
```diff
@@ -1,178 +1,5 @@
-# Upgrading from rspec-rails-2.x to rspec-rails-3
+# Upgrading from rspec-rails 3.x to version 4
 
-For detailed information on the general RSpec 3.x upgrade process see the [RSpec
-Upgrade docs](https://relishapp.com/rspec/docs/upgrade).
+There is nothing that needs to be done to upgrade from RSpec Rails 3 to version 4.0.
```
pirj, line 3: Can't think of anything, really.
Should we mention that the minimum Ruby version is now 2.2, and we recomment to update Rails 4.2 to a newer version?

features/.nav
```diff
@@ -1,5 +1,3 @@
-- upgrade:
-  - from_1x_to_2x.md (From rspec-rails-1.x to rspec-rails-2)
```
pirj, line 2: Ooops, leftover.

features/upgrade/README.md
```diff
@@ -1,178 +1,5 @@
...
-`ActiveRecord` behavior complexity which would otherwise be good to expose.
-
-Some alternatives are:
-
-- Wrap calls to `ActiveRecord` objects in more specific domain models and
-  services
-- Use new unsaved `ActiveRecord` instances (e.g. `Model.new`)
-- Consider [partial mocks](https://www.relishapp.com/rspec/rspec-mocks/docs/basics/partial-test-doubles) on an `ActiveRecord` instance
-- Let the specs hit database directly where appropriate
+If you are upgrading from an earlier version of RSpec Rails, please consult [this document](https://relishapp.com/rspec/rspec-rails/v/3-9/docs/upgrade).
```
JonRowe, line 5:
```suggestion
If you are upgrading from an earlier version of RSpec Rails, please consult [the upgrading 2.x to 3.x guide](https://relishapp.com/rspec/rspec-rails/v/3-9/docs/upgrade).
```

features/directory_structure.feature
```diff
@@ -63,7 +63,7 @@ Feature: Directory Structure
   their location on the filesystem. This was both confusing to new users and not
   desirable for some veteran users.
 
-  In RSpec 3, this behavior must be explicitly enabled:
+  In RSpec 4, this behavior must be explicitly enabled:
```
JonRowe, line 66:
```suggestion
  This behaviour must be explicitly enabled:
```

features/RailsVersions.md
JonRowe: 👍

features/GettingStarted.md
JonRowe: Explicit is better :/

features/GettingStarted.md
JonRowe: Is rails available in 5?

pirj: It is https://github.com/rails/rails/blob/5-0-stable/guides/source/command_line.md#binrails

features/upgrade/README.md
benoittgt: I think it's a good idea :)

features/GettingStarted.md
```diff
@@ -1,19 +1,19 @@
...
 
-    $ rails new example
-    $ cd example
+    $ rails new example_app
+    $ cd example_app
 
 ### Add rspec-rails to the Gemfile
 
-    $ echo 'gem "rspec-rails", :group => [:development, :test]' >> Gemfile
+    $ echo 'gem "rspec-rails", group: [:development, :test]' >> Gemfile
```
benoittgt, line 12: Side note but isn't against properly organizing your Gemfile?

pirj: ¯\\\_(ツ)_/¯

features/GettingStarted.md
JonRowe: Can you clarify? I thought `bundle` was short for `bundle install` and should be sufficient for a new app...

features/upgrade/README.md
JonRowe: Yes this should mention that Rspec Rails is intended for Rails 5 and 6 and that the minimum ruby version is 2.2

features/GettingStarted.md
JonRowe: One liner is a one liner :joy:

features/upgrade/README.md
```diff
@@ -1,178 +1,9 @@
...
 
-For detailed information on the general RSpec 3.x upgrade process see the [RSpec
-Upgrade docs](https://relishapp.com/rspec/docs/upgrade).
+RSpec Rails 4 requires Ruby 2.2.
 
-There are several changes specific to `rspec-rails` to be aware of:
+Please also keep in mind that even though RSpec Rails 4 has soft support for Rails 4.2, it's recommended to use Rails 5+.
 
-- [Default helper files created in RSpec 3.x have changed](#default-helper-files)
+Other than that, there is nothing that needs to be done to upgrade from RSpec Rails 3 to version 4.0.
```
JonRowe, line 7: I would rephrase this, I feel its important to emphasise the last part, and less so the first part.

```suggestion
RSpec Rails 4 is a major version under semantic versioning, it allowed us to change the supported Rails versions to 5 and 6 only. There are no changes required to upgrade to RSpec Rails 4 if you are using Rails 5 or 6.

If you are using Rails 4.2 you can use RSpec Rails 4, but note that support for it is not maintained, we consider this a breaking change hence the version change, and you must be on Ruby 2.2 as a minimum.
```

pirj: Much better, thanks.

features/GettingStarted.md
pirj: > Install the gems specified in your Gemfile(5). If this is the first time you run bundle install (and a Gemfile.lock does not exist), Bundler will fetch all remote sources, resolve dependencies and install all needed gems.

You're right. I had an impression that `bundle install` is not supposed to create/update `Gemfile.lock`. It seems that it's only the case for `bundle install --frozen`.
Will revert this.

PR #2346, opened 2020-05-25
Add extra spec for time offsets / frozen time

JonRowe:


spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -223,6 +223,14 @@ def self.name; "LoggingJob"; end
       }.to have_enqueued_job.at(time)
     end
 
+    it "works with time offsets", skip: !defined?(ActiveSupport::Testing::TimeHelpers) do
+      class_eval { include ActiveSupport::Testing::TimeHelpers }
```
pirj, line 227: Maybe add a `context` and use a plain include?

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -33,6 +33,8 @@ def self.find(_id)
 end
 
 RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_active_job? do
+  include ActiveSupport::Testing::TimeHelpers if defined?(ActiveSupport::Testing::TimeHelpers)
```
pirj, line 36: When it might not be available exactly?

benoittgt:
LGTM. I am looking for the sprockets bumping issue on the CI.

spec/rspec/rails/matchers/active_job_spec.rb
JonRowe: Rails 4.2, 5.0 and 5.1

pirj: I mean it's clear that `freeze_time` isn't part of it, but [TimeHelpers](https://api.rubyonrails.org/v4.2/classes/ActiveSupport/Testing/TimeHelpers.html) should be always available, and it's not necessary to check if they are defined (I hope we're requiring them somewhere).

JonRowe: Oh ok, I was just checking `freeze_time` feel free to open a PR if you'd like to improve it, maybe backfill an implementation fir `freeze_time` if you have time?

pirj: > backfill an implementation for freeze_time

Noway! :D
If just for tests - it's kind of useless since we already run those specs against Rails that have it. For usage in tests - there's TimeCop (god forgive me) - wouldn't dare to write yet another one we'll have to support.

PR #2347, opened 2020-05-25
Prevent sprocket install failure when we are using Ruby below 2.5

benoittgt:
Randomly sprocket 4 can be installed on the CI on Ruby version below 2.5.

We should not have this behavior. 

This change prevent this kind of random issue on the CI.

Related:
* https://github.com/rails/sprockets/pull/604

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,6 +63,10 @@ end
 
 gem "rubyzip", '~> 1.2'
 
+if RUBY_VERSION.to_f < 2.5
+  gem 'sprockets', '~> 3.7.0'
+end
+
 if RUBY_VERSION.to_f >= 2.3
   gem 'rubocop', '~> 0.80.1'
 end


Gemfile
```diff
@@ -63,6 +63,10 @@ end
 
 gem "rubyzip", '~> 1.2'
 
+if RUBY_VERSION.to_f < 2.5
+  gem 'sprockets', '~> 3.7.0'
```
pirj, line 67: Won't Bundler figure it out on its own? They do specify a version:
```
  s.required_ruby_version = '>= 2.5.0'
```
shouldn't `bundle` take this into account and check out `sprockets` < 4.0 when it understands that 4+ won't work?

benoittgt: This what I thought but the CI failed earlier for this PR https://github.com/rspec/rspec-rails/pull/2346. But I am not able to find it again. It was on Ruby 2.4.

I will close it for now. We will see if it happen again.

PR #2349, opened 2020-05-27
Check if TestCase responds to the fixture_path setter

koenpunt:
The existence of the getter doesn't mean there's also a setter, so this resulted in failures like the following:

```
1) Admin::CollectionsController POST #create creates a record
     Failure/Error: banner_image: fixture_file_upload('alpaca.jpg', 'image/jpg')

     NoMethodError:
       undefined method `fixture_path=' for ActionController::TestCase:Class
       Did you mean?  fixture_path
```

### DIFF
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -18,6 +18,10 @@ module ExampleAppHooks
       copy_file 'app/views/some_templates/bar.html'
       copy_file 'spec/verify_custom_renderers_spec.rb'
       copy_file 'spec/verify_fixture_warning_spec.rb'
+      copy_file 'spec/verify_fixture_file_upload_spec.rb'
+      File.open('spec/fixtures/file_upload.txt', 'w') do |f|
+        f.write('Hello World')
+      end
       run('bin/rake db:migrate')
     end
 
--- a/example_app_generator/spec/verify_fixture_file_upload_spec.rb
+++ b/example_app_generator/spec/verify_fixture_file_upload_spec.rb
@@ -1,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe 'Example App' do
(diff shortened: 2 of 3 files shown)


benoittgt:
Do you think it is possible to add a test?

koenpunt:
I think so, will have a look tonight.

JonRowe:
Are you seeing an issue from this @koenpunt? I think this method is always available isn't it?

koenpunt:
Yeah I am, `fixture_file_upload` method doesn't work in my Rails 6 application

JonRowe:
In that case we'd definitely like an integration spec for this so we can catch such problems in future :joy:

benoittgt:
@koenpunt, if you need any help feel free to ask. I will be happy to help you.

benoittgt:
Little bump on this one. Still available if you need help @koenpunt.

koenpunt:
Yes, sorry, just started a new job so didn't really had time to take a look at this again. Will try to get back to this ASAP.

koenpunt:
I've now added a test which I hoped was going to fail without my changes, but that isn't the case. Any pointers for how I could try to implement the test differently?

koenpunt:
After some further investigation I realised that I actually added the `fixture_path` method to `ActionDispatch::TestProcess` to allow usage of `fixture_file_upload` from FactoryBot factories. But I didn't implement the setter, so this caused `rspec-rails` to raise the `NoMethodError`.

I've now changed my implementation to not modify `ActionDispatch::TestProcess`, but instead monkey patch the `fixture_file_upload` method into FactoryBot, which solves the issue.

That said, I still think the `respond_to?(:fixture_path)` isn't correct, because it checks the getter, but then uses the setter.

Should I create a test where I add the `fixture_path` method, to demonstrate the failure?

JonRowe:
Closed by #2370

PR #2350, opened 2020-05-28
Improve enqueued_job at specs and warn about precision.

JonRowe:
There is an issue with our time matching for `enqueued_job`, Rails allows microsecond (usec) precision in job queue values, and uses `.to_f` for exact time scheduling, which we replicated in #2157, without this `to_f` you cannot check exact time scheduled jobs.

However when Rails schedules a job in the future using `wait: n` it is doing `n.seconds.from_now` which does not get passed through `to_f`, there is no way for us to detect which is correct so we'll stick with out existing handling, but warn about this precision issue, adding `change(usec: 0)` or switching to `n.seconds.from_now`...

Ironically I found this by trying to remove `freeze_time` from our own specs, it appears that `travel_to` specific time, keeps `usec` precision, but not `freeze_time`?

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -163,7 +163,29 @@ module RSpec
             return job[:at].nil? if @at == :no_wait
             return false unless job[:at]
 
-            values_match?(@at, Time.at(job[:at]))
+            scheduled_at = Time.at(job[:at])
+            values_match?(@at, scheduled_at) || check_for_inprecise_value(scheduled_at)
+          end
+
+          def check_for_inprecise_value(scheduled_at)
+            return unless Time === @at && values_match?(@at.change(usec: 0), scheduled_at)
+
+            RSpec.warn_with((<<-WARNING).gsub(/^\s+\|/, '').chomp)
+            |[WARNING] Your expected `at(...)` value does not match the job scheduled_at value
+            |unless microseconds are removed. This precision error often occurs when checking
+            |values against `Time.current` / `Time.now` which have usec precision, but Rails
+            |uses `n.seconds.from_now` internally which has a usec count of `0`.
+            |
+            |Use `change(usec: 0)` to correct these values. For example:
+            |
+            |`Time.current.change(usec: 0)`
+            |
+            |Note: RSpec cannot do this for you because jobs can be scheduled with usec
+            |precision and we do not know wether it is on purpose or not.
+            |
+            |
+            WARNING
+            false
           end
 
           def set_expected_number(relativity, count)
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -163,7 +163,27 @@ def at_match?(job)
...
+          def check_for_inprecise_value(scheduled_at)
+            return unless Time === @at && values_match?(@at.change(usec: 0), scheduled_at)
+
+            RSpec.warn_with((<<-WARNING).gsub(/^\s+\|/, '').chomp)
+            |[WARNING] Your expected `at(...)` value does not match the job scheduled_at value
+            |unless microseconds are removed. This precision error often occurs when checking
+            |values against `Time.current` / `Time.now` which have usec precision, but Rails
+            |uses `n.seconds.from_now` internally which has a usec count of `0`.
+            |
+            |Use `change(usec: 0)` to correct these values.
```
benoittgt, line 179: Maybe add something that can be copy pasted? Like
```suggestion
            |Use `Time.current.change(usec: 0)` or `Time.now.change(usec: 0)` to correct these values.
```

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -163,7 +163,27 @@ def at_match?(job)
...
 
-            values_match?(@at, Time.at(job[:at]))
+            scheduled_at = Time.at(job[:at])
+            values_match?(@at, scheduled_at) || check_for_inprecise_value(scheduled_at)
+          end
+
+          def check_for_inprecise_value(scheduled_at)
+            return unless Time === @at && values_match?(@at.change(usec: 0), scheduled_at)
+
+            RSpec.warn_with((<<-WARNING).gsub(/^\s+\|/, '').chomp)
```
benoittgt, line 173: Is there a possibility that the user want to silence this warning?

lib/rspec/rails/matchers/active_job.rb
JonRowe: I thought about that, but you can copy paste the addition to your code, as its not necessarily going to be either of those values...

lib/rspec/rails/matchers/active_job.rb
JonRowe: There is, but this is a warning that is caused in a failing scenario only, if the times don't match at all you won't see it, only if the times match except in `usec` values should you see it.

I think this is niche enough to not need suppressing, as the solution to silencing it is to fix your passed in value.

lib/rspec/rails/matchers/active_job.rb
benoittgt: Yes completely. 
Maybe:
```suggestion
            |Use `change(usec: 0)` to correct these values. For example with current time: `Time.current.change(usec: 0)`
```

This is not a blocker for me. 😄

lib/rspec/rails/matchers/active_job.rb
benoittgt: Thanks :)

PR #2352, opened 2020-06-05
Combine scattered config

pirj:


pirj:
It was green, but Travis slow to report that.

PR #2353, opened 2020-06-07
Change the expectation for predicate methods

pirj:
Predicate methods typically return true or false, and should only exist for the uncountable and discrete, e.g. `use_transactional_fixtures?`, but not `max_formatted_output_length?`.

This relates to the changes made in https://github.com/rspec/rspec-core/pull/2736

### DIFF
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -1,3 +1,4 @@
+# rubocop: disable Metrics/ModuleLength
 module RSpec
   module Rails
     # Fake class to document RSpec Rails configuration options. In practice,
@@ -184,3 +185,4 @@ module RSpec
     initialize_configuration RSpec.configuration
   end
 end
+# rubocop: enable Metrics/ModuleLength
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -105,7 +105,35 @@ module RSpec
         end
 
         def render_views?
-          rendering_views
+          rendering_views?
+        end
+
+        def rendering_views?
+          !!rendering_views
+        end
+
+        # Define boolean predicates rather than relying on rspec-core due
+        # to the bug fix in rspec/rspec-core#2736, note some of these
+        # predicates are a bit nonsensical, but they exist for backwards
+        # compatibility, we can tidy these up in `rspec-rails` 5.
+        def fixture_path?
+          !!fixture_path
+        end
+
+        def global_fixtures?
+          !!global_fixtures
+        end
+
+        def infer_base_class_for_anonymous_controllers?
+          !!infer_base_class_for_anonymous_controllers
+        end
+
+        def use_instantiated_fixtures?
+          !!use_instantiated_fixtures
+        end
+
+        def use_transactional_fixtures?
+          !!use_transactional_fixtures
         end
 
         def infer_spec_type_from_file_location!
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -23,18 +23,28 @@ RSpec.describe "Configuration" do
     opts ||= {}
     default_value = opts[:default]
     alias_setting = opts[:alias_with]
-    query_method = "#{accessor}?".to_sym
+    predicate_method = "#{accessor}?".to_sym
     command_method = "#{accessor}=".to_sym
 
-    specify "`##{query_method}` is `#{default_value.inspect}` by default" do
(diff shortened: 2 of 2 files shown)


pirj:
I guess those two PRs will be red until any one of those PRs is merged. Unfortunately, our build system is not sophisticated enough to check out the "same branch" to test multi-repo changes.

pirj:
1. We seem to miss `include_examples "adds setting"` for `use_active_record` and `file_fixture_path`
2. Weird, what `fixture_path?`/`file_fixture_path?` are for? Maybe drop them?

Should I fix it here?

benoittgt:
Thanks @pirj 

For 1, I think it could be addressed in another PR
For 2, where do you see `fixture_path?`/`file_fixture_path?`?

pirj:
> For 2, where do you see `fixture_path?`/`file_fixture_path?`?

I don't see it used anywhere as well, but if you add this [here](https://github.com/rspec/rspec-rails/blob/e23efbb2cd6b40aaa8c6f33a36fe7ed4e724e2c7/spec/rspec/rails/fixture_support_spec.rb#L10):
```
expect(group).to respond_to(:fixture_path?)
```
it passes, same with `file_fixture_path?`.

JonRowe:
`rspec-rails` overrides `rspec-core` configuration, so this could be made to work, the question is should it.

lib/rspec/rails/configuration.rb
```diff
@@ -80,7 +81,8 @@ def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength
       config.include RSpec::Rails::FixtureSupport
 
       if ::Rails::VERSION::STRING > '5'
-        config.add_setting :file_fixture_path, default: 'spec/fixtures/files'
+        config.singleton_class.class_eval { attr_accessor :file_fixture_path }
```
pirj, line 84: There's `singleton_class` in Ruby 2.2.2, which is minimum requirement of Rails 5. But `attr_accessor` is still private there.

lib/rspec/rails/configuration.rb
```diff
@@ -67,7 +67,8 @@ def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength
       config.add_setting :use_transactional_fixtures, alias_with: :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
-      config.add_setting :fixture_path
+      (class << config; self; end).class_exec { attr_accessor :fixture_path }
```
pirj, line 70: There's singleton_class in Ruby 2.2. But attr_accessor is still private there.

lib/rspec/rails/configuration.rb
```diff
@@ -67,7 +67,8 @@ def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength
       config.add_setting :use_transactional_fixtures, alias_with: :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
-      config.add_setting :fixture_path
+      config.singleton_class.class_exec { attr_accessor :fixture_path }
```
pirj, line 70: There's `singleton_class` in Ruby 2.2, but `attr_accessor` is still private.

JonRowe:
@pirj I've taken this as base and improved it. Assuming it goes green we should merge it.

My reasoning here:
* The predicates already exist so can't be removed until `rspec-rails` 5
* We shouldn't depend on a specific version of rspec-core, so just fix it ourselves.

pirj:
Looks good! 👍

However, regarding backwards-compatibility. We're changing what those predicates return, don't we? Previously `file_fixture_path?` was returning a string, but now it's a true/false.

JonRowe:
> However, regarding backwards-compatibility. We're changing what those predicates return, don't we? Previously file_fixture_path? was returning a string, but now it's a true/false.

Yes which is my concern with rspec/rspec-core#2736 too but I've come to the decision this can be safely considered a bug fix, its not expected that predicate methods would be used except as truthy/falsely

JonRowe:
This is green on CI, just hasn't updated Github.

lib/rspec/rails/configuration.rb
```diff
@@ -105,7 +106,35 @@ def render_views
         end
 
         def render_views?
-          rendering_views
+          rendering_views?
+        end
```
pirj, line 110: `previous definition of use_transactional_fixtures? was here` in [`rspec-core` builds](https://travis-ci.org/github/rspec/rspec-core/jobs/695234603). We'll have to resort to `attr_accessor`.

PR #2354, opened 2020-06-10
Remove old predicates before redefining

JonRowe:
In an ideal world we'd not do this but because we are mixing into rspec-cores configuration this is preferable to inventing our own.


lib/rspec/rails/configuration.rb
```diff
@@ -117,22 +118,27 @@ def rendering_views?
         # to the bug fix in rspec/rspec-core#2736, note some of these
         # predicates are a bit nonsensical, but they exist for backwards
         # compatibility, we can tidy these up in `rspec-rails` 5.
+        undef :fixture_path? if respond_to?(:fixture_path?)
```
pirj, line 121: I'm on the fence between this and
```ruby
config.singleton_class.class_eval { attr_accessor :fixture_path }
# ...
def fixture_path?
  !!fixture_path
end
```

JonRowe: I'm not, I want to stick with `add_setting` as much as possible, these work arounds will be removed in the next version major version of rspec-rails.

PR #2356, opened 2020-06-17
Consistent request spec naming

### LINKED ISSUE #2355: Generated requests specs are not consistent with naming
We noticed while working on https://github.com/tpope/vim-rails/issues/368 , that generators does use a different naming scheme for requests specs.

For example, for a Posts Controller, the scaffold generator use [`spec/requests/posts_spec.rb`](https://github.com/rspec/rspec-rails/blob/master/spec/generators/rspec/scaffold/scaffold_generator_spec.rb#L9), while using the controller generator does create [`spec/requests/posts_request_spec.rb`](https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/spec/generators/rspec/controller/controller_generator_spec.rb#L9).

There is a shared example for requests specs [here](https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/spec/support/generators.rb#L46) that does use the first naming, that might be re-used everywhere to have a consistent naming scheme.

It seems like a minor issue, what are your opinions about this?

eloyesp:
Fixes #2355 by re-using the shared examples for "request specs generator" on controllers, that were used scaffold and requests generators.

It also make the generated code by controller generator a little bit better when no actions are added.


lib/generators/rspec/controller/controller_generator.rb
```diff
@@ -15,8 +17,9 @@ class ControllerGenerator < Base
       def generate_request_spec
         return unless options[:request_specs]
 
-        template 'request_spec.rb',
-                 File.join('spec/requests', class_path, "#{file_name}_request_spec.rb")
+        template_file = actions.empty? ? 'request_spec.rb' : 'request_with_actions_spec.rb'
```
JonRowe, line 20: You've renamed the file so where does the alternate come from? Any differences are also untested which I'd like to see rectified!

eloyesp: > You've renamed the file so where does the alternate come from?

The alternate template comes from the integration generator (added on the source_paths on line 7).

> Any differences are also untested which I'd like to see rectified!

It is reusing functionality of requests generator (and integration generator) because it is just re-using the template, but it is actually being tested on the [shared example](https://github.com/rspec/rspec-rails/pull/2356/commits/6cffc0549b6a7fa25ab2421f23c20d4da4c3daa0#diff-09c325a5eb95b7e1e766904a00127223R7).

Testing with that shared example we make sure that generators that generate request specs does use a consistent format (they all pass the same specs) while differences between specs are being also being tested.

eloyesp:
It seems that cucumber tests are failing because scaffold generator and controller generator are both trying to generate the same file (that is actually the expected behavior :) ) , but I'm not sure how to solve it.

lib/generators/rspec/controller/controller_generator.rb
pirj: Maybe generate `posts` with a scaffold, and generate `books` with controller generator?

JonRowe: You haven't added a test for the different behaviour, and there is different behaviour now.

eloyesp: @JonRowe I did add a shared example that it's testing just that new behavior. You can check that removing that code and specs will fail with meaningful errors.

The specs for that code are in [this support file](https://github.com/rspec/rspec-rails/blob/master/spec/support/generators.rb#L40). I'm just adding one line of test because I'm reusing existing specs.

Does it makes sense?

eloyesp: @JonRowe Runing the tests without the implementation changes gives those [rspec failures](https://gist.github.com/eloyesp/196cae8f0c36598609479d4c71441e70).

eloyesp:
Any update about this?

eloyesp:
@JonRowe @pirj Hi, can you review the changes, it seems to me that the change is simple, the only backward incompatible change is that when generating a controller the request spec generated will have a different file name and when it is generated without any action it will add just one spec for an index action with a message about adding a real spec.

All the changes are tested and the controller spec use the shared example for request specs (because it behaves as a request generator).

spec/generators/rspec/controller/controller_generator_spec.rb
```diff
@@ -4,9 +4,10 @@
 
 RSpec.describe Rspec::Generators::ControllerGenerator, type: :generator do
   setup_default_destination
+  it_behaves_like "a request spec generator"
```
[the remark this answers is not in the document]
klyonrad: Oh I see the other discussion now.... Not to crush your spirit here, but may I suggest to split the two behaviour changes (file naming & content of ControllerGenerator? The ControllerGenerator and the Request/Integration Generator are **not** the same thing.
There is no inheritance or module connection at all, sometimes repetition in tests is a good thing for understandability

ControllerGenerator executes also the rails generators, while the RequestGenerator **only** generates the test files (for example to test already existing untested controllers in a codebase)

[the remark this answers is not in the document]
eloyesp: > The ControllerGenerator and the Request/Integration Generator are not the same thing.

My idea here is not make them the same thing, but enforce them to behave as a "request generator". That way, given a certain configuration we can expect them to generate the same request specs.

> There is no inheritance or module connection at all, sometimes repetition in tests is a good thing for understandability.

In the case of the controller generator, there is no inheritance, but they share the same template file.

I prefer to use the shared example instead of repeat test, because I like the idea of make sure that this compatibility is maintained in new versions, so the file naming convention becomes a specification for requests generators.

JonRowe:
Is this replaced by #2376 can this one be closed?

klyonrad:
> Is this replaced by #2376 can this one be closed?

That was my intention, unless @eloyesp vetoes against that :)

eloyesp:
There is no need to close it manually, it should close automatically once #2376 is merged (that is the magic of the magic word), I don't see how could help manually closing it, but I'm not against it.

On the other hand, I do like the approach of @klyonrad of splitting the issue on two parts, but there are pending questions against the arguments from @JonRowe about doing both things on the same pull-request, so may be there is still something for me to learn from this.

JonRowe:
The changes to the generated file names will be merged from #2378 there is still an inconsistency here between the different generated request specs, integration generated request specs have a different content to controller generated request specs, I think the `integration/templates/request_spec.rb` is a better style than `controller/templates/request_spec.rb` and should probably be moved to the controller template, and then the integration generator updated to use that template so they have one file as a source of truth.

If you want to add more fleshing out of the actions similar to the scaffold (or prehaps reused) that could be looked at.

If either of you want to tackle that I'd be very happy, as I don't use generators.

klyonrad:
I can tackle it (after vacation though) since I am way too deep now anyway into these request spec generator issues 😉

JonRowe:
Thanks, our generators are a bit of a mess, I'd really like someone to do a "project" on cleaning them up

pirj:
It's getting better and better with each of your pull request guys, thanks!

klyonrad:
Ths was replaced by https://github.com/rspec/rspec-rails/pull/2378

PR #2357, opened 2020-06-21
Updates from rspec-dev (2020-06-21)

JonRowe:


benoittgt:
Do we want to partially use this update from `rspec-dev`? For example do not use the rubocop config.

JonRowe:
I'm closing this, but I would like to bring the script up to date so it updates the files that are shared, why is it deleting the base for example?

pirj:
> why is it deleting the base

We [don't inherit from base](https://github.com/rspec/rspec-rails/blob/master/.rubocop.yml#L1) anymore as [we do in other repos](https://github.com/rspec/rspec-core/blob/master/.rubocop.yml#L1).

JonRowe:
> We don't inherit from base anymore as we do in other repos.

That doesn't explain why the PR is deleting it, it suggests it should be adding it...

PR #2358, opened 2020-06-22
Backport config changes from 4-0-maintenance

JonRowe:
Backports the change the expectation for predicate methods so the build passes.


JonRowe:
Merged because build failure is pre existing, do you want to work on that @pirj? I can't get it to run locally due to weird rvm issues.

pirj:
I'll fix it.

PR #2359, opened 2020-06-24
No need to add rspec-rails to :development group

r7kamura:
Because rails generator works fine without it.

I confirmed by running the following command with rails 6.0.3.2 and rspec-rails 4.0.1:

```sh
rails new my_app
cd my_app
echo "gem 'rspec-rails', group: :test" >> Gemfile
bundle install
yarn install --check-files
rails g rspec:install
```


r7kamura:
I've noticed that other commands, such as `rails g model User`, won't generate expected test files if we remove rspec-rails from development gem group. So I'll close it.

PR #2272, opened 2020-01-31
Fix Rubocop offences

babrovka:

PR #2140, opened 2019-06-16
Update aruba to version 0.14.12

mvz:
- Bump Aruba version
- Force older sprockets on old Rubies
- Update overrides to work with Aruba 0.14.12

PR #2360, opened 2020-06-29
Fix 3-9-maintenance CI failures

pirj:
CI builds for `3-9-maintenance` branch fail.
Some failures are constant, like for earlier Rubies:
```
Warning: unrecognized cop BinaryOperatorParameterName found in /home/travis/build/rspec/rspec-rails/.rubocop.yml
```
or
```
ERROR:  While executing gem ... (Gem::InstallError)
    gem "bundler" cannot be uninstalled because it is a default gem
```
or
```
Bundler could not find compatible versions for gem "cucumber":
```
to a complete nonsense in newer Rubies
```
NameError:
  uninitialized constant ActionView::TestCase::Behavior
```
```
Failures:

  1) RSpec::Rails::ViewExampleGroup#view is accessible to hooks
     Failure/Error: expect(run_count).to eq 1

       expected: 1
            got: 0
```

Example builds: [[1](https://travis-ci.org/github/rspec/rspec-rails/builds/700803200), [2](https://travis-ci.org/github/rspec/rspec-rails/builds/703287126)].

- [ ] BONUS: forward-port select fixes to `master`/`4-0-maintenance`


JonRowe:
Its fine to bump Rubies to the latest versions permanently if you want

pirj:
I'm getting this when installing 2.2.10 with rvm, and remains in that condition, same error when running `gem install bundler` etc. I'll check other failures meanwhile.
```
/Users/pirj/.rvm/src/rubygems-3.0.8/lib/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- openssl (LoadError)
```

.travis.yml
```diff
@@ -38,57 +38,57 @@ dist: trusty
 matrix:
   include:
     # Rails dev / 6 builds >= 2.4.4
-    - rvm: 2.6.3
+    - rvm: 2.6.6
```
eloyesp, line 41: May I ask why not to use just `rvm: 2.6` that would require less manual updates. It seems [in the documentation](https://docs.travis-ci.com/user/languages/ruby/#specifying-ruby-versions-and-implementations) that it should work.

JonRowe: We have found that unreliable in the past, it works for some of the Rubies but can return older Rubies than the current release readily, and wasn't available for all Rubies.

pirj:
I'm getting this when installing 2.2.10 with rvm, and remains in that condition, same error when running `gem install bundler` etc. I'll check other failures meanwhile.
```
/Users/pirj/.rvm/src/rubygems-3.0.8/lib/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- openssl (LoadError)
```

rbenv claims this:
```
ruby-build installs a non-Homebrew OpenSSL for each Ruby version installed and these are never upgraded.

To link Rubies to Homebrew's OpenSSL 1.1 (which is upgraded) add the following
to your ~/.zshrc:
  export RUBY_CONFIGURE_OPTS="--with-openssl-dir=$(brew --prefix openssl@1.1)"

Note: this may interfere with building old versions of Ruby (e.g <2.4) that use
OpenSSL <1.1.
```
not sure if it's good or bad, but I'm going to give it a shot.

pirj:
RVM installs 2.3.8 with no issues.

Found [here](https://github.com/rvm/rvm/issues/4562#issuecomment-627403352):
```ruby
brew install rbenv/tap/openssl@1.0
PKG_CONFIG_PATH=/usr/local/Cellar/openssl@1.0/1.0.2t/lib/pkgconfig rvm install 2.2.10 --with-openssl-dir=/usr/local/Cellar/openssl@1.0/1.0.2t --with-openssl-lib=/usr/local/Cellar/openssl@1.0/1.0.2t/lib --with-openssl-include=/usr/local/Cellar/openssl@1.0/1.0.2t/include --rubygems ignore
```

JonRowe:
Yeah I had to use rvm's custom install for open ssl which installs to a special folder then use that

pirj:
Confused by [this example](https://github.com/rspec/rspec-rails/blob/c0a8c22186a1232e3b1612c44bc88052ec56c945/spec/rspec/rails/example/view_example_group_spec.rb#L250):
```ruby
      it 'is accessible to hooks', pending: pending_only_on_ruby_22_rails_52 do
        with_isolated_config do          run_count = 0
          RSpec.configuration.before(:each, type: :view) do
            allow(view).to receive(:render) { :value }
            run_count += 1
          end
            specify { expect(true).to eq true }
          end
          group.run(failure_reporter)
          expect(failure_reporter.exceptions).to eq []
          expect(run_count).to eq 1
        end
      end
```

Firstly, this spec passes as a whole, but if you focus on this example, it fails on all Ruby versions.
There's a `pending: ` on `master. Wondering how this focused failure might be related to:
```
        pending_only_on_ruby_22_rails_52 = """
          Rails 5.2.4.2 has a syntax error in ActionDispatch::Request::Session.
          (A &. usage which does not work in 2.2.10)
```
Also fails on latest released 5.2.4.3.

pirj:
Another confusion is (ok, solved) is related to our `RAILS_VERSION` env var. I assumed it works as `~> #{RAILS_VERSION}`. But a `RAILS_ENV=5.2` setting causes `rails` 5.2.0 to be installed.

It seems we're using `rails/rails`'s git branch, so `5-2-stable` should work. Except the following fails with **in some cases**:
```
  if RUBY_VERSION  >= "2.2"
    gem_list << 'puma' if version > '5-0-stable'
  end

  gem_list.each do |rails_gem|
    gem rails_gem, :git => "https://github.com/rails/rails.git", :branch => version
  end
```

```
Could not find gem 'puma' in https://github.com/rails/rails.git (at 5-2-stable@92fd1c2).
The source does not contain any versions of 'puma'
```

Hard to argue with that. And easy to fix.

I'm not sure if it's worth it to add a note on how to set `RAILS_VERSION` (there's nothing on `master` either), should be:
 - `5.2` means 5.2.0, from RubyGems
 - `'~> 5.2'` means latest 5.2.x
 - `5-2-stable`, `master` mean a branch from git

pirj:
The third confusion is this spec itself. I'm struggling to understand what does it test exactly, and I have no good idea.
In `master` we at least make an expectation that `(failure_reporter.exceptions).to eq []`, but it's not in `3-9-maintenance`.

pirj:
The fourth confusion is how it manages to work when it works, given:

```ruby
      def add_hook_to_existing_matching_groups(meta, scope, &block)
        if scope == :example || scope == :each || scope.nil?
          world.example_groups.each(&block)
```
So the trick is to add this hook to all known top-level-groups (wondering what the impact is for thousands of tl-groups and dozens of hooks).

So I thought that putting the group declaration before hook definition will work, but it doesn't, as it doesn't add a tl-group:

```ruby
RSpec::Core::ExampleGroup.describe('a view') { it { } }
RSpec.configuration.world.example_groups # => []
```

I'm running it on Ruby 2.5.1 and Rails 5.2.4.3.

pirj:
The fifth confusion is why it fails when focused, while passes just fine along with the whole spec.

pirj:
It seems that those two live in two different worlds:
```ruby
    2308:   if scope == :example || scope == :each || scope.nil?
    2309:     require 'pry'; binding.pry
 => 2310:     world.example_groups.each(&block)
    2311:   else
 [1] pry(#<RSpec::Core::Configuration>)> RSpec.world.example_groups
=> [RSpec::ExampleGroups::AView]
[2] pry(#<RSpec::Core::Configuration>)> world.example_groups
=> []
```
and `RSpec::Core::ExampleGroup.describe('a view') { it { } }` does actually add a top-level example group to `RSpec.world.example_groups`, but `RSpec.configuration.before` ignores that.

pirj:
Ok, I now understand the spec in question more or less.
It tells that a `view` should be available in hooks.
But why specifically global config hooks, what makes them different from regular hooks?

After fixing the issue with different worlds I'm still able to reproduce the focused failure.

pirj:
I guess the focusing issue is due to the example created with `RSpec::Core::ExampleGroup.describe` being filtered out from the spec run. No big deal if we don't run them one by one ourselves.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -219,18 +219,16 @@ def controller
...
-        with_isolated_config do
-          run_count = 0
-          RSpec.configuration.before(:each, :type => :view) do
-            allow(view).to receive(:a_stubbed_helper) { :value }
-            run_count += 1
-          end
-          group = RSpec::Core::ExampleGroup.describe 'a view', :type => :view do
-            specify { true }
-          end
-          group.run NullObject.new
```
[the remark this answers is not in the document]
pirj: Correction: keep `failure_reporter` in place in `master`.
Reason: https://github.com/rspec/rspec-rails/pull/2332#discussion_r426153128

pirj:
@JonRowe @benoittgt Welcome to the green festival! https://travis-ci.org/github/rspec/rspec-rails/builds/710525837
image

I've switched the Ruby 2.2.10 to Rails `5-2-stable` from `~> 5.2.4.3` because of the problem with the safe navigation operator being used even though it was only introduced in Ruby 2.3. `5-2-stable` never use `&.`, and there's no tag with a fix.
However, there's a difference:
```diff
Widget management
  User creates a new widget

-1 example, 0 failures, 1 pending
+1 example, 0 failures
```
it doesn't seem essential to me, and I intend to add this build to allowed failures.

Just in case, I'll drop Travis build cache to see how it's able to run the build without bundler's caches.

What are your thoughts on all of this, guys?

Rakefile
```diff
@@ -60,7 +60,7 @@ namespace :generate do
 
       # Rails 4 cannot use a `rails` binstub generated by Bundler
       sh "rm -f #{bindir}/rails"
-      sh "bundle exec rails new ./tmp/example_app --no-rc --skip-javascript --skip-sprockets --skip-git --skip-test-unit --skip-listen --skip-bundle --template=example_app_generator/generate_app.rb"
```
pirj, line 63: TODO: Forward-port

Rakefile
```diff
@@ -183,7 +183,7 @@ namespace :no_active_record do
 
         # Rails 4 cannot use a `rails` binstub generated by Bundler
         sh "rm -f #{bindir}/rails"
-        sh "bundle exec rails new #{example_app_dir} --no-rc --skip-active-record --skip-javascript --skip-bootsnap --skip-sprockets --skip-git --skip-test-unit --skip-listen --skip-bundle --template=example_app_generator/generate_app.rb"
```
pirj, line 186: TODO: Forward-port

.rubocop_rspec_base.yml
```diff
@@ -69,10 +69,6 @@ LineLength:
 MethodLength:
   Max: 15
 
-# Who cares what we call the argument for binary operator methods?
```
JonRowe, line 72: Are these just in line with 4-0?

pirj: No, those rules have diverged in https://github.com/rspec/rspec-rails/pull/2272, and we got rid of `.rubocop_rspec_base.yml` altogether when separated `rspec-rails` from the rest of RSpec.

spec/support/helpers.rb
```diff
@@ -4,6 +4,7 @@ module Helpers
   def with_isolated_config
     original_config = RSpec.configuration
     RSpec.configuration = RSpec::Core::Configuration.new
+    RSpec.configuration.world = RSpec.world
```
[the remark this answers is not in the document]
benoittgt: Why it was needed?

benoittgt:
> it doesn't seem essential to me

Me too.

Congrats for all the job. There is lot's of work. I will look at it again tomorrow but for the moment it looks good to me.

spec/support/helpers.rb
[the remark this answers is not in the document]
pirj: Otherwise, the two worlds are out of sync inside `with_isolated_config`:
```ruby
    with_isolated_config do
      RSpec::Core::ExampleGroup.describe('a view') { it { } }
      RSpec.configuration.world.example_groups # => []
    end
```
without `with_isolated_config`:

```ruby
      RSpec::Core::ExampleGroup.describe('a view') { it { } }
      RSpec.configuration.world.example_groups # => [<Example ...>]
```

I got rid of `with_isolated_config` in this example, so this change is to make `with_isolated_config` more future-proof.

.travis.yml
```diff
@@ -38,57 +38,61 @@ dist: trusty
...
-    - rvm: 2.3.7
+    - rvm: 2.3.8
       env: RAILS_VERSION='~> 5.2.0'
     - rvm: 2.2.10
-      env: RAILS_VERSION='~> 5.2.0'
+      # An incompatibility with Ruby 2.2 in Rails 5.2.4.0 (still not fixed 5.2.4.3)
+      # prevents from testing against this combination. This bug is not in 5-2-stable:
+      #   lib/action_dispatch/request/session.rb:96: syntax error, unexpected '.'
+      #             id&.public_id
+      env: RAILS_VERSION='5-2-stable'
```
JonRowe, line 62: On master we test both, and allow failure direct in the matrix.

rspec-rails.gemspec
```diff
@@ -51,6 +51,11 @@ Gem::Specification.new do |s|
   end
 
   s.add_development_dependency 'cucumber', '~> 1.3.5'
-  s.add_development_dependency 'aruba',    '~> 0.14.12'
+  if ENV['RAILS_VERSION'].gsub(/[^\d\.-]/, '').tr('-', '.') < '4.0'
+    # aruba > 0.9 depends on thor ~> 0.19, while railties < 4.0 depend on thor ~> 0.14.6
+    s.add_development_dependency 'aruba', '~> 0.5.4'
```
JonRowe, line 56: So why not use aruba 0.8?

spec/support/helpers.rb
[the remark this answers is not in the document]
JonRowe: If this doesn't fix a failing spec I'm going to say we should leave this as it was on this branch at least.

script/downgrade_bundler_on_old_rails
```diff
@@ -11,9 +11,11 @@ if ruby -e "exit(ENV['RAILS_VERSION'].scan(/\d+\.\d+.\d+/)[0].to_f < 5)"; then
 
   gem uninstall -aIx bundler || echo "Warning error occured removing bundler via gem"
 
+  echo "Does it fail to execute the script further here?"
+
   # this only works on Ruby 2.3 which is luckily the version we need to fix
   if ruby -e "exit(RUBY_VERSION.to_f == 2.3)"; then
-    rvm @global do gem uninstall -aIx bundler
+    rvm @global do gem uninstall -aIx bundler || echo "Warning: error occured removing bundler from the global RVM gemset"
```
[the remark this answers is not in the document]
JonRowe: I thought we removed / changed this on master?

JonRowe:
Good work, some cross feedback to do with #2362 but mostly minor points otherwise

.travis.yml
pirj: Do you suggest adding a job for `~> 5.2.0` here and allow its failure?

rspec-rails.gemspec
pirj: We [used to use](https://github.com/rspec/rspec-rails/pull/2140/files#diff-73c25cf975ccb4c89d8659202506c353L58) 0.5.4 before the update. Tried 0.8 and there were some failures here and there. Decided to stick with what used to work.

spec/support/helpers.rb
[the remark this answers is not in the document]
pirj: Removed. Let's see if it affects the build. At least it this change doesn't seem to affect `view_example_group_spec.rb` on 2.2.10/5.2.4.3.

pirj:
Ruby 1.9.2 segfaults time to time. Usually 
```
Fetching gem metadata from https://rubygems.org/../home/travis/.rvm/rubies/ruby-1.9.2-p330/lib/ruby/1.9.1/pathname.rb:300: [BUG] /home/travis/build/rspec/rspec-rails/script/travis_functions.sh: line 13:  7862 Segmentation fault      (core dumped) RUBYOPT=--disable=gem:' --enable rubygems' bundle install --gemfile ./Gemfile --path /home/travis/build/rspec/rspec-rails/../bundle --retry=3 --jobs=3

rake aborted!
```

```
Fetching gem metadata from https://rubygems.org/..../home/travis/.rvm/gems/ruby-1.9.2-p330/gems/bundler-1.17.3/lib/bundler/compact_index_client.rb:56: [BUG] Segmentation fault

ruby 1.9.2p330 (2014-08-07 revision 47094) [x86_64-linux]
```

Restarting usually helps. Maybe add it to allowed failures as well?
Otherwise all green.

JonRowe:
> Restarting usually helps. Maybe add it to allowed failures as well? Otherwise all green.

No because then we are not notified of legitimate failures...

rspec-rails.gemspec
JonRowe: Do want to update the comment to that effect?

pirj: Do you think a comment is really necessary?
It's in the blame https://github.com/rspec/rspec-rails/blame/a8dd5e1764b7142e67dc7903842f332e1a4a7daa/rspec-rails.gemspec#L56 https://github.com/rspec/rspec-rails/commit/5781ea1be5c474d6cd47bb2a9d12d106044c7e57

>  Use the previous aruba for older Rails

> https://github.com/rspec/rspec-rails/pull/2140/files#diff-73c25cf975ccb4c89d8659202506c353L58 bumped aruba from 0.5.4 straight to 0.14.12. Using the version that used to work for older rails.

JonRowe: The comment above it is misleading currently.

rspec-rails.gemspec
```diff
@@ -51,6 +51,11 @@ Gem::Specification.new do |s|
   end
 
   s.add_development_dependency 'cucumber', '~> 1.3.5'
-  s.add_development_dependency 'aruba',    '~> 0.14.12'
+  if ENV['RAILS_VERSION'].gsub(/[^\d\.-]/, '').tr('-', '.') < '4.0'
+    # aruba > 0.9 depends on thor ~> 0.19, while railties < 4.0 depend on thor ~> 0.14.6
```
JonRowe, line 55:
```suggestion
    # We use aruba `~> 0.5.4` as the last known working version and later versions (> 0.9) depend on thor ~> 0.19, which is incompatible with railties < 4.0.
```

pirj: Much better, thanks!

PR #2362, opened 2020-07-25
Minor improvements

pirj:
Follow-up to #2360

Explanations in commit messages.


spec/support/helpers.rb
```diff
@@ -4,6 +4,7 @@ module Helpers
   def with_isolated_config
     original_config = RSpec.configuration
     RSpec.configuration = RSpec::Core::Configuration.new
+    RSpec.configuration.world = RSpec.world
```
JonRowe, line 7: The whole point of isolated configuration is to be isolated from the global test run, we don't want the child configuration sharing the parent world, else those specs get counted in various places. What were you seeing that made you want to do this?

pirj: Sorry, I disagree. `with_isolated_config` sounds like the configuration is isolated, not that what is inside the block should be in a different world.

[We wrap each example](https://github.com/rspec/rspec-rails/blob/71c43641759b41ec26e7e8b85d285c20f1eabb14/spec/spec_helper.rb#L54) to run in its own world:
```ruby

    real_world = RSpec.world
    example.run
  end
```

Also, all current usages of `with_isolated_config` wrap the whole example it is used in, there are no separate "inside" and "outside" example parts that would be prone to affect each other.

WDYT of making `with_isolated_config` a metadata-induced `around` hook to make this more obvious?

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -248,19 +248,17 @@ def _view; end # Stub method
...
-            allow(view).to receive(:render) { :value }
-            run_count += 1
-          end
-          group = RSpec::Core::ExampleGroup.describe 'a view', type: :view do
-            specify { expect(true).to eq true }
-          end
-          group.run(failure_reporter)
-          expect(failure_reporter.exceptions).to eq []
-          expect(run_count).to eq 1
+        group = RSpec::Core::ExampleGroup.describe 'a view', type: :view do
```
JonRowe, line 251: Can you restore this, it was a regression check specifically for configuration hooks added (7 years ago!!) in #833

spec/support/helpers.rb
JonRowe: Wow I had no idea we did that, ok, well you need that world, I don't think you can rely on the global here...

pirj: So it seems that the real problem, that prevents us from using a unique isolated configuration is [in `RSpec::Core::ExampleGroup.describe`](https://github.com/rspec/rspec-core/blob/a01aba79715dda8ede1bdb7d1c5611094939fc1d/lib/rspec/core/example_group.rb#L260):
```ruby
RSpec.world.example_groups
```

Should it be a `RSpec.configuration.world.example_groups`?

This results in:

image

> I don't think you can rely on the global here...

The whole point is that I don't rely on the global, real one. The global is kept in `real_world` during the examples, and `RSpec.world` is assigned (using `instance_variable_set`) a `World.new`.
But `RSpec::Core::Configuration.new` gets its own world:
```ruby
  def with_isolated_config
```
which is an `RSpec::Core::World::Null`.

I strongly doubt it would make sense to change the Core to:
```diff
- RSpec.world.example_groups
+ RSpec.configuration.world.example_groups
```

But in this case, setting an isolated world initialized specifically for this example to an isolated config that was also initialized for this example makes total sense.

spec/rspec/rails/example/view_example_group_spec.rb
pirj: Aha, got it 👍

JonRowe: Happy for the doc string / or a comment to be changed to mention this though, as I had to check to find it :joy:

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -248,19 +248,17 @@ def _view; end # Stub method
...
+          specify { true }
+        end
+        run_count = 0
+        group.before(:each) do
+          allow(view).to receive(:render) { :value }
+          run_count += 1
         end
+        group.run(failure_reporter)
+        expect(failure_reporter.exceptions).to eq []
```
pirj, line 260: By the way, if I add some nonsense to the hook, `failure_reporter` doesn't seem to care:
```ruby
          RSpec.configuration do
            before(:each) do
              allow(sl).to receive(:renddlfkr) { :value }
              run_count += 1
            end
          end
```
and `        expect(failure_reporter.exceptions).to eq []` is passing just fine.

So this trick doesn't prevent us from not having access to the `view`.

Does it make sense to keep it anyway @JonRowe ?

pirj: Taking my words back.
If you comment out the whole content of `def setup` method in SetupAdnTeardownAdapter you'll get:
```diff
       -[]
       +[#<NoMethodError: undefined method `_helpers' for nil:NilClass>]
```
failure, which is much more explanatory than a `0` vs `1` failure messge.

JonRowe: If you add a `fail` to the hook in the example you'll get a `RuntimeError` in the array etc, a nonsense allow statement is not an error of course, as it just allows something to be called.

It does highlight that the spec is a bit strange, the allow is not itself used, but instead is demonstrating that you can access the view in the hook, this could be improved by changing to this:

```
      # This is a regression test from rspec/rspec-rails#833
      it 'is accessible to configuration hooks', pending: pending_only_on_ruby_22_rails_52 do
        with_isolated_config do
          run_count = 0
          RSpec.configuration.before(:each, type: :view) do
            # `view` is provided from the view example type, and serves to demonstrate this
            # hook is run in the correct context.
            expect(view).to receive(:render) { :value }
            run_count += 1
          end
            specify { expect(view.render).to eq :value }
          end
          group.run(failure_reporter)
          expect(failure_reporter.exceptions).to eq []
          expect(run_count).to eq 1
        end
      end
```

pirj: This is exactly what I've ended up with. But thanks for you time anyway!

spec/support/helpers.rb
JonRowe: Can you drop this change for now, and open another PR to work on this? I'm happy to merge the rest of these changes, but I'd like to look at this a bit more

pirj: Makes sense. Done.

PR #2365, opened 2020-07-27
Get rid of the code poking around RSpec's internals

pirj:
## Problem

We used to access and replace `RSpec.configuration` and `RSpec.world` in several places.
This was prone to error, since those two can run out of sync, e.g.:

    with_isolated_config do
      RSpec::Core::ExampleGroup.describe('a view') { it { } }
      RSpec.configuration.world.example_groups # => []
    end

This case specifically is due to [in `RSpec::Core::ExampleGroup.describe`](https://github.com/rspec/rspec-core/blob/a01aba79715dda8ede1bdb7d1c5611094939fc1d/lib/rspec/core/example_group.rb#L260):
```ruby
RSpec.world.example_groups
```

Also, we were using this:

```ruby
    real_world = RSpec.world
    example.run
  end
```

`RSpec::Core::Configuration.new` gets its own world:
```ruby
  def with_isolated_config
```
which is an `RSpec::Core::World::Null`.

I have strong doubts if it would make sense to change the Core to:
```diff
- RSpec.world.example_groups
+ RSpec.configuration.world.example_groups
```

Not to say that those hooks and helpers dig too deep into RSpec's internals.

## Solution

It's our lucky day! RSpec [provides sandboxing out of the box](https://github.com/rspec/rspec-core/blob/master/lib/rspec/core/sandbox.rb).


JonRowe:
So we shouldn't be doing this:

```
  real_world = RSpec.world
  example.run
end
```

pirj:
I think I've found a better solution.

pirj:
Better viewed as https://github.com/rspec/rspec-rails/pull/2365/files?diff=split&w=1

spec/spec_helper.rb
```diff
@@ -51,10 +52,18 @@ def self.run_all(reporter = nil)
...
   config.around(:example) do |example|
-    real_world = RSpec.world
-    RSpec.instance_variable_set(:@world, RSpec::Core::World.new)
-    example.run
-    RSpec.instance_variable_set(:@world, real_world)
+    RSpec::Core::Sandbox.sandboxed do |sandbox_config|
+      # If there is an example-within-an-example, we want to make sure the inner
+      # example does not get a reference to the outer example (the real spec) if
+      # it calls something like `pending`.
+      sandbox_config.before(:context) { RSpec.current_example = nil }
```
pirj, line 62: This actually didn't affect the build.
Spotted this as an example to `sandboxed do` and in `rspec-core`'s `spec_helper.rb`. Adding just in case.

I think resetting `current_example` should be part of `sandboxed`, as this `current_example` is global. How does it play with threaded RSpec at all?

spec/spec_helper.rb
```diff
@@ -51,10 +52,18 @@ def self.run_all(reporter = nil)
...
+      RSpec::Rails.initialize_configuration(sandbox_config)
+      example.run
+    end
   end
+
+  config.include RSpec::Rails::FeatureCheck
```
pirj, line 68: Was included in `Helpers` that didn't hesitate to include itself in `RSpec.configuration`.

spec/spec_helper.rb
benoittgt: I tried to understand why we need this line without finding. 

What will be a scenario where this line could be needed?

I tried:
```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rspec", "~> 3.8"
end

require 'rspec/autorun'
require 'rspec/core/sandbox'

RSpec.configure do |config|
  config.formatter = :documentation
    RSpec::Core::Sandbox.sandboxed do |sandbox_config|
      # disable or not this is the same, this is never triggered
      sandbox_config.before(:context) { RSpec.current_example = nil }
      example.run
    end
  end
end

RSpec.describe do
  it do
    group = RSpec::Core::ExampleGroup.describe("an example group")
    example = group.example("an example")

    RSpec.current_example = example
    expect(RSpec.current_example).to be(example)
  end
end
```

pirj: Tried defining several groups, running them, adding examples and groups to the host spec, couldn't make it fail either way with this hook commented out. 🤷‍♂️ 

I have no hard preference to keep or to remove it since it's only for our specs and it won't affect anyone.

We use `current_example` in [`lib/rspec/rails/example/system_example_group.rb`](https://github.com/rspec/rspec-rails/blob/60dc9c4ebdc52f408ba9345089f884897df8deb1/lib/rspec/rails/example/system_example_group.rb#L30), but no specs fail if the hook is removed.

Considering `current_example`'s doc:
>  # The primary audience for this method is library authors who need access
>  # to the example currently being executed and also want to support all
>  # versions of RSpec 2 and 3.

I guess some third-party libraries use it: a [quick search](https://github.com/search?l=Ruby&q=RSpec.current_example&type=Code).

`RSpec.current_examples` was originally introduced here https://github.com/rspec/rspec-core/pull/1004

JonRowe: We need to keep `RSpec.current_example` its public API and it is used.

This line stems from `rspec-cores` test suite which does fail without it, specifically, things leak, here's a standalone reproduction:

```
# rspec_sandbox.rb
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "rspec", "~> 3.8"
end

require 'rspec/autorun'
require 'rspec/core/sandbox'

RSpec.configure do |config|
  config.formatter = :documentation
    RSpec::Core::Sandbox.sandboxed do |sandbox_config|
      sandbox_config.before(:context) do
        if ENV["NO_RESET"].nil? == true
          RSpec.current_example = nil
        end
      end
      example.run
    end
  end
end

class RaiseOnFailuresReporter < RSpec::Core::NullReporter
  def self.example_failed(example)
    raise example.exception
  end
end

RSpec.describe "RSpec.current_example" do
  it do
    group = RSpec::Core::ExampleGroup.describe("an example group")
    group.before(:context) { skip("not done"); fail }
    group.example("an example 1") {}
    group.example("an example 2") {}

    expect(group.run RaiseOnFailuresReporter).to eq true

    expect(group.examples.first).to be_skipped
    expect(group.examples.last).to be_skipped
  end
end
```

Running this without the env var set causes the spec to pass (bundler output removed, and ignore the generated title).
```
ruby sandbox.rb
RSpec.current_example
  is expected to be skipped

Finished in 0.00421 seconds (files took 0.07455 seconds to load)
1 example, 0 failures
```

Setting the env var to skip the reset will cause the skip to leak to the top level
```
NO_RESET=1 ruby sandbox.rb

RSpec.current_example
  is expected to be skipped (FAILED - 1)

Failures:

  1) RSpec.current_example is expected to be skipped FIXED
     Expected pending 'not done' to fail. No error was raised.
     # sandbox.rb:34

Finished in 0.00448 seconds (files took 0.08593 seconds to load)
1 example, 1 failure

Failed examples:

rspec sandbox.rb:34 # RSpec.current_example is expected to be skipped
```

spec/spec_helper.rb
pirj: 36 failures, 100% in generator specs, all `undefined method `type_metatag'`.
Besides `type_metatag`, generator specs use no methods from `FeatureCheck`.

I'll include this module exclusively to generator specs.

Frankly, I don't see a good reason for this method to exist:
```ruby
      def type_metatag(type)
        "type: :#{type}"
      end
```
but I'd prefer to avoid this unrelated refactoring here.

spec/spec_helper.rb
pirj: It's more or less clear.
So we're keeping this line here as well, right?

JonRowe: Yes

spec/spec_helper.rb
```diff
@@ -51,10 +52,18 @@ def self.run_all(reporter = nil)
...
+      # If there is an example-within-an-example, we want to make sure the inner
+      # example does not get a reference to the outer example (the real spec) if
+      # it calls something like `pending`.
+      sandbox_config.before(:context) { RSpec.current_example = nil }
+      RSpec::Rails.initialize_configuration(sandbox_config)
+      example.run
+    end
   end
+
+  config.include RSpec::Rails::FeatureCheck, type: :generator
```
JonRowe, line 68: Can we not just include this where it's used?

pirj: We can, but `type_metatag` is used in every one of them except for install generator spec.

pirj: I intended to get rid of this method altogether, replacing all usages to `type: :some_type`, and this `include` as well.
Ok to keep it here and remove in a follow-up?
Or keep `type_metatag` and inline `include`s into generator specs?
Or keep this include as is and leave `type_metatag` alone?
I don't have a preference here, up to you.

JonRowe: So you can remove this from here, and add it here:

```
module RSpec
  module Rails
    module Specs
      module Generators
        module Macros
        # spec/support/generators.rb:18
        def self.included(klass)
          klass.extend(Macros)
          klass.include(RSpec::Rails::FeatureCheck)
        end
```

pirj: Good point. Done.

PR #2366, opened 2020-07-30
Lock gems that fail to install on certain Ruby versions

pirj:
Bundler fails to install several gems. It seems that Ruby version constraint is ignored.

E.g. Sprockets 4.0 depend on Ruby >= 2.5 https://rubygems.org/gems/sprockets/versions/4.0.0
This results in build failures (https://travis-ci.org/github/rspec/rspec-rails/jobs/713253057):

    Installing sprockets 4.0.2
    Gem::RuntimeRequirementNotMetError: sprockets requires Ruby version >= 2.5.0.
    The current ruby version is 2.4.10.364.

Surprisingly, Ruby 2.3.8 and Ruby 2.2.10 builds don't fail, and install sprockets 3.7.2.

This is a forgotten forward-port from https://github.com/rspec/rspec-rails/pull/2360/files#diff-8c3955be04b733dbd31e54e50844fb84R31

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -35,11 +35,22 @@ in_root do
     gsub_file "Gemfile", /.*rails-controller-testing.*/, "gem 'rails-controller-testing', git: 'https://github.com/rails/rails-controller-testing'"
   end
 
+  if RUBY_VERSION < "2.3.0"
+    gsub_file "Gemfile", /.*childprocess.*/, "gem 'childprocess', '< 2.0.0'"
+    gsub_file "Gemfile", /.*i18n.*/, "gem 'i18n', '< 1.5.2'"
+    gsub_file "Gemfile", /.*nio4r.*/, "gem 'nio4r', '< 2.4.0'"
+    gsub_file "Gemfile", /.*public_suffix.*/, "gem 'public_suffix', '< 4.0.0'"
+    gsub_file "Gemfile", /.*rack.*/, "gem 'rack', '< 2.2.0', '!= 2.1.0'"
+    gsub_file "Gemfile", /.*xpath.*/, "gem 'xpath', '< 3.2.0'"
+  end
+
   if Rails::VERSION::STRING >= "5.1.0"
     if RUBY_VERSION < "2.3"
       gsub_file "Gemfile", /.*capybara.*/, "gem 'capybara', '~> 3.1.0'"
     elsif RUBY_VERSION < "2.4"
       gsub_file "Gemfile", /.*capybara.*/, "gem 'capybara', '~> 3.15.0'"
+    elsif RUBY_VERSION < "2.5"
+      gsub_file "Gemfile", /.*capybara.*/, "gem 'capybara', '~> 3.32.0'"
     end
     if Rails::VERSION::STRING >= "5.2.0"
       gsub_file "Gemfile", /.*chromedriver-helper.*/, "gem 'webdrivers', '< 4.0.0'"
--- a/Gemfile
+++ b/Gemfile
@@ -49,8 +49,12 @@ gem 'rake', '~> 12'
 
 gem 'mime-types', "~> 3"
 
-if RUBY_VERSION.to_f == 2.2
+if RUBY_VERSION.to_f < 2.3
   gem 'capybara', '~> 3.1.0'
+elsif RUBY_VERSION.to_f < 2.4
+  gem 'capybara', '< 3.16'
+elsif RUBY_VERSION.to_f < 2.5
+  gem 'capybara', '< 3.33'
 else
   gem 'capybara', '>= 2.13', '< 4.0', require: false
 end
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -18,6 +18,7 @@ when /stable$/
   gem_list = %w[rails railties actionmailer actionpack activerecord activesupport activejob actionview]
   gem 'puma', "3.12.1" if version > '5-0-stable'
   gem 'activerecord-jdbcsqlite3-adapter', git: 'https://github.com/jruby/activerecord-jdbc-adapter', platforms: [:jruby]
+  gem "sprockets", '~> 3.0' if RUBY_VERSION < '2.5'
 
   gem_list.each do |rails_gem|
     gem rails_gem, :git => "https://github.com/rails/rails.git", :branch => version
@@ -28,7 +29,7 @@ when nil, false, ""
   gem 'activerecord-jdbcsqlite3-adapter', platforms: [:jruby]
 else
   gem "rails", version
-
+  gem "sprockets", '~> 3.0' if RUBY_VERSION < '2.5'
   gem "puma" if version >= '5-1-stable'
 
   if version.gsub(/[^\d\.]/,'').to_f >= 6.0


pirj:
It's like a house of cards.
```
Fetching capybara 3.33.0
Installing capybara 3.33.0

Gem::RuntimeRequirementNotMetError: capybara requires Ruby version >= 2.5.0. The current ruby version is 2.4.10.364.

An error occurred while installing capybara (3.33.0), and Bundler cannot continue.

Make sure that `gem install capybara -v '3.33.0' --source 'https://rubygems.org/'` succeeds before bundling.
```

It also happens quite randomly. This time Ruby 2.4 build broke, but not 2.3 nor 2.2.

Another thing that I've noticed is that Ruby version constraints for some gems differ when you check them on https://rubygems.org/gems/<gem>/versions/<version> and then you check it with `gem specification --remote <gem> --version <version>`. Don't have any example handy, but pretty sure I can find it.

Any newly released gem with a Ruby version constraint can break our build. Do you think those problems are worth Bundler core team attention?

pirj:
Builds are green. I'm going to delete all repository caches, squash the commits and see if the builds are still green.

example_app_generator/generate_app.rb
```diff
@@ -26,20 +26,29 @@
   gsub_file "Gemfile", /.*gem..sqlite3.*/, "gem 'sqlite3', '~> 1.3.6'"
   gsub_file "Gemfile", /.*bootsnap.*/, ""
 
-  if Rails::VERSION::STRING >= '5.0.0'
```
JonRowe, line 29: Leave this, we're still soft supporting 4.2 and this makes it tricky to properly install locally.

JonRowe:
👍  Merge on green.

benoittgt:
> Do you think those problems are worth Bundler core team attention?

Maybe @deivid-rodriguez have an idea? 🙂

pirj:
This is green, but the question still remains.
@deivid-rodriguez can you suggest what we are doing wrong here?
We run several build jobs, each for a different Ruby/Rails combination. The Gemfile we provide, doesn't specify a version of e.g. `i18n`, `rack`, `childprocess`, `capybara` to use, it's what we rely on from Rails' `Gemfile`. On [Ruby 2.4 and RAILS_VERSION='~> 5.0.0' (resolved to 5.0.7.2)](https://travis-ci.org/github/rspec/rspec-rails/jobs/713253057), `bundler` attempts to install `sprockets` version 4.0.2 for some reason:
```
Installing sprockets 4.0.2

Gem::RuntimeRequirementNotMetError: sprockets requires Ruby version >= 2.5.0. The current ruby version is 2.4.10.364.
An error occurred while installing sprockets (4.0.2), and Bundler cannot continue.
Make sure that `gem install sprockets -v '4.0.2' --source 'https://rubygems.org/'` succeeds before bundling.

In Gemfile:

  rails was resolved to 5.0.7.2, which depends on
    sprockets-rails was resolved to 3.2.1, which depends on
      sprockets

rake aborted!
```

[`sprocket-rails` 3.2.1 depend on `sprockets >= 3.0.0`](https://rubygems.org/gems/sprockets-rails/versions/3.2.1), and [`sprockets` 3.7.2 depend on Ruby >= 1.9.3](https://rubygems.org/gems/sprockets/versions/3.7.2).

*Question*: Is it Bundler that is unable to reduce the dependency considering the Ruby version constraint? Or should we look into what Travis' intermediate gem server is providing us with broken gem metadata?

The following versions of RubyGems/Bundler are used:
```
Bundler 2.1.4 installed
RubyGems 3.1.4 installed
```

deivid-rodriguez:
Hi! Thanks for letting me know.

This is indeed an issue with `bundler`. A long time ago, bundler used a dependency API that didn't have `required_ruby_version` information. These days bundler uses a better API that _does_ provide `required_ruby_version` information. However, under some circumstances, specifically, when the client is rate limited by rubygems.org because of too many concurrent requests, we still end up falling back to using the old API, causing this kind of intermittent problems.

I think we had relaxed the limits on the server side to minimize these issues, but the right fix is that `bundler` completely stops using the old API and instead waits a bit before retrying the new API again when being rate limited. This fix is being tracked at https://github.com/rubygems/rubygems/issues/3594.

pirj:
Thanks a lot @deivid-rodriguez, this makes total sense!

deivid-rodriguez:
No problem, and apologies for the trouble. We'll try to fix this.

benoittgt:
Ooooh good to know! Thank you very much!

On August 1, 2020 12:16:15 PM GMT+02:00, "David Rodríguez" <notifications@github.com> wrote:
>Hi! Thanks for letting me know.
>
>This is indeed an issue with `bundler`. A long time ago, bundler used a
>dependency API that didn't have `required_ruby_version` information.
>These days bundler uses a better API that _does_ provide
>`required_ruby_version` information. However, under some circumstances,
>specifically, when the client is rate limited by rubygems.org because
>of too many concurrent requests, we still end up falling back to using
>the old API, causing this kind of intermittent problems.
>
>I think we had relaxed the limits on the server side to minimize these
>issues, but the right fix is that `bundler` completely stops using the
>old API and instead waits a bit before retrying the new API again when
>being rate limited. This fix is being tracked at
>https://github.com/rubygems/rubygems/issues/3594. 
>
>-- 
>You are receiving this because your review was requested.
>Reply to this email directly or view it on GitHub:
>https://github.com/rspec/rspec-rails/pull/2366#issuecomment-667508360

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

deivid-rodriguez:
Hei! Discussing this internally, this might actually not be due to rate limiting but to a different error condition. Bundler should log the exception that's causing the fallback to the old API in `--verbose` mode [here](https://github.com/rubygems/rubygems/blob/5b9a5692793b1c523493c9574d6569299903f540/bundler/lib/bundler/fetcher/compact_index.rb#L23), I believe. If you could use the `--verbose` flag and report back when this happens again, that would give us more information about this.

pirj:
@deivid-rodriguez Awesome!
Got some results ([1](https://travis-ci.org/github/rspec/rspec-rails/jobs/714297172), [2](https://travis-ci.org/github/rspec/rspec-rails/jobs/714297185)) already. Please let me know if this is sufficient to pinpoint the issue.

Just to confirm that we're still experiencing the issue there's a similar failure from [another build](https://travis-ci.org/github/rspec/rspec-rails/jobs/714239589) for a [branch](https://github.com/rspec/rspec-rails/blob/2c23692be7638f4d4721e8161d01e12045c47505/example_app_generator/generate_app.rb#L40) where:
```
if RUBY_VERSION < "2.3.0"
    gem 'i18n', '< 1.5.2'
```
is supposedly set, and the failure is:
```
Gem::RuntimeRequirementNotMetError: i18n requires Ruby version >= 2.3.0. The current ruby version is 2.2.0.
An error occurred while installing i18n (1.8.5), and Bundler cannot continue.
```

pirj:
@deivid-rodriguez Ping. Is this of any help?
No pressure, we're in pretty good shape in terms of build stability right now. At the cost, though.
For me, it will be good enough if you are aware of the problem, there's a ticket, and it will eventually be resolved so we can clean our build-related dependency declaration code.

pirj:
@deivid-rodriguez Ping. Can I provide anything else to help to pinpoint the issue?

pirj:
@deivid-rodriguez Ping.

sonalkr132:
> Gem::RuntimeRequirementNotMetError: i18n requires Ruby version >= 2.3.0. The current ruby version is 2.2.0.
An error occurred while installing i18n (1.8.5), and Bundler cannot continue.

This should no longer be an issue. I ran build on the same commit and it is passing https://travis-ci.com/github/sonalkr132/rspec-rails/jobs/379918197

> they show a Net::OpenTimeout error. 

this was probably an intermittent error and bundler continued with retry.
we backfilled `required_ruby_version` on older versions (16-17 Aug) after you reported https://github.com/rubygems/rubygems/issues/3870 which would have fixed the mismatch in the required ruby version.

deivid-rodriguez:
Oh, right! This was fixed by backfilling those, I forgot about it!

PR #2367, opened 2020-08-01
Fix link in docs for be_a_new_matcher

mehagar:


JonRowe:
:wave: Thanks for spotting this, but I'm not sure this is anymore valid? The feature file is `matchers/new_record_matcher`

mehagar:
I wasn't sure whether the hyperlink should be based on the feature file, since I found it through the Cucumber docs and the url for that is https://relishapp.com/rspec/rspec-rails/v/4-0/docs/matchers/be-a-new-matcher

JonRowe:
Huh, yeah ok, seems to work! Thank you!

PR #2370, opened 2020-08-07
Remove old fixture_path support

JonRowe:
An alternative to #2349 which removes the old work around, this seems to no longer apply given our supported levels of Rails.


JonRowe:
@koenpunt can you try this branch out?

sudara:
I just wanted to note I tried this out after having issues on Rails edge and confirmed it's happy for me, in tandem with #2385

JonRowe:
Thanks @sudara

PR #2372, opened 2020-08-20
Add support for Rails engines routes in request specs

drwl:
Currently, when scaffolds are generated inside a Rails Engine, it generates requests specs that fail without modification. This happens because the Engine routes are not mounted under Rails.application.

To fix this, Engine routes have to be included so that url and path helpers can be usable from within the spec file.


pirj:
I guess we'll need a spec for that.

lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -16,6 +16,11 @@
 RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %> do
   # <%= class_name %>. As you add validations to <%= class_name %>, be sure to
   # adjust the attributes here as well.
+  <% if mountable_engine? -%>
```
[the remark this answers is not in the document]
JonRowe: I agree with @klyonrad here, I would put this new block above the comment, it also needs to be in all the request spec templates.

lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -16,6 +16,11 @@
...
+    include Engine.routes.url_helpers
```
JonRowe, line 20: I think this needs to be fully qualified? Not quite sure how you'd get that, maybe we can comment it out with a note if we can't (so the spec will run, but need editing)

JonRowe: I think we want something like `<%= namespace %>::Engine`

JonRowe:
Great start thanks for tackling, we usually test generators by running them in the example app, tweaking with removing comments etc and then letting them pass

drwl:
Thanks for the feedback everyone. I was trying to get a reproducable test case so have been working on that.

lib/generators/rspec/scaffold/templates/request_spec.rb
drwl: @JonRowe does it need to be fully qualified since it's already namespaced in the module on L15? I'm not opposed to adding it, just curious.

lib/generators/rspec/scaffold/templates/request_spec.rb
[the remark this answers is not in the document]
drwl: Thanks @klyonrad @JonRowe, I fixed this 👍 .

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
+Feature: engine named routes
```
JonRowe, line 1: I would add this as scenarios to the existing request specs feature.

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
...
+
+  Generated requeset specs inside a Rails Engine is able to use route and url helpers.
```
JonRowe, line 3: This has a typo, but I'd merge this into the main request specs feature more like 
```suggestion
  Request specs inside a Rails Engine can include the engines routes to use url helpers.
```

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
...
+
+  Scenario: include engine route helpers
```
JonRowe, line 5:
```suggestion
  Scenario: using engine route helpers
```

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
...
+        resources :links, :only => [:index]
+      end
+
+      Rails.application.routes.draw do
+        mount MyEngine::Engine => "/my_engine"
+      end
+
+      module MyEngine
+        RSpec.describe "Links", :type => :request do
+
```
JonRowe, line 33: I think most people don't include blank lines here so the generator shouldn't either.

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
...
+            get links_url
+            expect(response.body).to eq('hit_engine_route')
+          end
+        end
+      end
+      """
+    When I run `rspec spec`
+    Then the example should pass
+
+  Scenario: do not include engine route helpers
```
JonRowe, line 46: You don't need the sad path here, its documentation of how it works not how it doesn't :)

features/request_specs/engine_named_routes.feature
```diff
@@ -0,0 +1,83 @@
...
+
+          it "redirects to a random widget" do
+            get links_url
+            expect(response.body).to eq('hit_engine_route')
+          end
+        end
+      end
+      """
+    When I run `rspec spec`
+    Then the example should fail
```
JonRowe, line 83: This file needs a line ending.

lib/generators/rspec/scaffold/templates/request_spec.rb
JonRowe: Hm I guess not but its more clear.

JonRowe (changes requested):
I like the additional documentation, but it doesn't test the generator, that still needs to be added ideally.

drwl:
@JonRowe I've added a test to make sure the route helpers get included. Wanted to get your thoughts on the namespaced include engine route/url helpers -- it looks like Rails generators does not do that, so does it make sense to keep it similar?

JonRowe (approved):

pirj:
Thanks!

PR #2378, opened 2020-08-26
Consistent request spec naming

JonRowe:
Bare minimum changes required to get consistent spec naming, replaces #2376


JonRowe:
Thanks for getting this over the line @klyonrad

example_app_generator/generate_stuff.rb
```diff
@@ -85,7 +85,10 @@ def using_source_path(path)
 generate('rspec:install')
 generate('controller wombats index') # plural
 generate('controller welcome index') # singular
-generate('rspec:request wombats')
+
+# request specs are now the default
+generate('rspec:controller wombats --no-request-specs --controller-specs --no-view-specs')
```
klyonrad, line 90: Funny, I didn't know that about this generator. But to be honest I don't understand the purpose of this line 🤔 

One more testing of controller specs? I thought the whole issue was about request specs

klyonrad (commented):
Also thanks for finishing this up. Got one question though

example_app_generator/generate_stuff.rb
JonRowe: The `rspec:` bit means just the rspec generator, not the rails generator, so its the same as invoking `controller wombats ...` only it does't produce a controller.

> One more testing of controller specs? I thought the whole issue was about request specs

This is fixing an issue the rename raised, this lines original purpose was to complete the coverage of the generators.

At the time the lines above generated controller specs for wombats, so we added a request spec to complete the mix. When the default became request specs this should have changed to a controller spec to restore the coverage.

JonRowe: So the rename showed a duplicate spec being generated, which resolving the original coverage issue solves.

klyonrad: Okay, now I see what you were reading into those lines... I was interpreting them in a very different way.

I thought that
```ruby
generate('rspec:request wombats')
generate('integration_test widgets')
```

were just for testing the RequestSpecGenerator or to have test coverage of request specs in general.

---

> The `rspec:` bit means just the rspec generator, not the rails generator, so its the same as invoking `controller wombats ...` only it does't produce a controller.
> 

But couldn't this be also achieved (and be more explicit) by changing line 87 to `generate('controller wombats index --controller-specs')`?

JonRowe: Ah yep I guess that'd also work

PR #2379, opened 2020-09-01
Experimental removal of Rubygems workarounds (3-9-maintenance)

pirj:
Due to unstable version resolution and lack of Ruby version constraints on older gemspecs in the Rubygems database, we had to manually declare versions to be used.

This has seemingly been fixed.

Related:
https://github.com/rubygems/rubygems/issues/3463
https://github.com/rubygems/rubygems.org/pull/2474
https://github.com/rubygems/rubygems/pull/2662

Let's see if it works now, and if the issue with IPv6 fallback is still the case.

PR #2380, opened 2020-09-01
Experimental removal of Rubygems workarounds (main branch)

pirj:
Due to unstable version resolution and lack of Ruby version constraints on older gemspecs in the Rubygems database, we had to manually declare versions to be used.

This has seemingly been fixed.

Related:
https://github.com/rubygems/rubygems/issues/3463
https://github.com/rubygems/rubygems.org/pull/2474
https://github.com/rubygems/rubygems/pull/2662

Let's see if it works now, and if the issue with IPv6 fallback is still the case.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,61 +14,30 @@ group :documentation do
   gem 'relish', '~> 0.7.1'
 end
 
-platforms :jruby do
-  gem "jruby-openssl"
-end
-
-case RAILS_VERSION
-when /main/
-  MAJOR = 6
-  MINOR = 0
-when /5-2-stable/
-  MAJOR = 5
-  MINOR = 2
-when /stable/
-  MAJOR = 6
-  MINOR = 0
-when nil, false, ""
-  MAJOR = 6
-  MINOR = 0
-else
-  match = /(\d+)(\.|-)(\d+)/.match(RAILS_VERSION)
-  MAJOR, MINOR = match.captures.map(&:to_i).compact
-end
-
-if MAJOR >= 6
-  gem 'selenium-webdriver', '~> 3.5', require: false
-  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
-else
-  gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
-end
-
-gem 'ffi', '~> 1.9.25'
-
 gem 'rake', '~> 12'
-
-gem 'mime-types', "~> 3"
-
-if RUBY_VERSION.to_f < 2.3
-  gem 'capybara', '~> 3.1.0'
-elsif RUBY_VERSION.to_f < 2.4
-  gem 'capybara', '< 3.16'
-elsif RUBY_VERSION.to_f < 2.5
-  gem 'capybara', '< 3.33'
-else
-  gem 'capybara', '>= 2.13', '< 4.0', require: false
-end
-
-if MAJOR < 6
-  gem 'nokogiri', '1.9.1'
-else
-  gem 'nokogiri', '>= 1.10.8'
-end
-
-gem "rubyzip", '~> 1.2'
 
 if RUBY_VERSION.to_f >= 2.3
   gem 'rubocop', '~> 0.80.1'
+end
+
+gem 'capybara'
+
+MAJOR =
+  case RAILS_VERSION
+  when /5-2-stable/
+    5
+  when /main/, /stable/, nil, false, ''
+    6
+  else
+    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
+  end
+
+if MAJOR >= 6
+  # sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
+  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
+else
+  # Similarly, Rails 5.0 only supports '~> 1.3.6'. Rails 5.1-5.2 support '~> 1.3', '>= 1.3.6'
+  gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
 end
 
 custom_gemfile = File.expand_path('Gemfile-custom', __dir__)
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ group :documentation do
   gem 'relish', '~> 0.7.1'
 end
 
-gem 'rake', '~> 12'
+gem 'rake', '> 12'
 
 if RUBY_VERSION.to_f >= 2.3
   gem 'rubocop', '~> 0.80.1'
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -21,54 +21,40 @@ in_root do
   gsub_file 'Gemfile', /^.*\bgem 'rails.*$/, ''
   gsub_file "Gemfile", /.*web-console.*/, ''
   gsub_file "Gemfile", /.*debugger.*/, ''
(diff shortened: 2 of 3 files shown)


benoittgt:
Ouhao. Super happy if it is working.

pirj:
Ran the build several times, it's green apart from a [`mountable_engine?` failure](https://travis-ci.org/github/rspec/rspec-rails/jobs/723440858) that appears randomly in some builds jobs, but disappears on the second attempt.

[`3-9-maintenance`](https://github.com/rspec/rspec-rails/pull/2379) will probably be even more interesting.
I'll handle the other repos as well after those two.

JonRowe:
Have we seen the mountable_engine? failure on main yet? It must be related to #2372

Gemfile
```diff
@@ -14,63 +14,33 @@ group :documentation do
   gem 'relish', '~> 0.7.1'
 end
 
-platforms :jruby do
-  gem "jruby-openssl"
-end
+gem 'rake', '~> 12'
```
JonRowe, line 17: Why not `> 12`, 13 is out?

Gemfile
```diff
@@ -14,63 +14,33 @@ group :documentation do
...
+    5
+  when /main/, /stable/, nil, false, ''
+    6
+  else
+    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
+  end
+
 if MAJOR >= 6
-  gem 'selenium-webdriver', '~> 3.5', require: false
+  # Rails 6.0 specify '~> 1.4' directly in the adapter.
```
JonRowe, line 36: Do you mean `> 1.4?` Either way we should let Rails decide what version to use?

example_app_generator/generate_app.rb
```diff
@@ -21,54 +21,37 @@
...
   gsub_file "Gemfile", /.*web-console.*/, ''
   gsub_file "Gemfile", /.*debugger.*/, ''
-  gsub_file "Gemfile", /.*byebug.*/, "gem 'byebug', '~> 9.0.6'"
-  gsub_file "Gemfile", /.*puma.*/, ""
-  gsub_file "Gemfile", /.*gem..sqlite3.*/, "gem 'sqlite3', '~> 1.3.6'"
-  gsub_file "Gemfile", /.*bootsnap.*/, ""
-
-  if Rails::VERSION::STRING >= '5.0.0'
-    append_to_file('Gemfile', "gem 'rails-controller-testing'\n")
-  end
```
JonRowe, line 31: This should stay due to the "soft" 4.2 support

JonRowe: Nevermind, I see we're now adding it for all.

Gemfile
pirj: Isn't `~>` a different way to describe the same? When looking at the logs I had an impression it installs 13.0.something.

example_app_generator/generate_app.rb
pirj: You are actually right here. In our builds - sure, we don't need that condition. But for whoever decides to debug why RSpec doesn't work with their Rails 4.2 that will be a problem since `rails-controller-testing` [depends on Rails 5+](https://github.com/rails/rails-controller-testing/blob/v1.0.0/rails-controller-testing.gemspec#L21) (this is the first time I see `~> 5.x` syntax).

Going to revert this removal and add a comment.

Gemfile
pirj: The comment below makes no sense. When `sqlite3` 1.5.0 is out, it still matches the [Rails' requirement](https://github.com/rails/rails/blob/6-0-stable/activerecord/lib/active_record/connection_adapters/sqlite3_adapter.rb#L13).

But if it happens that `sqlite3` decides to jump straight to 2.0, our builds will be broken if we omit this constraint.

I'll fix the comment.

Gemfile
JonRowe: Oh of course, no points, `> 12` is clearer here, `~>` is usually used to install a point release

Gemfile
JonRowe: Sorry it was late when I wrote this, what I mean is if the Rails constraint is `~> 1.4` then this has no effect and can be removed. If the rails constraint is `> 1.4` then this has an effect but again I don't feel we should constrain this, we need to match Rails here.

JonRowe (approved):

Gemfile
pirj: Point is `sqlite3` is an optional Rails dependency, they only call `gem` when the adapter is set to `sqlite` in `config/database.yml`. But still they have a version constraint defined there, which Bundler has no awareness of, since it's outside of the `Gemfile`.

If we remove it, `sqlite3` won't be installed and the build would break.
If we remove the version constraint, the build will break for Rails 6.0 when (eventually, far in the future) `sqlite3` 2.0 is out.

JonRowe: Right ok, thats not clear from the comment at all.

Gemfile
```diff
@@ -14,63 +14,33 @@ group :documentation do
...
+  # This is to prevent breakages when `sqlite3` 2.0.0 is out.
```
JonRowe, line 37:
```suggestion
  # sqlite3 is a optional, unspecified, dependancy and Rails 6.0 only supports `~> 1.4`
```

Gemfile
```diff
@@ -14,63 +14,33 @@ group :documentation do
...
   gem 'sqlite3', '~> 1.4', platforms: [:ruby]
 else
+  # Rails 5.0 specify '~> 1.3.6', 5.1 and 5.2 - "~> 1.3", ">= 1.3.6" directly in the adapter.
```
JonRowe, line 40:
```suggestion
  # Similarly Rails 5.0 only supports '~> 1.3.6',  with 5.1 and 5.2 supporting "~> 1.3", ">= 1.3.6"
```

pirj:
All green, except for `mountable_engine?` failures. Restarted those jobs - now all green.

example_app_generator/generate_app.rb
JonRowe: Are you going to revert this?

pirj: Oops, moved it [down below](https://github.com/rspec/rspec-rails/pull/2380/files#diff-a8a7f2d3f06e0c6f9a7717dd8c7cb715R37). Let me quickly move it back.

pirj: Done.

JonRowe (approved):

Gemfile
```diff
@@ -14,63 +14,32 @@ group :documentation do
...
+  when /main/, /stable/, nil, false, ''
+    6
+  else
+    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
+  end
+
 if MAJOR >= 6
-  gem 'selenium-webdriver', '~> 3.5', require: false
+  # sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
   gem 'sqlite3', '~> 1.4', platforms: [:ruby]
```
pirj, line 37: `example_app_generator/generate_app.rb` ends up with
```ruby
gem 'sqlite3'
```
in example_app's Gemfile. Why there's no version mismatch, and how the proper version is resolved is beyond my understanding.
I used a different approach for 3-9-maintenance branch, moved this selection to `Gemfile-rails-dependencies`, and added a statement to strip example_app's `Gemfile` from `gem 'sqlite3'`. https://github.com/rspec/rspec-rails/pull/2386/files#diff-8c3955be04b733dbd31e54e50844fb84

PR #2381, opened 2020-09-01
Remove predicates that apply to unsupported Rubies

pirj:


benoittgt (approved):
Thanks @pirj for the cleaning.

Ok for me because errors are Yard related.

JonRowe:
I don't think we should merge this because its much easier to keep as much of the build in sync with rspec-dev as possible, yes there are some superfluous predicates but its easier to update this file with the others if one we use does change?

pirj:
`is_ruby_23_plus` will become redundant when we drop support for Ruby 2.2.
`additional_specs_available` is unused.
We use `style_and_lint_enforced` only because we support more Ruby versions than our RuboCop version does. Also can be removed along with Ruby 2.2.

The only two useful ones are `is_mri`/`is_jruby`. But those two are so simple and highly unlikely to change over time.
Maybe we could get rid of them as well if as you recently mentioned we support MRI-compatible Rubies, and JRuby et al should strive for that compatibility.

JonRowe:
Yes and when we move to RSpec 4 we can drop the functions for all repos, my point is that they are currently (to the best of my knowledge) still synced by the update pr, thus to reduce the risk of churn we can leave the unused functions.

pirj:
The update PR applies in the part of the common text files, but not travis, rubocop and scripts I believe.

> Yes and when we move to RSpec 4

If it happens soon, there won't be much in that release besides dropping older Rubies.

Anyway, I don't mind closing this, as I work on a couple more cleanups already, it just caught my eye, and I couldn't resist sending this.

Do you guys happen to know what is causing this `mountable_engine?` flaky failure?

pirj:
I was wrong with the removal of `documentation_enforced` predicate, we did that on purpose as Yard fails with that weird error on JRuby.

JonRowe:
> If it happens soon, there won't be much in that release besides dropping older Rubies 😆

You joke but really thats one of the main purposes of a major release for us, we've talked about that being a thing.

> Anyway, I don't mind closing this, as I work on a couple more cleanups already, it just caught my eye, and I couldn't resist sending this.

Closing for now.

> Do you guys happen to know what is causing this mountable_engine? flaky failure?

No but I don't think its a coincidence that we've recently merged an generator that uses this...

Issue #2382, opened 2020-09-13
Drop `git` in gemspec


utkarsh2102:
Hi @JonRowe & @pirj,

Thanks for your great work on this! ❤️ 

Whilst maintaining this in Debian, we found that this library relies on `git` to produce a list of files, which can be achieved by a plain Ruby alternative. Using `git` in gemspec is problematic in general. We're working with the bundler team to fix this there itself.

It'd be great if we can drop `git` from gemspec here, at least! If you concur, I'd be happy to raise a PR doing the same!
Please let me know what you think! 😄

JonRowe:
We're happy to consider an alternative providing it produces the same output, in particular the point of using `git` is to avoid a gem spec accidentally including uncommitted files during a release, given that everyone releasing rspec up until now has had git installed this has never been a problem.

pirj:
Hi!

I understand your concern about building deb packages. Is it problematic to add a git dependency for building them? Or are you building outside of a git repository?

This was also done this way to avoid publishing a gem with extra files that don't really belong to the git repository, e.g. bundler binstubs etc that are in the `.gitignore` file.
I glanced over our repos's `gemspec` files, and don't really see any technical obstacles for using `Dir[]` instead of `git` shell-out. `Dir[]` seems to be supported on Ruby 1.8.6+, this shouldn't be a problem.
Did you plan to change this in this repository only, or all of them? `rspec-collection_matchers` differs slightly.

> avoid a gem spec accidentally including uncommitted files during a release

As the maintainer, @JonRowe has a much better view of the problem. I've just heard it's not that straightforward as it seems.

pirj:
RuboCop's opinion on this topic https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#packaginggemspecgit

utkarsh2102:
Hello 👋🏻 

> We're happy to consider an alternative providing it produces the same output, in particular the point of using git is to avoid a gem spec accidentally including uncommitted files during a release, given that everyone releasing rspec up until now has had git installed this has never been a problem.

> Is it problematic to add a git dependency for building them? Or are you building outside of a git repository?
> This was also done this way to avoid publishing a gem with extra files that don't really belong to the git repository, e.g. bundler binstubs etc that are in the .gitignore file.


Ah, so bundler now doesn't let you release if you've got uncommited changes in your repository. So that's good to go!
[...]

PR #2384, opened 2020-09-20
Drop `git` in gemspec

utkarsh2102:
Hello @JonRowe & @pirj  👋🏻,

Thanks for your work on this again! ❤️ 

As discussed in #2382, we found that `rspec-rails` relies on `git ls-files` to produce a list of files, which could also be achieved using a pure Ruby alternative. Using `git` in gemspec, in general, is a bit problematic.
The rationale behind that can be found here: https://docs.rubocop.org/rubocop-packaging/cops_packaging.html#gemspec-git-rationale

This PR drops the usage of `git` and uses `Rake::FileList` instead w/o changing the output:

```
irb(main):001:0> require "rake/file_list"
=> true
irb(main):002:0> `git ls-files -- lib/*`.split("\n").sort == Rake::FileList["lib/**/{.*,*}"].exclude(*File.read(".gitignore").split).reject { |f| File.directory?(f) }
=> true

```

Also adds the `Packaging` extension to help the downstream maintainers as well! 😄 

Let me know what you think? 💫 

Closes: #2382

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -1,6 +1,6 @@
 # Over time we'd like to get this down, but this is what we're at now.
 Metrics/CyclomaticComplexity:
-  Max: 9 # default: 6
+  Max: 10 # default: 6
 
 # Over time we'd like to get this down, but this is what we're at now.
 Layout/LineLength:
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -1,5 +1,6 @@
 $LOAD_PATH.unshift File.expand_path('lib', __dir__)
 require "rspec/rails/version"
+require "rake/file_list"
 
 Gem::Specification.new do |s|
   s.name        = "rspec-rails"
@@ -21,7 +22,7 @@ Gem::Specification.new do |s|
     'source_code_uri'   => 'https://github.com/rspec/rspec-rails',
   }
 
-  s.files            = `git ls-files -- lib/*`.split("\n")
+  s.files            = Rake::FileList["lib/**/{.*,*}"].exclude(*File.read(".gitignore").split).reject { |f| File.directory?(f) }
   s.files           += %w[README.md LICENSE.md Changelog.md Capybara.md .yardopts .document]
   s.test_files       = []
   s.rdoc_options     = ["--charset=UTF-8"]
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,9 @@
 inherit_from:
   - .rubocop_todo.yml
 
+require:
+  - rubocop-packaging
+
 AllCops:
   Exclude:
     # Templates are really ERB which Rubocop does not parse
@@ -119,6 +122,9 @@ Lint/EmptyExpression:
   Enabled: false
 
 Lint/ImplicitStringConcatenation:
+  Enabled: false
+
+Lint/MissingSuper:
   Enabled: false
 
 Lint/NestedMethodDefinition:
@@ -280,9 +286,6 @@ Style/IfInsideElse:
   Enabled: false
 
 Style/IfUnlessModifier:
-  Enabled: false
-
-Style/MethodMissingSuper:
   Enabled: false
 
 Style/MissingRespondToMissing:
--- a/Gemfile
+++ b/Gemfile
@@ -16,7 +16,7 @@ end
 
 gem 'rake', '> 12'
 
-if RUBY_VERSION.to_f >= 2.4
+if RUBY_VERSION.to_f >= 2.4 && RUBY_ENGINE == 'ruby'
   gem 'rubocop', '~> 0.91.0'
   gem 'rubocop-packaging', '~> 0.5'
 end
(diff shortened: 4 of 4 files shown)


rspec-rails.gemspec
```diff
@@ -21,7 +22,7 @@ Gem::Specification.new do |s|
     'source_code_uri'   => 'https://github.com/rspec/rspec-rails',
   }
 
-  s.files            = `git ls-files -- lib/*`.split("\n")
+  s.files            = Rake::FileList["lib/**/{.*,*}"].exclude(*File.read(".gitignore").split).reject { |f| File.directory?(f) }
```
pirj, line 25: Does `exclude` make any difference?

pirj, line 25: Is there a difference between `Rake::FileList.ne` and `Dir[]` for our use case?

pirj, line 25: Is rejection of directories necessary? Will the resulting gem contain those dirs as as empty dirs, or their files as well? Or will those dirs be skipped?

pirj (commented):
To reduce the difference in how this is done in different repos, do you mind to send similar pulls to the other ones to see if this solution fits all of them?

Thanks for updating RuboCop along the way. I didn't think it will work out that easily 👍 
RuboCop is outdated in Core and Mocks as far as I remember, I suggest you skip that part in those repos.

rspec-rails.gemspec
utkarsh2102: Not really, just helps to respect the `.gitignore` file.

utkarsh2102: I don't think so. I prefer `Dir` over `Rake::FileList`. But to respect the `.gitignore`, I went ahead with `Rake::FileList`, do you prefer `Dir`? I'll be happy to implement whatever you're happy with! 😄

utkarsh2102: When you use `git ls-files`, it skips the empty directories. `.reject` does the same thing since `Rake::FileList` and `Dir` are not capable of this. I personally just feel `Dir` is sufficient alone (without `.exclude` and `.reject`). What do you think?

pirj: I got it. Surely using `Dir` would be nice. But if it happens that we need to `exclude`, `FileList` is a better option.
Let's see if in the other repos `exclude` makes any difference, let's go with `FileList`.

utkarsh2102: You want me to open PRs for all the repositories now itself?
I'd say, let's go one by one instead? 🙈

pirj: The only reason I'm asking is to be able to make this code similar across apps.
If it happens that some other repo need to `exclude`, it would be nice to use `FileList` everywhere, not only in that other repo.

pirj:
@JonRowe @benoittgt Have you seen this JRuby failure before?
```
Failures:
  1) Action Mailer railtie hook in the development environment respects a custom `preview_path`
     Failure/Error: CaptureExec.new(out, $?.exitstatus)
     NoMethodError:
       undefined method `exitstatus' for nil:NilClass
     # ./spec/verify_mailer_preview_path_spec.rb:27:in `capture_exec'
     # ./spec/verify_mailer_preview_path_spec.rb:61:in `block in <main>'
```

rspec-rails.gemspec
utkarsh2102: Okay, so I'll open PRs this weekend for:
- rspec-support 
- rspec-dev
- rspec-expectations
- rspec-mocks
- rspec-its

Please let me know if I'm missing something 😅

pirj: `rspec-dev` isn't backing any gem, but https://github.com/rspec/rspec does.

Optional repos you may want to include:
 - https://github.com/rspec/rspec-collection_matchers
 - https://github.com/rspec/rspec-its
 - https://github.com/rspec/rspec-activemodel-mocks
 - https://github.com/rspec/rspec-legacy_formatters
 - https://github.com/rspec/rspec-autotest

.rubocop_todo.yml
```diff
@@ -1,6 +1,6 @@
 # Over time we'd like to get this down, but this is what we're at now.
 Metrics/CyclomaticComplexity:
-  Max: 9 # default: 6
+  Max: 10 # default: 6
```
JonRowe, line 3: You need to not increase our max cyclomatic complexity, we're trying to reduce it.

rspec-rails.gemspec
JonRowe: This still doesn't avoid uncommitted files...

JonRowe (changes requested):

JonRowe:
I think really if we want to avoid a command line git dependency we might need a ruby based dependency to ensure the same result. Although linux packagers might just clone the repo and release directly, I can tell you, as the maintainer I rely on the `git ls-files` behaviour to not accidentally introduce files...

JonRowe:
Closing as stale. Any Ruby replacement for the git commands needs to provide the same safeguards as the git commands, or allow for the git commands to be used when git is available.

PR #2385, opened 2020-09-20
ActionDispatch::TestProcess needs to access :file_fixture_path on Rails 6.1

sudara:
I had a similar problem to @koenpunt in #2349 where `fixture_file_upload` [stopped working](https://travis-ci.com/github/sudara/alonetone/builds/185469701).

```
     NoMethodError:
       undefined method `file_fixture_path' for RSpec::Rails::FixtureFileUploadSupport::RailsFixtureFileWrapper:Class
       Did you mean?  fixture_path
     # /actionpack/lib/action_dispatch/testing/test_process.rb:25:in `fixture_file_upload'
     # /rspec-rails-dd093928f021/lib/rspec/rails/fixture_file_upload_support.rb:5:in `fixture_file_upload'

```

In my case, it's because I'm on Rails 6.1 alpha. As of [these recent changes]( https://github.com/rails/rails/pull/39086), `ActionController::TestCase` now needs access to `file_fixture_path` within `fixture_file_upload`.

There's not currently any CI for Rails 6.1 (it's not out yet) but I thought I would get a head start. I added the reader on a separate line, as the `fixture_path` reader is turning into an accessor on #2370.

### DIFF
--- a/lib/rspec/rails/fixture_file_upload_support.rb
+++ b/lib/rspec/rails/fixture_file_upload_support.rb
@@ -23,6 +23,7 @@ module RSpec
 
         class << self
           attr_accessor :fixture_path
+          attr_reader :file_fixture_path
 
           # Get instance of wrapper
           def instance
--- a/spec/rspec/rails/fixture_file_upload_support_spec.rb
+++ b/spec/rspec/rails/fixture_file_upload_support_spec.rb
@@ -27,7 +27,7 @@ module RSpec::Rails
 
     context 'ActionDispatch::TestProcess on Rails 6.1' do
       it 'can read the file_fixture_path attribute' do
-          expect(FixtureFileUploadSupport::RailsFixtureFileWrapper).to respond_to(:file_fixture_path)
+        expect(FixtureFileUploadSupport::RailsFixtureFileWrapper).to respond_to(:file_fixture_path)
       end
     end
 
--- a/spec/rspec/rails/fixture_file_upload_support_spec.rb
+++ b/spec/rspec/rails/fixture_file_upload_support_spec.rb
@@ -25,6 +25,12 @@ module RSpec::Rails
       end
     end
 
+    context 'ActionDispatch::TestProcess on Rails 6.1' do
+      it 'can read the file_fixture_path attribute' do
+          expect(FixtureFileUploadSupport::RailsFixtureFileWrapper).to respond_to(:file_fixture_path)
+      end
+    end
+
     def fixture_file_upload_resolved(fixture_name, fixture_path = nil)
       RSpec::Core::ExampleGroup.describe do
         include RSpec::Rails::FixtureFileUploadSupport


pirj:
Wondering if it makes sense to extend the matrix to run against Rails `master`. It would be nice to have a failing test along with the fix.

benoittgt:
> Wondering if it makes sense to extend the matrix to run against Rails `master`.

I think it will add lot's of work. What about follow alpha,beta and first releases when they are published?

> It would be nice to have a failing test along with the fix.
+1

sudara:
I wasn't sure how rspec-rails has historically coordinated Rails releases.

> extend the matrix to run against Rails master

Having rspec-rails generally track rails master would be a recipe for constant breakage on main, but this would work well for a 6.1 branch. When 6.1 is released, the test target could be switched from master to 6.1. But again, I'm not sure how rspec-rails usually does things.

> It would be nice to have a failing test along with the fix.

There will be failing tests aplenty when testing against Rails 6.1. I added a failing test specific to the `file_fixture_path` accessor, but I think its usefulness is debatable.

I'm happy with this remaining open or merged into 6.1 branch is supported. My primary intent was to communicate the problem/fix to others!

JonRowe:
Can you rebase this?

sudara:
✅ Rebased

Though I'm seeing some sort of Thor failure in the suite, looks like the issue has been on main recently:

```
      Thor::UndefinedCommandError:
11180        Could not find command "mountable_engine?".
```

pirj:
It's flaky, I've restarted the job, should most probably pass.

lib/rspec/rails/fixture_file_upload_support.rb
```diff
@@ -23,6 +23,7 @@ class RailsFixtureFileWrapper
 
         class << self
           attr_accessor :fixture_path
+          attr_reader :file_fixture_path
```
JonRowe, line 26: If this is only for the next version of rails, we need some kind of feature flag.

spec/rspec/rails/fixture_file_upload_support_spec.rb
```diff
@@ -25,6 +25,12 @@ module RSpec::Rails
       end
     end
 
+    context 'ActionDispatch::TestProcess on Rails 6.1' do
+      it 'can read the file_fixture_path attribute' do
+        expect(FixtureFileUploadSupport::RailsFixtureFileWrapper).to respond_to(:file_fixture_path)
```
JonRowe, line 30: This explains the how, but not the why, any chance of an integration spec for 6.1? Or do some of our existing integration specs fail on 6.1 for this reason?

JonRowe (changes requested):

spec/rspec/rails/fixture_file_upload_support_spec.rb
sudara: > Or do some of our existing integration specs fail on 6.1 for this reason?

Exactly, the wrapped class `ActionDispatch::TestProcess` requires access to this attribute as of Rails 6.1, otherwise fixture_file_upload bombs out. As to why that class needs to be wrapped in the first place, I have no idea. 

I'm not sure what the convention is for upcoming releases: is 6.1 running on a branch somewhere? Is it trivial for me to set that up locally?

lib/rspec/rails/fixture_file_upload_support.rb
sudara: Prior versions of Rails do not access `file_fixture_path` from within `ActionDispatch::TestProcess#fixture_file_upload`. Is there a benefit to not exposing the reader in those cases? Otherwise it feels a bit heavy handed.

JonRowe: These is because the presence of this reader might be seen by other libraries feature detection and the wrong version of Rails assumed, we don't use it ourselves so theres no point having a defined reader that does nothing but might potentially false flag other libraries.

sudara: Maybe I'm missing something, but it doesn't feel right to me to add code to support a hypothetical edge case such as a third party library leaning on this particular class in RSpec to determine a Rails version. However, this is your project, so please merge / modify / close out this PR as you wish.

JonRowe:
Thanks for researching this and bringing it to our attention, there was more work needed to ensure it worked without deprecation warning but this is closed in #2398

PR #2386, opened 2020-09-24
Remove Ruby version constraint workarounds

pirj:
The point of this PR is to remove workaround previously needed to work around Rubygems API not providing necessary gem Ruby constraints under certain conditions.

The problem was presumably fixed in Rubygems.
A couple of passed builds have no odd gem version resolutions resulting in Ruby version incompatibility.

Follow-up to https://github.com/rspec/rspec-rails/pull/2380

Related to https://github.com/rubygems/rubygems/issues/3463

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,42 +18,6 @@ group :documentation do
   end
 end
 
-platforms :jruby do
-  gem "jruby-openssl"
-end
-
-gem 'sqlite3', '~> 1.3.6'
-
-if RUBY_VERSION >= '2.4.0'
-  gem 'json', '>= 2.0.2'
-end
-
-if RUBY_VERSION < '1.9'
-  gem 'ffi', '< 1.9.19' # ffi dropped Ruby 1.8 support in 1.9.19
-elsif RUBY_VERSION < '2.0'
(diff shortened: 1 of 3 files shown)


pirj:
Green! 
image

Gemfile
```diff
@@ -63,28 +27,14 @@ if RUBY_VERSION >= '1.9.3'
...
-  # Only < 1.0 supports 1.8.7
-  gem 'rubyzip', '< 1.0'
-elsif RUBY_VERSION < '2.4'
-  gem 'rubyzip', '>= 1.2.2', '< 2.0.0'
-end
+gem 'mime-types'
+gem 'nokogiri'
+gem 'rake'
+gem 'rack-cache'
+gem 'rubyzip'
```
JonRowe, line 37: Are these needed directly? We mostly specified them to pin them, I don't think anything but `rake` is needed?

pirj: Good point, thanks.

pirj:
It was a nice excursion back into time when we had 72 build jobs.
https://travis-ci.org/github/rspec/rspec-rails/builds/734613066 is green
Squashed commits and waiting for yet another green build.

JonRowe (approved):
If it goes green feel free to merge

benoittgt:
Yes!! Thanks

PR #2387, opened 2020-09-25
Integration generator deprecation

### LINKED ISSUE #2374: Deprecate IntegrationGenerator
**Is your ticket related to a problem? Please describe.**

There is a IntegrationGenerator class and the code comments mention something about "deprecation for this class before rspec-rails 4". I read the commit message and I don't quite understand if it is technically necessary or not? That's why I am opening this ticket for discussion/explanation 🤔 

Maybe if we delete the class (and the shared examples) we could reduce the development complexity about generators a bit.

From a user perspective I would say that having that integration generator options is just a bit confusing.

**Additional context**

https://github.com/rspec/rspec-rails/pull/1378

klyonrad:
Closes #2374

With these changes executing `rails generate rspec:integration` produces the following message

```
DEPRECATION: 
  The integration generator is deprecated and will be deleted in RSpec-Rails 5. Please use the request generator instead.
 Called from ...../rspec-rails/lib/generators/rspec/integration/integration_generator.rb:18:in `generate_request_spec'
```

# Test Plan

```bash
rake generate:app
cd tmp/example_app/
rails generate rspec:integration Bottles
# DEPRECATION: 
#  IntegrationGenerator is deprecated and will be deleted in RSpec 5. Please use `rspec:generator` instead. It should behave in the same way.
# Called from .../rspec-rails/lib/generators/rspec/integration/integration_generator.rb:18:in `generate_request_spec'.
#      create  spec/requests/bottles_spec.rb
rails generate rspec:request Bottles
#    identical  spec/requests/bottles_spec.rb
```


klyonrad:
TODO: test out the scaffolds generation

spec/generators/rspec/integration/integration_generator_spec.rb
```diff
@@ -4,5 +4,83 @@
 
 RSpec.describe Rspec::Generators::IntegrationGenerator, type: :generator do
   setup_default_destination
-  it_behaves_like "a request spec generator"
```
JonRowe, line 7: Rather than removing this, it would be better to modify the shared example by passing in a flag to specifying when to define that extra spec.

lib/generators/rspec/integration/integration_generator.rb
```diff
@@ -1,19 +1,27 @@
...
+      source_paths << File.expand_path('../request/templates', __dir__)
+
       def generate_request_spec
         return unless options[:request_specs]
 
+        RSpec.warn_deprecation <<-WARNING.gsub(/\s*\|/, ' ')
+          |IntegrationGenerator is deprecated
+          |and will be deleted in RSpec 5.
+          |Please use `rspec:generator` instead.
+          |It should behave in the same way.
```
JonRowe, line 22:
```suggestion
          |The integration generator is deprecated
          |and will be deleted in RSpec-Rails 5.
          |Please use the request generator instead.
```

JonRowe (changes requested):

klyonrad (commented):
I checked the scaffold integration: Fine ✔️ 

I tried the test suite out what happens `lib/generators/rspec/integration/integration_generator.rb` is deleted. This helped to catch the now unnecessary `require 'generators/rspec/integration/integration_generator'` statement 😁

spec/generators/rspec/integration/integration_generator_spec.rb
JonRowe: > So when the IntegrationGenerator class is deleted, should the shared examples stay there even though it will be only used one time?

No when they're deleted they can be pulled into the resulting specs, but thats not going to happen until the next major version.

> My interpretation was that the shared examples were only there because there was the inheritance of RequestGenerator < IntegrationGenerator. Adding a flag to an inheritance (which the shared examples are) is a code smell in my opinion.

No they are there because the two code paths, er, share examples, that there was an inheritance to share behaviour is an implementation detail, the examples would be valid if they both had vendored copies of the lass or completely different implementations.

pirj:
Thank you!
My apologies that it took us longer to process this. So long that I had to change "will be deleted in RSpec-Rails 5" to "RSpec-Rails 7" 😄

klyonrad:
@pirj No Problemo. I don't even remember doing this 😂

PR #2391, opened 2020-09-27
Be able to run ruby 3 preview 1 on the CI with allow_failure on

benoittgt:


benoittgt:
It should be good soon with https://github.com/rails/rails/pull/40281

JonRowe:
We could add the gem in the mean time to make it green?

benoittgt:
Errors are now different but more related to changes on tests helpers and method arity.

PR #2396, opened 2020-10-10
Drop 'main' from conditional as it's unrelated to Rails versions

pirj:
We probably meant 'master', but we don't run against Rails edge.


pirj:
JRuby jobs failed, it's unrelated and we had them failing previously, fixing in https://github.com/rspec/rspec-rails/pull/2397

Issue #2375, opened 2020-08-21
Improve generated request specs when no controller actions are added


klyonrad:
**Is your feature request related to a problem? Please describe.**

When controller generator is used it generates a requests_spec file that is quite empty. It consists only of the `describe .. do .. end` block without any helping indication that the user can use stuff like `expect(response).to have_http_status(200)` like it happens when the simpler RequestSpecGenerator is used.

**Describe the solution you'd like**

when generating a controller the request spec generated will have a different file name and when it is generated without any action it will add just one spec for an index action with a message about adding a real spec.

**Describe alternatives you've considered**

TBD

**Additional context**

This issue is extracted from a wish in https://github.com/rspec/rspec-rails/pull/2356

I had the idea that we can move the original PR along by separating the issues.

JonRowe:
This should be fairly easy for someone to add, but I don't have the time to work on the generators myself.

klyonrad:
This was fixed with https://github.com/rspec/rspec-rails/pull/2399

PR #2399, opened 2020-10-20
Improve controller template when no action

ThHareau:
Fix for https://github.com/rspec/rspec-rails/issues/2375

A generated file looks like this: 

```rb
require 'rails_helper'

RSpec.describe "Posts", type: :request do
  describe "GET /index" do
    it "returns http success" do
      pending "add some scenarios (or delete) #{__FILE__}"
    end
  end
end
```

### DIFF
--- a/lib/generators/rspec/controller/templates/request_spec.rb
+++ b/lib/generators/rspec/controller/templates/request_spec.rb
@@ -1,7 +1,14 @@
 require 'rails_helper'
 
 RSpec.describe "<%= class_name.pluralize %>", <%= type_metatag(:request) %> do
-<% namespaced_path = regular_class_path.join('/') %>
+<% namespaced_path = regular_class_path.join('/') -%>
+<% if actions.empty? -%>
+  describe "GET /index" do
+    it "returns http success" do
+      pending "add some scenarios (or delete) #{__FILE__}"
+    end
+  end
+<% end -%>
 <% for action in actions -%>
   describe "GET /<%= action %>" do
     it "returns http success" do
--- a/lib/generators/rspec/controller/controller_generator.rb
+++ b/lib/generators/rspec/controller/controller_generator.rb
@@ -14,7 +14,6 @@ module Rspec
 
       def generate_request_spec
         return unless options[:request_specs]
-
         template 'request_spec.rb',
                  File.join('spec/requests', class_path, "#{file_name}_spec.rb")
       end
--- a/lib/generators/rspec/controller/controller_generator.rb
+++ b/lib/generators/rspec/controller/controller_generator.rb
@@ -14,6 +14,7 @@ module Rspec
 
       def generate_request_spec
         return unless options[:request_specs]
+
         template 'request_spec.rb',
                  File.join('spec/requests', class_path, "#{file_name}_spec.rb")
       end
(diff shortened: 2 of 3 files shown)


ThHareau:
Made with @tsleal29, @yann120, @fwuensche and @musait

lib/generators/rspec/controller/templates/request_spec.rb
```diff
@@ -1,7 +1,14 @@
 require 'rails_helper'
 
 RSpec.describe "<%= class_name.pluralize %>", <%= type_metatag(:request) %> do
-<% namespaced_path = regular_class_path.join('/') %>
+<% namespaced_path = regular_class_path.join('/') -%>
+<% if actions.empty? -%>
+  describe "GET /index" do
+    it "returns http success" do
+      pending "add some scenarios (or delete) #{__FILE__}"
```
pirj, line 8: Not sure if "scenarios" is the best term to be used here. Maybe "examples"?

lib/generators/rspec/controller/templates/request_spec.rb
```diff
@@ -1,7 +1,14 @@
 require 'rails_helper'
 
 RSpec.describe "<%= class_name.pluralize %>", <%= type_metatag(:request) %> do
-<% namespaced_path = regular_class_path.join('/') %>
+<% namespaced_path = regular_class_path.join('/') -%>
+<% if actions.empty? -%>
+  describe "GET /index" do
+    it "returns http success" do
```
pirj, line 7: This is a failing example rather than a pending one.
There's a difference in how `pending` works inside an example and when used to declare an example.
When used inside an example, it expects the example to fail, and marks it as pending. If the example doesn't fail, it ... fails.

So I suggest changing this to:
```ruby
  describe "GET /index" do
    pending "add some scenarios (or delete) #{__FILE__}"
  end
```

pirj, line 7: Not specifically related to this pull request.

I'm slightly opposed to adding "returns http success" in request specs. From my point of view, they should test more than just http statuses, and providing this as a reference example in the template is promoting a dubious practice.

pirj (changes requested):
Amazing, thanks!
Just a few edits away from getting it done.

lib/generators/rspec/controller/templates/request_spec.rb
ThHareau: Indeed. Maybe I can update the template for that as well? Do you have an idea regarding what we can provide? 

I'm thinking to `exists` or `is reachable` (I have a preference for the second option)

pirj: Maybe `"returns a list of #{class_name.pluralize.downcase}"`? At least this is what an "index" originally means.

pirj: > update the template for that as well

That would be very helpful!

benoittgt (approved):
CI errors are on JRuby and unreleated. Your patch looks good to me. Thanks @pirj for the review :)

Thanks @ThHareau, @tsleal29, @yann120, @fwuensche and @musait

pirj (approved):
Thanks!

PR #2400, opened 2020-11-02
[wip] Changes to support Rails 6.1

JonRowe:
@benoittgt these are your changes, feel free to edit this PRs title / description.


.travis.yml
```diff
@@ -36,6 +36,9 @@ before_script:
 
 script: "script/run_build 2>&1"
 
+env:
+  - USE_GIT_REPOS=1
+
```
JonRowe, line 41: We should be automatically using git repos because theres no local ones

example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -14,15 +14,9 @@
...
   CODE
-  gsub_file 'config/initializers/action_mailer.rb',
-            /ExampleApp/,
-            Rails.application.class.parent.to_s
 
   copy_file 'spec/support/default_preview_path'
   chmod 'spec/support/default_preview_path', 0755
-  gsub_file 'spec/support/default_preview_path',
-            /ExampleApp/,
-            Rails.application.class.parent.to_s
```
JonRowe, line 25: Why are we doing this? It should be replaced not removed no?

lib/rspec/rails/matchers/action_mailbox.rb
```diff
@@ -24,7 +24,11 @@ def initialize(message)
 
           def matches?(mailbox)
             @mailbox  = mailbox
-            @receiver = ApplicationMailbox.router.send(:match_to_mailbox, inbound_email)
+            if ApplicationMailbox.router.respond_to?(:mailbox_for)
```
JonRowe, line 27: We should be conditionally defining the  method based on this.

example_app_generator/generate_action_mailer_specs.rb
benoittgt: We had errors like `undefined method 'parent' for ExampleApp::Application:Class `. I noticed that this gsub is not needed anymore.

lib/rspec/rails/matchers/action_mailbox.rb
benoittgt: Good idea, I will make the change

.travis.yml
benoittgt: That's strange because on failures, bundler was pointing to a local gem sources. With this ENV it now pass.

benoittgt:
Sorry @JonRowe I should have open a PR instead of pushing to your branch.

lib/rspec/rails/matchers/action_mailbox.rb
benoittgt: I switched back because I add and error on `ActionMailbox` and I find it more complicated to duplicated all the block. Maybe you prefer in a method? No problem if you want to push another solution on this branch. :)

On `script/run_build`
```
An error occurred while loading spec_helper.
Failure/Error: if ApplicationMailbox.router.respond_to?(:mailbox_for)

NameError:
  uninitialized constant RSpec::Rails::Matchers::ActionMailbox::ReceiveInboundEmail::ApplicationMailbox
# ./spec/spec_helper.rb:27:in `<top (required)>'
No examples found.
```

benoittgt:
Our jruby errors may be linked to https://github.com/mikel/mail/issues/912
I will also look at issues in "allow failures", it seems easy to fix.

lib/rspec/rails/matchers/action_mailbox.rb
benoittgt: But we do not want to evaluate every-time `ApplicationMailbox.router.respond_to?(:mailbox_for)` when method is called. I will switch back and fix the error.

benoittgt: https://github.com/rspec/rspec-rails/pull/2400/commits/5c4186f58202c177ebc5d783e35e9f15efe2ad7e

benoittgt:
ATM I am not able to reproduce failing build on jruby localy. I switched to last jruby release instead of head.

benoittgt:
@JonRowe what do you think about merging it to rails-6-1-dev? I will reopen a PR with Rails master branch to see if we have something that could break rspec-rails.

JonRowe:
Er this needs picking apart and cleaning up, its way to confused right now.

benoittgt:
Ok sorry.

The thing is rails-6-1-dev is getting too much behind main? Should we rebase main into rails-6-1-dev first? It will remove lot's of commits from here.

benoittgt:
Ok. ATM I see no need to keep this PR open. 

Contributions to `rails-6-1-dev` should be smaller. `rails-6-1-dev` has no issue with Ruby3 and the Rails RC1, so I think we are pretty good.

Issue #2394, opened 2020-10-05
`undefined method `except' for class `Hash'` when `expect({}).to receive(:except)` on ruby-2.2


ojab:
### Subject of the issue

It requires running rails application and I'm not that interested to minimize testcase more since I don't use ruby-2.2 and encountered this by making PR to another project.
 
```ruby
require 'rails'
require 'rspec'

module Dummy
  class Application < Rails::Application
  end
end

Dummy::Application.initialize!



describe 'fail' do
  subject { {} }

  it do
    expect(subject).to receive(:except)
    subject.except
  end
end
```
### Your environment
* Ruby version: 2.2.10
* rspec-mocks version: 3.9.1
* rails version: 5.2.2

### Steps to reproduce
`bundle exec rspec /tmp/tst.rb`

### Expected behavior
No errors

### Actual behavior
```
Failures:

  1) fail is expected to receive except(*(any args)) 1 time
     Failure/Error: object_singleton_class.__send__(@original_visibility, method_name)

     NameError:
       undefined method `except' for class `Hash'
     # /home/ojab/.rvm/gems/ruby-2.2.10/gems/rspec-mocks-3.9.1/lib/rspec/mocks/method_double.rb:108:in `public'
```

JonRowe:
Rails is messing with the hash here, causing a confusing error, this might need to be transferred back to rspec-mocks if its fixable.

pirj:
Off the top of my head you need to require `rails/all`, or specifically `active_support/core_ext/hash` that defines `Hash#extract` before running the test.

Can you confirm that `{}.except(:a)` does not error out if you add it as the first line of your test example, @ojab ?

ojab:
It fails only if expectation is set, so `subject.except(:a)` on the first line works.
I've tried to require rails in different combinations (including `rails/all`) trying to minimize it and the error is not reproducible, so rails app initialization is the key.

pirj:
Less trouble - I can reproduce on Ruby 2.6.3 as well.

pirj:
I could find a way to work this around:
```diff
      # @private
      def restore_original_visibility
        return unless @original_visibility &&
          MethodReference.method_defined_at_any_visibility?(object_singleton_class, @method_name)
+        return if MethodReference.instance_method_visibility_for(object_singleton_class, @method_name) == @original_visibility
```

but I couldn't understand what the root cause is.

[...]

PR #2404, opened 2020-11-18
Move MRI build to Github Actions

benoittgt:
JRuby not moved yet because there is too many issues with it at the moment.

To do in another PR?
- [x] fix ruby 3
- [x] fix flaky test with "mountable_engine?" error
- [ ] ~~add JRuby~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,5 +89,5 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - run: script/clone_all_rspec_repos
-      - run: bundle binstubs rspec rspec-core rake cucumber rails
+      - run: bundle install --binstubs
       - run: script/run_build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,6 @@ jobs:
        include:
          # Rails 6.0 builds >= 2.5.0
          - ruby: 3.0.0-preview1
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 6.0.0'
          - ruby: 2.7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,95 @@
+name: RSpec CI
+on:
+  push:
+    branches:
+      - 'main'
+      - '*-maintenance'
+      - '*-dev'
+  pull_request:
+    branches:
+      - '*'
+jobs:
+  test:
+    name: 'Ruby: ${{ matrix.ruby }}, Rails: ${{ matrix.env.RAILS_VERSION }}'
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+       include:
+         # Rails 6.0 builds >= 2.5.0
+         - ruby: 3.0.0-preview1
+           allow_failure: true
+           env:
+             RAILS_VERSION: '~> 6.0.0'
+         - ruby: 2.7
+           env:
+             RAILS_VERSION: '~> 6.0.0'
+         - ruby: 2.6
+           env:
+             RAILS_VERSION: '~> 6.0.0'
+         - ruby: 2.5
+           env:
+             RAILS_VERSION: '~> 6.0.0'
+         # Rails 5.2 builds >= 2.2.2
+         - ruby: 2.6.6
+           env:
+             RAILS_VERSION: '~> 5.2.0'
+         - ruby: 2.5.8
+           env:
+             RAILS_VERSION: '~> 5.2.0'
+         - ruby: 2.4.10
+           env:
+             RAILS_VERSION: '~> 5.2.0'
+         - ruby: 2.3.8
+           env:
+             RAILS_VERSION: '~> 5.2.0'
+         - ruby: 2.2.10
+           allow_failure: true
+           env:
+             RAILS_VERSION: '~> 5.2.0'
+         - ruby: 2.2.10
+           env:
+             RAILS_VERSION: '5-2-stable'
+         # Rails 5.1 Builds >= 2.2.2
+         - ruby: 2.6.6
+           env:
+             RAILS_VERSION: '~> 5.1.0'
+         - ruby: 2.5.8
+           env:
+             RAILS_VERSION: '~> 5.1.0'
+         - ruby: 2.4.10
+           env:
+             RAILS_VERSION: '~> 5.1.0'
+         - ruby: 2.3.8
+           env:
+             RAILS_VERSION: '~> 5.1.0'
+         - ruby: 2.2.10
+           env:
+             RAILS_VERSION: '~> 5.1.0'
+         # Rails 5.0 Builds >= 2.2.2
+         - ruby: 2.6.6
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.5.8
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.4.10
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.3.8
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.2.10
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: false
+      - run: script/update_rubygems_and_install_bundler
+      - run: script/clone_all_rspec_repos
+      - run: bundle install --binstubs && script/run_build
+        env: ${{ matrix.env }}
+        continue-on-error: ${{ matrix.allow_failure || false }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,16 +81,13 @@ jobs:
          - ruby: 2.2.10
            env:
              RAILS_VERSION: '~> 5.0.0'
+    env: ${{ matrix.env }}
+    continue-on-error: ${{ matrix.allow_failure || false }}
     steps:
       - uses: actions/checkout@v2
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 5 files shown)


pirj (approved):
Wonderful! Thanks!

PS `mountable_engine?` strikes back. Unrelated, it was the case for long.

benoittgt:
Ok. For the flaky test. I nearly find it.

If you have this
```ruby
# spec/generators/rspec/scaffold/scaffold_generator_spec.rb
# Generators are not automatically loaded by Rails
require 'generators/rspec/scaffold/scaffold_generator'
require 'support/generators'

RSpec.describe Rspec::Generators::ScaffoldGenerator, type: :generator do
  setup_default_destination

  describe 'standard request specs' do
    subject { file('spec/requests/posts_spec.rb') }

    describe 'with --no-request_specs' do
      before { run_generator %w[posts --no-request_specs] }
      it { is_expected.not_to exist }
    end

    describe 'in an engine' do
      before do
        allow_any_instance_of(::Rails::Generators::NamedBase).to receive(:mountable_engine?).and_return(true)
        run_generator %w[posts --request_specs]
      end

      it do
        is_expected.to contain('Engine.routes.url_helpers')
      end
    end
  end
end
```
`bundle exec rspec spec/generators/rspec/scaffold/scaffold_generator_spec.rb --seed 58348`, the stub have to be run first.

It seems the stub is leaking to other test. We are not able to reset properly to the right state after the engine test pass.

It's very late here. So maybe tomorrow I will try to keep digging, but if you have an idea... Maybe it is related to [ammeter](https://github.com/alexrothenberg/ammeter)?

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
```diff
@@ -40,7 +40,7 @@
       it { is_expected.to contain('renders a JSON response with errors for the post') }
     end
 
-    describe 'in an engine' do
+    xdescribe 'in an engine' do
```
JonRowe, line 43: This suggests its a load order issue with the generators right?

benoittgt: Yes. The call to this https://github.com/rspec/rspec-rails/blob/5d2c5eddd09a476baeca176ea2f596cbf290504c/spec/generators/rspec/scaffold/scaffold_generator_spec.rb#L45 breaks test in the same file that come after `describe 'in an engine'`.

Like if the stub was not restoring properly the behavior of `mountable_engine?`.

JonRowe: Try running it in a sub process, at least that should isolate any poisoning of the globals.

benoittgt: Good idea! It works!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
...
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.2.10
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: false
```
benoittgt, line 90: I had weird surprise when it was `true`. Like

Capture d’écran 2020-11-17 à 21 47 57

.github/workflows/ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - run: script/update_rubygems_and_install_bundler
+      - run: script/clone_all_rspec_repos
+      - run: bundle install --binstubs && script/run_build
+        env: ${{ matrix.env }}
```
benoittgt, line 94: I had it only here, not sure it is needed before (CI pass). I saw Jon that you add it to all steps.

.github/workflows/ci.yml
pirj: 😬

.travis.yml
```diff
@@ -38,66 +38,23 @@ script: "script/run_build 2>&1"
...
-      env: RAILS_VERSION='~> 5.0.0'
-    - rvm: 2.2.10
-      env: RAILS_VERSION='~> 5.0.0'
 
   fast_finish: true
+
+branches:
+  only:
+    - master
+    - /^\d+-\d+-maintenance$/
```
pirj, line 60: `-dev` as well?

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
```diff
@@ -41,12 +43,13 @@
     end
 
     describe 'in an engine' do
-      before do
-        allow_any_instance_of(::Rails::Generators::NamedBase).to receive(:mountable_engine?).and_return(true)
-        run_generator %w[posts --request_specs]
+      it 'generates files with Engine url_helpers' do
+        in_sub_process do
```
pirj, line 47: 👏 

Wondering why it was wiping out the original method along the way.

benoittgt: I don't know 🤔

.travis.yml
benoittgt: https://github.com/rspec/rspec-rails/pull/2404/commits/6b7eec040692ce4cf6a6dc9fcfdff7a2aec3a54a

.github/workflows/ci.yml
JonRowe: It's because of the cache name

.github/workflows/ci.yml
JonRowe: You need it for each step where it's expected, I'm looking into a similar issue to you with versions.

.github/workflows/ci.yml
benoittgt: You mean because we use the same ruby version ? 🤔 

I read this https://github.com/ruby/setup-ruby/blob/master/action.yml and didn't find much more information to handle it properly.

JonRowe: Can you link a build here https://github.com/ruby/setup-ruby/issues/112

.github/workflows/ci.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: false
+      - run: script/update_rubygems_and_install_bundler
+        env: ${{ matrix.env }}
+      - run: script/clone_all_rspec_repos
+        env: ${{ matrix.env }}
+      - run: bundle install --binstubs && script/run_build
+        env: ${{ matrix.env }}
+        continue-on-error: ${{ matrix.allow_failure || false }}
```
JonRowe, line 96:
```suggestion
    env: ${{ matrix.env }}
    continue-on-error: ${{ matrix.allow_failure || false }}
    steps:
      - uses: actions/checkout@v2
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
      - run: script/clone_all_rspec_repos
      - run: bundle binstubs rspec rspec-core rake cucumber rails
      - run: script/run_build
```

.github/workflows/ci.yml
```diff
@@ -89,5 +89,6 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
       - run: script/clone_all_rspec_repos
+      - run: bundle
```
JonRowe, line 92: I had to restore the script to update bundler to get the main builds working 🤔 😬

.github/workflows/ci.yml
```diff
@@ -0,0 +1,93 @@
...
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.3.8
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+         - ruby: 2.2.10
+           env:
+             RAILS_VERSION: '~> 5.0.0'
+    env: ${{ matrix.env }}
+    continue-on-error: ${{ matrix.allow_failure || false }}
```
benoittgt, line 85: I doesn't seem to work. :/
https://github.com/rspec/rspec-rails/runs/1438441078?check_suite_focus=true

JonRowe:
```
SyntaxError: /home/runner/work/rspec-rails/bundle/ruby/2.2.0/gems/actionpack-5.2.4.4/lib/action_dispatch/request/session.rb:96: syntax error, unexpected '.'
          id&.public_id
```
That build is failing genuinely.

JonRowe: I think they introduced a 2.2.x bug for 5.2... `&.` is 2.3 and above.

pirj: We have this in 3-9-maintenance https://github.com/rspec/rspec-rails/commit/6a8fd077f14c791a7ef39f596735a7d867335a9e#diff-6ac3f79fc25d95cd1e3d51da53a4b21b939437392578a35ae8cd6d5366ca5485R58

pirj: Oh, sorry, I've missed 2.2.2 is tested against 5-2-stable already.

benoittgt: > I think they introduced a 2.2.x bug for 5.2... &. is 2.3 and above.

Yes completely and I didn't see a patch to change that ATM.

benoittgt:
I think the PR can be merged. I will have extra work for 6.1.0rc1 that I will probably do in https://github.com/rspec/rspec-rails/pull/2400

I will try to move JRuby over GA in another PR.

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
JonRowe: I suspect, and this is just a theory on my part, because when it stubs the method doesn't exist, it's then loaded, and the then stub blows it away as it was never there originally.

pirj:
:tada:

benoittgt:
Yeehaa

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
pirj: :exploding_head: :scream:
Reminds me of https://github.com/rspec/rspec-rails/issues/2394#issuecomment-706372093

PR #2406, opened 2020-11-23
Renable cache because we set env variables before the steps

benoittgt:
See: https://github.com/ruby/setup-ruby/issues/112#issuecomment-731607162

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
       - run: script/clone_all_rspec_repos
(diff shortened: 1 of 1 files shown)


JonRowe:
The default is `true` isn't it? The failures seem to be rubocop related, did it upgrade?

benoittgt:
> The default is true isn't it? The failures seem to be rubocop related, did it upgrade?

You're write. https://github.com/ruby/setup-ruby/blob/a699edbce608a2c128dedad88e3b6a0e28687b3c/index.js#L14

Also

> Note that any step doing bundle install (for the root Gemfile) or gem install bundler can be removed with bundler-cache: true.

I should maybe avoid having `bundle install` after... but I need few binstubs and I will probably have the same issues with Rubocop.

benoittgt:
Close for now because it's stale.

PR #2398, opened 2020-10-15
Support Rails 6.1

JonRowe:
This branch will become the root of Rails 6.1 development. The plan would be to launch this as a new version when Rails 6.1 is released, likely a major version as Rails versions supported will change.

*Edit* This will not be merged until we are ready to release the next version, it will be used as an available branch for those working on 6.1.

Todo:
- [x] View specs have changed behaviour surrounding helper modules, investigate and fix.
- [x] A controller spec has a broken config somehow
- [x] Generated controller spec "uploads" fails

PR #2408, opened 2020-12-03
Make rspec-rails compatible with Rails v6.1.0.rc2

benoittgt:
This PR add:

* Rails v6.1.0.rc2 to the CI build
* Fix 3 errors encountered when running the CI

Answer: https://github.com/rspec/rspec-rails/pull/2398#issuecomment-737757306

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,18 +18,17 @@ jobs:
        include:
          # Rails 6.1 builds >= 2.5
          - ruby: ruby-3.0.0-preview1
-           allow_failure: true
            env:
-             RAILS_VERSION: 'master'
+             RAILS_VERSION: '6.1.0.rc2'
          - ruby: 2.7.1
            env:
-             RAILS_VERSION: 'master'
+             RAILS_VERSION: '6.1.0.rc2'
          - ruby: 2.6.6
            env:
-             RAILS_VERSION: 'master'
+             RAILS_VERSION: '6.1.0.rc2'
          - ruby: 2.5.8
            env:
-             RAILS_VERSION: 'master'
+             RAILS_VERSION: '6.1.0.rc2'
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -16,9 +16,8 @@ jobs:
       fail-fast: false
       matrix:
        include:
-         # Rails 6.1 builds >= 2.5
+         # Rails Master builds >= 2.5
          - ruby: ruby-3.0.0-preview1
-           allow_failure: true
```
JonRowe, line 21: If we're adding Rails master builds as well as rc builds master should be allowed to fail.

example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -14,15 +14,10 @@
...
-  gsub_file 'config/initializers/action_mailer.rb',
-            /ExampleApp/,
-            Rails.application.class.parent.to_s
 
   copy_file 'spec/support/default_preview_path'
   chmod 'spec/support/default_preview_path', 0755
-  gsub_file 'spec/support/default_preview_path',
-            /ExampleApp/,
-            Rails.application.class.parent.to_s
+
```
JonRowe, line 20: I'm convinced these are still needed, its just the Rails 6.1 implementation might need to change?

benoittgt: I deep dive a little bit on this few weeks ago. The `Rails.application.class.parent` on previous Rails is `ExampleApp`.  So I didn't understand why the gsub was needed.
Also if the CI pass isn't what we want?

.github/workflows/ci.yml
benoittgt: Ok

JonRowe:
Closing as this has been mostly merged.

PR #2413, opened 2020-12-11
Update builds for 6 1

JonRowe:


benoittgt (approved):

example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -14,15 +14,20 @@
...
-  gsub_file 'config/initializers/action_mailer.rb',
-            /ExampleApp/,
-            Rails.application.class.parent.to_s
+
+  rails_parent =
+    if Rails.version.to_f >= 6.0
+      Rails.application.class.module_parent.to_s
+    else
+      Rails.application.class.parent.to_s
+    end
```
JonRowe, line 23: @benoittgt this is what replaced `parent` in 6.0 and above, I'm not sure if the 6.1 failure is real or a flake, I'll restart it.

JonRowe:
Merged because its more green than currently, but need to fix that last spec this weekend, as well as solve the assert no error issue, to get a release out.

example_app_generator/generate_action_mailer_specs.rb
benoittgt: Nice finding!

PR #2418, opened 2020-12-16
Correct route_to documentation in README

meshulam:
The `route_to` matcher [delegates to assert_recognizes](https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/matchers/routing_matchers.rb#L27). This assertion only tests path=>params, not that the provided params also generate the expected path [as `assert_routing` does](https://api.rubyonrails.org/v6.1.0/classes/ActionDispatch/Assertions/RoutingAssertions.html). 

It would be great to have an rspec matcher for the inverse (`generate_route` perhaps?) but at the very least I figured the readme should be corrected. The full docs on relishapp.com seem to be correct already.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,17 +187,17 @@ In addition to [the matchers that come standard in RSpec][],
 here are some extras that make it easier
 to test the various parts of a Rails system:
 
-| RSpec matcher            | Delegates to      | Available in                    | Notes                                                    |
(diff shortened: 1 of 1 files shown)


benoittgt (approved):
Thanks @meshulam for this change. LGTM

JonRowe (approved):

JonRowe:
Thanks!

PR #2422, opened 2020-12-21
Set RSpec version from env

JonRowe:
@benoittgt and @pirj the intent here is to set RSpec's desired version from the other repos, and then drop the `pre` check in the line below.

This will then allow people to install pre-release versions of `rspec-rails` from Github directly, without other gems. We can always bump up the minimum required version, or otherwise alter this check should be we need to pin, but with `rspec-rails` being independent now I think this is an improvement.


benoittgt (approved):
LGTM but shouldn't we document it? 😊

JonRowe:
We don't document a lot of CI things, the variable is documented in the workflow, but we can add a comment here when I drop the old check?

PR #2423, opened 2020-12-22
Fix WrongScopeError

### LINKED ISSUE #2417: WrongScopeError: `name` when `use_active_record = false`
I have the following spec:

```
require 'rails_helper'

describe Command do
  it 'foo' do
    Command.new
  end
end
```

Using RSpec 3.9.1 + Rails 6.1 I get:

```
RSpec::Core::ExampleGroup::WrongScopeError: `name` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).

  0) Command foo
     Failure/Error:
       raise WrongScopeError,
             "`#{name}` is not available from within an example (e.g. an " \
             "`it` block) or from constructs that run in the scope of an " \
             "example (e.g. `before`, `let`, etc). It is only available " \
             "on an example group (e.g. a `describe` or `context` block)."

       `name` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb:746:in `method_missing'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/activerecord-6.1.0/lib/active_record/test_fixtures.rb:102:in `run_in_transaction?'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/activerecord-6.1.0/lib/active_record/test_fixtures.rb:116:in `setup_fixtures'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/activerecord-6.1.0/lib/active_record/test_fixtures.rb:10:in `before_setup'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-rails-3.9.0/lib/rspec/rails/adapters.rb:126:in `block (2 levels) in <module:MinitestLifecycleAdapter>'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:450:in `instance_exec'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:450:in `instance_exec'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/hooks.rb:381:in `execute_with'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/hooks.rb:614:in `block (2 levels) in run_around_example_hooks_for'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:345:in `call'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/hooks.rb:615:in `run_around_example_hooks_for'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/hooks.rb:472:in `run'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:460:in `with_around_example_hooks'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:503:in `with_around_and_singleton_context_hooks'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example.rb:254:in `run'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb:633:in `block in run_examples'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb:629:in `map'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb:629:in `run_examples'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/example_group.rb:595:in `run'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:121:in `block (3 levels) in run_specs'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:121:in `map'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:121:in `block (2 levels) in run_specs'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/configuration.rb:2031:in `with_suite_hooks'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:116:in `block in run_specs'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/reporter.rb:74:in `report'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:115:in `run_specs'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:89:in `run'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:71:in `run'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/lib/rspec/core/runner.rb:45:in `invoke'
     # /home/mgarcia/.rbenv/versions/2.5.7/lib/ruby/gems/2.5.0/gems/rspec-core-3.9.1/exe/rspec:4:in `<top (required)>'
     # /home/mgarcia/.rbenv/versions/2.5.7/bin/rspec:23:in `load'
     # /home/mgarcia/.rbenv/versions/2.5.7/bin/rspec:23:in `<main>'
     # 
     #   Showing full backtrace because every line was filtered out.
     #   See docs for RSpec::Configuration#backtrace_exclusion_patterns and
     #   RSpec::Configuration#backtrace_inclusion_patterns for more information.
```

pirj:
Fixes #2417
Related: #2215, https://github.com/rails/rails/pull/37770

To test locally:
(pass)
```
bundle exec rake snippets
```

(fails, without the fix for `WrongScopeError`)
```
git checkout HEAD~1
bundle exec rake snippets
```

See https://github.com/rspec/rspec-rails/issues/2417#issuecomment-749598704 for an explanation of at least one of the possible causes

### DIFF
--- a/BUILD_DETAIL.md
+++ b/BUILD_DETAIL.md
@@ -66,6 +66,21 @@ $ bundle exec cucumber
 $ bin/cucumber
 ```
 
+## Snippets
+
+RSpec Rails uses snippets, self-contained examples that are used to cover
+cases and regressions that don't need a full-blown example application to
+reproduce.
+
+Snippets reuse the already installed gems, and don't attempt to install gem
+versions that are not on the system already to prevent version mismatches.
+
+Run with:
+
+```
+$ script/run_snippets.sh
+```
+
 ## YARD documentation
 
 RSpec uses [YARD](https://yardoc.org/) for API documentation on the [rspec.info site](https://rspec.info/).
--- a/Gemfile-sqlite-dependencies
+++ b/Gemfile-sqlite-dependencies
@@ -1,0 +1,20 @@
+version_file = File.expand_path('.rails-version', __dir__)
+RAILS_VERSION = ENV['RAILS_VERSION'] || (File.exist?(version_file) && File.read(version_file).chomp) || ""
+
+MAJOR =
+  case RAILS_VERSION
+  when /5-2-stable/
+    5
+  when /master/, /stable/, nil, false, ''
+    6
+  else
+    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
+  end
+
+if MAJOR >= 6
+# sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
+  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
+else
+# Similarly, Rails 5.0 only supports '~> 1.3.6'. Rails 5.1-5.2 support '~> 1.3', '>= 1.3.6'
+  gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
+end
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -9,6 +9,13 @@ module RSpec
         include RSpec::Rails::MinitestAssertionAdapter
         include ActiveRecord::TestFixtures
 
+        if ::Rails.version.to_f >= 6.1
+          # @private return the example name for TestFixtures
+          def name
+            @example
+          end
+        end
+
         included do
           if RSpec.configuration.use_active_record?
             include Fixtures
@@ -48,13 +55,6 @@ module RSpec
                   orig_implementation.bind(self).call(*args, &blk)
                 end
               end
-            end
-          end
-
-          if ::Rails.version.to_f >= 6.1
-            # @private return the example name for TestFixtures
-            def name
-              @example
             end
           end
         end
--- a/snippets/use_active_record_false.rb
+++ b/snippets/use_active_record_false.rb
@@ -1,0 +1,61 @@
+if __FILE__ =~ /^snippets/
+  fail "Snippets are supposed to be run from their own directory to avoid side " \
+       "effects as e.g. the root `Gemfile`, or `spec/spec_helpers.rb` to be " \
+       "loaded by the root `.rspec`."
+end
+
+# We opt-out from using RubyGems, but `bundler/inline` requires it
+require 'rubygems'
+
+require "bundler/inline"
+
+# We pass `false` to `gemfile` to skip the installation of gems,
+# because it may install versions that would conflict with versions
+# from the main `Gemfile.lock`.
+gemfile(false) do
+  source "https://rubygems.org"
+
+  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+
+  # Those Gemfiles carefully pick the right versions depending on
+  # settings in the ENV, `.rails-version` and `maintenance-branch`.
+  Dir.chdir('..') do
+    eval_gemfile 'Gemfile-sqlite-dependencies'
+    # This Gemfile expects `maintenance-branch` file to be present
+    # in the current directory.
+    eval_gemfile 'Gemfile-rspec-dependencies'
+    # This Gemfile expects `.rails-version` file
+    eval_gemfile 'Gemfile-rails-dependencies'
+  end
+
+  gem "rspec-rails", path: "../"
+end
+
+# Run specs at exit
+require "rspec/autorun"
+
+# This snippet describes the case when ActiveRecord is loaded, but
+# `use_active_record` is set to `false` in RSpec configuration.
+
+# Initialization
+require "active_record/railtie"
+require "rspec/rails"
+
+# This connection will do for database-independent bug reports
+ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
+
+# RSpec configuration
+RSpec.configure do |config|
+  config.use_active_record = false
+end
+
+# Rails project code
+class Foo
+end
+
+# Rails project specs
+RSpec.describe Foo do
+  it 'does not not break' do
+    Foo
+  end
+end
--- a/Changelog.md
+++ b/Changelog.md
@@ -26,6 +26,7 @@ Bug Fixes:
   (Phil Pirozhkov, Jon Rowe, #2353, #2354)
 * Remove old #fixture_path feature detection code which broke under newer Rails.
   (Koen Punt, Jon Rowe, #2370)
+* Fix an error when `use_active_record` is `false` (Phil Pirozhkov, #2423)
 
 ### 4.0.1 / 2020-05-16
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v4.0.0...v4.0.1)
--- a/Changelog.md
+++ b/Changelog.md
@@ -16,6 +16,7 @@ Enhancements:
   (Andrew W. Lee, #2372)
 * Improve request spec "controller" scafold when no action is specified.
   (Thomas Hareau, #2399)
+* Introduce testing snippets concept (Phil Pirozhkov, Benoit Tigeot, #2423)
 
 Bug Fixes:
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,4 @@
 source "https://rubygems.org"
-version_file = File.expand_path('.rails-version', __dir__)
-RAILS_VERSION = ENV['RAILS_VERSION'] || (File.exist?(version_file) && File.read(version_file).chomp) || ""
 
 gemspec
 
@@ -22,28 +20,11 @@ end
 
 gem 'capybara'
 
-MAJOR =
-  case RAILS_VERSION
-  when /5-2-stable/
-    5
-  when /master/, /stable/, nil, false, ''
-    6
-  else
-    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
-  end
-
-if MAJOR >= 6
-  # sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
-  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
-else
-  # Similarly, Rails 5.0 only supports '~> 1.3.6'. Rails 5.1-5.2 support '~> 1.3', '>= 1.3.6'
-  gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
-end
-
 # Until 1.13.2 is released due to Rubygems usage
 gem 'ffi', '~> 1.12.0'
 
 custom_gemfile = File.expand_path('Gemfile-custom', __dir__)
 eval_gemfile custom_gemfile if File.exist?(custom_gemfile)
 
+eval_gemfile 'Gemfile-sqlite-dependencies'
 eval_gemfile 'Gemfile-rails-dependencies'
--- a/script/run_build
+++ b/script/run_build
@@ -20,6 +20,8 @@ fi
 
 fold "cukes" run_cukes
 
+fold "snippets" script/run_snippets.sh
+
 if documentation_enforced; then
   fold "doc check" check_documentation_coverage
 fi
--- a/script/run_snippets.sh
+++ b/script/run_snippets.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+set -e
+
+(
+  cd snippets
+  # This is required to load `bundle/inline`
+  unset RUBYOPT
+  for snippet in *.rb;
+  do
+    echo Running $snippet
+    ruby $snippet
+  done
+)


pirj:
I fail to reproduce it in tests so far. Will take a look later this week unless you guys want to take over @benoittgt @JonRowe 

For [this example app](https://github.com/bmulholland/rspec-rails-6-1) the following fixed the failure (`lib/rspec/rails/fixture_support.rb`:
```diff
       module RSpec
         module Fixtures
           extend ActiveSupport::Concern
 
+          if ::Rails.version.to_f >= 6.1
+            # @private return the example name for TestFixtures
+            def name
+              @example
+            end
+          end
+
           class_methods do
             def fixtures(*args)
               orig_methods = private_instance_methods
@@ -50,13 +57,6 @@ module RSpec
               end
             end
           end
-
-          if ::Rails.version.to_f >= 6.1
-            # @private return the example name for TestFixtures
-            def name
-              @example
-            end
-          end
         end
       end
     end
```

JonRowe:
This is a standalone reproduction that can be run in a rails app generated with `rails new --minimal`
```
require File.expand_path('../config/environment', __dir__)
ENV['RAILS_ENV'] ||= 'test'
require 'rspec/rails'

RSpec.configure do |config|
  config.use_active_record = false
end

class MyClass; end

RSpec.describe MyClass do
  it 'foo' do
    MyClass.new
  end
end
```

benoittgt:
I will try to look at it if I have time around Christmas.

benoittgt:
It seems that when people are using `use_active_record = false` they can have a rails app with still lot's of active_record loaded.

To reproduce the same behavior as in #2417. I had to de-comment `require "active_record/railtie"` in `application.rb`, then re-ad a `database.yml` then re-add sqlite in the Gemfile. 

I think the patch should be validated with another specs, closer to file_fixtures. I will keep searching for another proposal for your patch @pirj, if you have any idea. Feel free. :)

pirj:
Awesome find, @benoittgt !
I was stuck on adding a failing spec for that fix. Sounds like we could use yet another `example_app_generator` app, with AR (accidentally left) but `use_active_record = false` in our tests.
Did the fix above work for this reproduction snippet?

benoittgt:
Hello @pirj 
>  Did the fix above work for this reproduction snippet?

Yes, I can confirm it is working.

I am wondering if adding a new rails app isn't too much. 🤔  Trying to write a non regressive test similar to https://github.com/rails/rails/commit/d4367eb72601f6e5bba3206443e9d141e9753af8#diff-76f5025b3bc6be04d08bb7df8f22ee2f0c69afe58aa4ba4933e6d6b9c328dc21R35-R38 could be an idea? What do you think?

JonRowe:
I'd like us to be able to run snippets like @benoittgt's above within our build, I think they are easier to grok than the whole app builds for one spec; we shouldn't get ride of the app builds, they are important to demonstrate working generators etc but for one off issues like this it's easier if we can do `ruby some_file.rb` to test something

benoittgt:
I love the idea. I wrote something similar recently :

The spec: https://github.com/WeTransfer/format_parser/blob/master/spec/active_storage/rails_app_spec.rb
The rails app: https://github.com/WeTransfer/format_parser/blob/master/spec/integration/active_storage/rails_app.rb

Do you think about something similar? If yes I can import similar code to rspec-rails code base.

benoittgt:
Curious about your point of view on my last comment.

For the initial PR from @pirj I think we should try to at least add a spec on fixture_support ?

pirj:
Awesome idea, @benoittgt ! If we reuse already installed gems and not re-download them, that should also work pretty quickly.
Not to say that otherwise our `example_app_generator` is getting pretty crowded.

pirj:
@JonRowe @benoittgt Implemented the concept of testing snippets, added them to the build.

To test locally:
(pass)
```
bundle exec rake snippets
```

(fails, without the fix for `WrongScopeError`)
```
git checkout HEAD~1
bundle exec rake snippets
```

pirj:
Snippets seem to be reusing `spec/spec_helper.rb` from our repo, and we don't want that.
A quick attempt to isolate it
```ruby
   Dir.chdir('snippets') do
     Dir['*.rb'].each do |snippet|
       sh "echo Running #{snippet}"
       sh "ruby #{snippet}"
     end
    end
```
failed with:
```
[!] There was an error parsing `Gemfile`:
[!] There was an error parsing `Gemfile-rspec-dependencies`: No such file or directory @ rb_sysopen - /home/pirj/source/rspec-rails/snippets/maintenance-branch. Bundler cannot continue.

 #  from Gemfile-rspec-dependencies:1
 #  -------------------------------------------
 >  branch = File.read(File.expand_path("../maintenance-branch", __FILE__)).chomp
```

Adding an `.rspec` that doesn't `--require spec_helper` to `snippets/` doesn't help, neither does adding a `snippets/spec_helper.rb`.

It also seems that `sh ruby <snippet>` from Rake suffers from the cached `BUNDLE_GEMFILE` issue, we have to reset it, otherwise, an attempt to load `Gemfile` is made. `unset BUNDLE_GEMFILE` doesn't help.

benoittgt:
Great proposal. I love the idea to easily test behavior separately. 

I had the similar issue with the `spec_helper` but in my case it was related to ammeter and loading issue. I will try to play tonight with it if you didn't find a solution yet.

benoittgt:
I push a small change. Locally it pass. I have some duplication on SQLite because we do not `eval` `Gemfile`. This could be moved somewhere I think, or dedicated to a file we could eval in both scenarios.

@pirj fill free to remove/squash it if you have better ideas. 😊

Rakefile
```diff
@@ -224,7 +224,15 @@ end
 
 task acceptance: ['smoke:app', 'no_active_record:smoke:app', :cucumber]
 
-task default: [:spec, :acceptance]
+desc "Run short, self-contained snippets"
+task :snippets do
+  sh "echo Running snippets"
+  sh <<-SH
+     ruby snippets/use_active_record_false.rb
```
JonRowe, line 231: Does this definitely fail if it er... fails?

Gemfile-sqlite-dependencies
```diff
@@ -0,0 +1,20 @@
...
+    /(\d+)[\.|-]\d+/.match(RAILS_VERSION).captures.first.to_i
+  end
+
+if MAJOR >= 6
+# sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
+  gem 'sqlite3', '~> 1.4', platforms: [:ruby]
+else
+# Similarly, Rails 5.0 only supports '~> 1.3.6'. Rails 5.1-5.2 support '~> 1.3', '>= 1.3.6'
+  gem 'sqlite3', '~> 1.3.6', platforms: [:ruby]
+end
```
JonRowe, line 20: Why?

JonRowe: Ah nm, so you can customise it later.

snippets/use_active_record_false.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  # Those Gemfiles carefully pick the right versions depending on
+  # settings in the ENV, `.rails-version` and `maintenance-branch`.
+  eval_gemfile './Gemfile-sqlite-dependencies'
+  eval_gemfile './Gemfile-rspec-dependencies'
+  eval_gemfile './Gemfile-rails-dependencies'
+
+  gem "ammeter"
+  gem "rspec-rails", path: "./"
+end
```
JonRowe, line 16: Do we want to eventually extract this to `snipets/gemfile.rb` or similar? And require that?

snippets/use_active_record_false.rb
```diff
@@ -0,0 +1,39 @@
...
+# Run specs on exit
+require "rspec/autorun"
+
+# RSpec configuration
+RSpec.configure do |config|
+  config.use_active_record = false
+end
+
+# Rails project code
+class Command
```
JonRowe, line 31: I'd rather we did't use `Command` purely because this does happen with any name and `Command` sounds special, which its not

Rakefile
pirj: It does, yes.

snippets/use_active_record_false.rb
pirj: Sure, I'll rename it. I believe it was coming from the original reproduction snippet.

pirj:
@benoittgt Thanks a lot for fixing this, it seems to work!

There was a glitch,
```
/Users/pirj/.rvm/rubies/ruby-2.7.1/lib/ruby/2.7.0/bundler/runtime.rb:312:in `check_for_activated_spec!': You have already activated loofah 2.8.0, but your Gemfile requires loofah 2.9.0. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```
but it resolved after:
```
$ bundle update --conservative loofah
Using loofah 2.9.0 (was 2.8.0)
```
It seems that if any of the dependencies with flexible version constraints update, we have to keep the repo's `Gemfile.lock` up to date to keep them in sync to avoid version mismatches.

I'll play around with it, maybe something like
```ruby
require 'bundle/setup'
```
helps.

pirj:
It seems it still loads `spec/spec_helper.rb`, if you remove `gem 'ammeter'`, it breaks.
Also, the `Gemfile`, still.

pirj:
So far:
```
cd snippets
ruby use_active_record_false.rb
```
works.
`Gemfile`/`spec_helper.rb` are not loaded by that snippet.

But the rake task doesn't work.

By running snippets from the root directory, I couldn't find a way not to load `.rspec` file that tells RSpec runner to also load `spec/spec_helper.rb` that has `require 'ammeter/init'` in it, making snippet runs less clean and dependent on `spec_helper`.

The only way to [skip loading `.rspec` is to pass `--custom_options_file` from the command line](
https://github.com/rspec/rspec-core/blob/6543181b4ae91e2ecc7e1a7385532b4aca716bcb/lib/rspec/core/configuration_options.rb#L122), but we run snippets as `ruby ...`, not `rspec ...`, so this seems to be a dead end.
One way would be to also accept `custom_options_file` from `ENV`, not just the command line, but I'm not sure if this is worth it just for this narrow use case.

There's also no way to reset `RSpec.configuration.requires`, as its setter uses `@requires += paths`. In any case, even `instance_variable_set(:@requires)` from the snippet won't work, since it sets `requires = 'spec_helper'` to the config in `at_exit`.

pirj:
Running `rake snippets` does something awful, after that:
```
$ bundle check

The following gems are missing
 * racc (1.5.2)
 * nio4r (2.5.4)
 * websocket-driver (0.7.3)
 * sqlite3 (1.4.2)
Install missing gems with `bundle install`
```
and only `bundle install` helps.

Tried with Bundler 2.1.4 (Ruby 2.7) and 2.2.3 (Ruby 3.0) with slightly different results. The former only claims that `racc` went missing.

I intend to skip adding snippets as a rake task, and leave it for the brave.

benoittgt:
I tried a little bit to fix that, but I think we wil have tol stay in shell script. I am surprised how much rake context is different from a basic terminal... :/

pirj:
@JonRowe @benoittgt This is ready.
To test that snippets work as a concept:
```
$ source script/custom_build_functions.sh
$ run_snippets
Running snippets
Running ./use_active_record_false.rb
.

Finished in 0.02198 seconds (files took 0.76826 seconds to load)
1 example, 0 failures

$ git checkout HEAD~1
$ run_snippets
Running snippets
Running ./use_active_record_false.rb
F

Failures:

  1) Foo does not not break
     Failure/Error:
       raise WrongScopeError,
             "`#{name}` is not available from within an example (e.g. an " \
             "`it` block) or from constructs that run in the scope of an " \
             "example (e.g. `before`, `let`, etc). It is only available " \
             "on an example group (e.g. a `describe` or `context` block)."

       `name` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
```

When running snippets, you can add any junk to `.rspec`, `Gemfile`, and `spec/spec_helper.rb` to test that they are not loaded.

I'm happy to shape snippets up further, e.g. turn [`example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb`](https://github.com/rspec/rspec-rails/blob/main/example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb) into a snippet if you have no objections on the idea (in a separate PR preferably).

pirj:
Moved to drafts, as I fail to require `bundler` from a script.

pirj:
Green (at least before squashing).

BUILD_DETAIL.md
```diff
@@ -66,6 +66,22 @@ $ bundle exec cucumber
...
+reproduce.
+
+Snippets reuse the already installed gems, and don't attempt to install gem
+versions that are not on the system already to prevent version mismatches.
+
+Run with:
+
+```
+$ cd snippets
+$ find . -name '*.rb' | xargs -I_ -L1 sh -c "echo Running _ && ruby _"
```
JonRowe, line 82: Do you want to extract that to a `script/run_snippets`?

script/custom_build_functions.sh
```diff
@@ -18,3 +18,16 @@ function run_cukes {
...
+    cd snippets
+    # This is required to load `bundle/inline`
+    unset RUBYOPT
+    for snippet in *.rb;
+    do
+      echo Running $snippet
+      ruby $snippet
+    done
+  )
+}
```
JonRowe, line 33: `script/run_snippets` could even just import this file and call the function :)

lib/rspec/rails/fixture_support.rb
```diff
@@ -9,6 +9,13 @@ module FixtureSupport
         include RSpec::Rails::MinitestAssertionAdapter
         include ActiveRecord::TestFixtures
 
+        if ::Rails.version.to_f >= 6.1
+          # @private return the example name for TestFixtures
+          def name
+            @example
+          end
+        end
```
JonRowe, line 17: This is just moved to change the visibility right?

JonRowe (approved):
Overall I approve, great work, WDYT about the suggested change to the script? And can you squash :)

script/custom_build_functions.sh
```diff
@@ -18,3 +18,16 @@ function run_cukes {
 function run_all_spec_suites {
   fold "one-by-one specs" run_specs_one_by_one
 }
+
+function run_snippets {
+  (
+    cd snippets
+    # This is required to load `bundle/inline`
+    unset RUBYOPT
```
benoittgt, line 26: Nice!

benoittgt (approved):
Well done! Super happy with that change.

lib/rspec/rails/fixture_support.rb
pirj: Yes, that's what has fixed the issue.

BUILD_DETAIL.md
pirj: Good idea. I've seen CI fail with a nonsense message about `_` in this statement, so using a `for` inside a single runnable script file is totally reasonable.
Done, and squashed.

PR #2425, opened 2020-12-23
Add a check to support both RSpec 3 & 4

pirj:
RSpec 4 has removed OperatorMatcher, as it was only supported by the should syntax, and not expect syntax.

See:
 - https://github.com/rspec/rspec-core/pull/2803
 - https://github.com/rspec/rspec-expectations/pull/1245


pirj:
@JonRowe Quick approve? This is needed for extraction of monkey patching.

PR #2440, opened 2021-01-07
Support RSpec v3.10.x in rspec-rails v3.x

hosamaly:
Just updated the version and maintenance branch to enable support for RSpec v3.10.x in rspec-rails v3.x so that it may be used in Rails < v5 and Ruby < v2.3 (which are no longer supported in rspec-rails v4.x).

### DIFF
--- a/lib/rspec/rails/version.rb
+++ b/lib/rspec/rails/version.rb
@@ -3,7 +3,7 @@ module RSpec
     # Version information for RSpec Rails.
     module Version
       # Current version of RSpec Rails, in semantic versioning format.
-      STRING = '3.9.1'
+      STRING = '3.10.0'
     end
   end
 end
--- a/maintenance-branch
+++ b/maintenance-branch
@@ -1,1 +1,1 @@
-3-9-maintenance
+3-10-maintenance


pirj:
You are probably looking to update this https://github.com/rspec/rspec-rails/blob/487fa59db3a65e5ffde3580224580d710c10cd2b/rspec-rails.gemspec#L49 instead.

@JonRowe It seems that `rspec-rails` 3.9.1 won't support RSpec 3.10/3.11/3.99 as we resolve RSpec dependencies to `"~> 3.9.0"`.

Would you agree to use `gem "rspec-*", "~> 3.9"`?

```diff
-      s.add_runtime_dependency "rspec-#{name}", "~> #{RSpec::Rails::Version::STRING.split('.')[0..1].concat(['0']).join('.')}"
+      s.add_runtime_dependency "rspec-#{name}", "~> #{RSpec::Rails::Version::STRING.split('.')[0..1].join('.')}"
```

JonRowe:
RSpec Rails 3.x is unsupported.

hosamaly:
@JonRowe I understand that it's unsupported but I'd like to use it for a client that is stuck on some old versions.

I wanted to open the PR against a new branch but GitHub doesn't allow me to do that. If you create such a branch, then I should be able to reopen the PR against the new branch.

Is there anything else I need to do?

JonRowe:
The older 3.9 versions still work, just use whats available on rubygems.

PR #2442, opened 2021-01-09
Conditionally use shared_context_metadata_behavior

pirj:
See https://github.com/rspec/rspec-core/pull/2834


pirj:
Going to merge it as soon as the build is green:
 - https://github.com/rspec/rspec-core/pull/2834 that depends on this is approved
 - it's an internal spec-only change

Issue #2452, opened 2021-02-01
No support in docs for rails 3.x


akcrono:
Following the directions in the [link for rails earlier than 5.0](https://github.com/rspec/rspec-rails/tree/3-9-maintenance) results in a a bundler error:

```
    rails (~> 3.2.0) was resolved to 3.2.22.5, which depends on
      actionpack (= 3.2.22.5)

    rspec-rails was resolved to 4.1.0.pre, which depends on
      actionpack (>= 4.2)
```

The docs should be updated to replace `'master'` with the correct maintenance branch name

pirj:
What is the problem you're experiencing with the following exactly?
```ruby
group :development, :test do
  gem 'rspec-rails', '~> 3.9'
end
```

akcrono:
That the other gems it depends on are not also current

pirj:
Can you please elaborate?

pirj:
I suggest you to skip declaring the version constraint, and let Bundler install the correct version of `rspec-rails` and dependent gems basing on the Rails version you use.

It's clearly a miss in the documentation, along with mentioning the now non-existent `master` branch (we're using `main`).
However, we're not too interested in keeping it up to date with other priorities in sight.

To keep you out of dependency conflicts, just keep in mind that `main` is 4.x now, and the most recent development in general. With an older version of Rails to work with, pick an earlier release from RubyGems.

You're welcome to send a PR to `main` and optionally to `rails-6-1-dev` branches to add

> Use rspec-rails 3.x for Rails 3 and 4.

next to

> Use rspec-rails 1.x for Rails 2.x.

JonRowe:
@pirj The issue is we link to the 3-9-maintenance readme from the `main` readme. So we should update that one to reflect correct instructions...

akcrono:
Skipping the version constraint results in a bundler error.  The only way I can update rspec for rails 3.2 is to explicitly define the maintenance branch via the block gem declaration.

And if it's clearly a miss in the documentation, then the request to update the documentation seems like a reasonable change to make.

JonRowe:
> The only way I can update rspec for rails 3.2 is to explicitly define the maintenance branch via the block gem declaration.

Surely not, this should work:

```
gem "rspec-rails", "~> 3.9.0"
```

akcrono:
It surely doesn't:

```Bundler could not find compatible versions for gem "rspec-mocks":
  In snapshot (Gemfile.lock):
[...]

PR #2453, opened 2021-02-01
Update docs to use the correct branch name for older rails versions

akcrono:
Addresses #2452

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,9 @@ Use **[`rspec-rails` 1.x][]** for Rails 2.x.
      gem 'rspec-rails', '~> 3.9'
    end
 
-   # Or, run against the master branch
-   # (requires master-branch versions of all related RSpec libraries)
-   # Replace branch name with the appropriate branch for your rails version
+   # Or, run against the maintenance branch
+   # (requires the same maintenance branch for all related RSpec libraries)
+   # Note that 3.9 is end of life and all changes have been released.
    group :development, :test do
      %w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
        gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => '3-9-maintenance'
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,9 +33,10 @@ Use **[`rspec-rails` 1.x][]** for Rails 2.x.
 
    # Or, run against the master branch
    # (requires master-branch versions of all related RSpec libraries)
+   # Replace branch name with the appropriate branch for your rails version
    group :development, :test do
      %w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
-       gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => 'master'
+       gem lib, :git => "https://github.com/rspec/#{lib}.git", :branch => '3-9-maintenance'
```
pirj, line 39: The latest on the `3-9-maintenance` branch is released and can be fetched from RubyGems with `~> 3.9` as described above.

akcrono: This isn't mentioned with the block gem declaration.

pirj:
Thank you for your contribution, but we do not maintain the previous versions.

JonRowe:
I think updating the readme is a reasonable ask here.

README.md
```diff
@@ -33,9 +33,10 @@ Use **[`rspec-rails` 1.x][]** for Rails 2.x.
 
    # Or, run against the master branch
    # (requires master-branch versions of all related RSpec libraries)
+   # Replace branch name with the appropriate branch for your rails version
```
JonRowe, line 36: This whole comment block should be:
```
# Or, run against the maintenance branch
# (requires the same maintenance branch for all related RSpec libraries)
```

JonRowe: Additionally prehaps we should add:

```
# Note that 3.9 is end of life and all changes have been released.
```

akcrono: Good points

PR #2454, opened 2021-02-02
`next` branch for users to follow the ongoing support for latest Rails

pirj:
This year is a bit busy, and with a lot of RSpec 4 development, it turned out we're late to release 4.1 to fully support Rails 6.1.

For the users it's not too obvious that they need to use the `rails-6-1-dev` branch to have all fixes and features for 6.1. Using `next` branch at all times with support for the "next" Rails versions would be easier for:

 - those who use the latest released Rails version that we do not fully support yet (e.g 4.1 for 6.1)
 - those who follow edge Rails versions from git

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'next'
       - '*-maintenance'
       - '*-dev'
   pull_request:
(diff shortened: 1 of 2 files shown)


pirj:
YARD fails with:
```
 [warn]: In file `README.md':34: Cannot resolve link to lib from text:
	...{lib}...
```
in a code block:
```
       gem lib, git: "https://github.com/rspec/#{lib}.git", branch: 'main'
```
I'll take a look later. I recall we have had this issue with JRuby previously.

JonRowe:
`main` is our `next` branch, I don't want another long running branch to keep open. We just need to fix the last few issues with Rails 6.1 and get it released, its my priority over rspec 4.

Issue #2451, opened 2021-02-01
`let(:name)` will conflict with Rails internals for fixtures


grekko:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.7.1
Rails version: 6.1.1
RSpec version: 4.0.2

## Observed behaviour

One of my specs started to fail while upgrading `rails` from `6.0` to `6.1`.

I noticed that a `let(:name)`-declaration in my spec was being called from [`ActiveRecord::TestFixtures#run_in_transaction?`](https://github.com/rails/rails/blob/d5386cfba2ec5c80f09e37d44b87bcae9c6ab6a8/activerecord/lib/active_record/test_fixtures.rb#L102) which is AFAIU supposed to call [`RSpec::Rails::FixtureSupport::Fixtures#name`](https://github.com/rspec/rspec-rails/blob/4b967e2e9ffb80e25a90e7b6704b939814a5dd79/lib/rspec/rails/fixture_support.rb#L56-L58) instead.

## Expected behaviour

I'd want to be warned if I'm redefining a reserved method via [`RSpec::Core::MemoizedHelpers#let`](https://github.com/rspec/rspec-core/blob/e95cfe3724de9b56f973e38f84fdb1b609d88e01/lib/rspec/core/memoized_helpers.rb#L306-L313) similar to how `initialize` is a reserved keyword.

## Can you provide an example app?

Start with a blank rails app (Version => `6.1`):

```sh
gem install rails -v '6.1.1'
rails new rspec-rails-issue
```

Install `rspec-rails` (Version >= `4.0.2`):

```sh
group :development, :test do
  gem 'rspec-rails', '~> 4.0.2'
end
```

Install RSpec:

```sh
rails generate rspec:install
```

Create `spec/some_spec.rb`:

```rb
require 'rails_helper'

RSpec.describe "observes a call to :name" do
  subject { true }

  let(:name) { puts caller; puts "I should never have been called" }

  it { is_expected.to be_truthy }
end
```

Run the spec:

```sh
rspec spec/some_spec.rb
[…path omitted]/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/rspec-core-3.10.1/lib/rspec/core/memoized_helpers.rb:317:in `block (2 levels) in let'
[...]

PR #2461, opened 2021-02-17
Prevent collisions on let(:name) and let(:method_name)

benoittgt:
Fix: https://github.com/rspec/rspec-rails/issues/2451

Two commits. One to see the failures then one to see the fix. I did nothing specific, just implement the [great work](https://github.com/rspec/rspec-rails/issues/2451#issuecomment-771213822) of @grekko and the choice from @JonRowe.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -9,11 +9,11 @@ module RSpec
         include RSpec::Rails::MinitestAssertionAdapter
         include ActiveRecord::TestFixtures
 
-        if ::Rails.version.to_f >= 6.1
-          # @private return the example name for TestFixtures
-          def name
-            @example
-          end
+        # @private prevent ActiveSupport::TestFixtures to start a DB transaction.
+        # Monkey patched to avoid collisions with 'let(:name)' in Rails 6.1 and after
+        # and let(:method_name) before Rails 6.1.
+        def run_in_transaction?
+          use_transactional_tests && !self.class.uses_transaction?(@example)
         end
 
         included do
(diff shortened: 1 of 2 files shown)


lib/rspec/rails/fixture_support.rb
```diff
@@ -9,11 +9,11 @@ module FixtureSupport
...
-        if ::Rails.version.to_f >= 6.1
-          # @private return the example name for TestFixtures
-          def name
-            @example
-          end
+        # @private prevent ActiveSupport::TestFixtures to start a DB transaction.
+        # Monkey patched to avoid collisions with 'let(:name)' in Rails 6.1 and after
+        # and let(:method_name) before Rails 6.1.
+        def run_in_transaction?
+          use_transactional_tests && !self.class.uses_transaction?(@example)
```
bcgraham, line 16: Not “reviewing,” just curious - were you able to determine whether any RSpec code still sets `@example` to a meaningful value?

pirj (approved):
Awesome, thank you!

lib/rspec/rails/fixture_support.rb
benoittgt: No I didn't search this yet. But I will.

JonRowe: For us this probably needs to be `example` or `self` as we don't assign `@example` within a test like mini test does afaik

pirj:
Rebased. Green.
Apart from ruby: 2.2.10 with Rails `5-2-stable`, probably flaky. Passed on `pull_request` trigger, failed on `push`.

PR #2464, opened 2021-02-18
[WiP] Update of request specs generator due to change in rails 6.1 controller default behaviour.

panSarin:
Since rails 6.1.1 changed default behaviour of CRUD response for invalid
attributes - to respond with 422 status, we have to update our scaffold
for generating specs for Create and Update with invalid attributes.

I updated both generators to check if controller responds with status
422 instead of 200.

Also, I didn't find any tests about those generators, so I couldn't update them - if there are some, please let me know so I can update them.

Linked issue: https://github.com/rspec/rspec-rails/issues/2463

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,4 +1,4 @@
-# This file was generated on 2020-12-23T21:24:00+01:00 from the rspec-dev repo.
+# This file was generated on 2020-12-25T18:48:30+00:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
 github: [JonRowe, benoittgt]
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,5 +1,5 @@
-# This file was generated on 2019-12-05T21:32:23+00:00 from the rspec-dev repo.
+# This file was generated on 2020-12-23T21:24:00+01:00 from the rspec-dev repo.
 # DO NOT modify it by hand as your changes will get lost the next time it is generated.
 
-github: [JonRowe]
+github: [JonRowe, benoittgt]
 open_collective: rspec
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,7 @@ jobs:
              RAILS_VERSION: '~> 6.1.0'
 
          # Rails 6.0 builds >= 2.5.0
-         - ruby: 3.0
+         - ruby: 3.0.0-rc1
            env:
              RAILS_VERSION: '~> 6.0.0'
          - ruby: 2.7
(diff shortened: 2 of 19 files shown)


pirj (changes requested):
We aim to support Rails 5.0, 5.1, 5.2 and 6.0 as well (actually, even 4.2 should probably work). Will the generator work in those versions correctly?

Do you mind adding a test that would run this generated test?
I'm mostly interested in reproducing the failure you observe in 6.1.1, that would fail without your fix.

panSarin:
Oh, so does it make sense to introduce a change that will work only with 6.1.1 and newer? Or maybe there should be some check in generator that check actual version of rails that called for generator? Please help me figure out how it should approach it and i will add tests and implement it. Or maybe since you didnt mention supporting 6.1, that should be frozen for now? Btw. thanks for such a quick response.

pirj:
https://github.com/rspec/rspec-rails/blob/a78519850c8c8c3cf86319b2cc0b214bf16a6372/example_app_generator/generate_stuff.rb#L97 should generate a controller and its specs. Those specs are run afterwards. It is worth checking what exactly is generated, and why generated request specs won't fail on 6.1.1.

panSarin:
Thanks for the hints! I am on it.

panSarin:
@pirj so I tried to test it by example app, and I encounter problems with generating example app for rails 6.1
I found there is opened https://github.com/rspec/rspec-rails/pull/2398 PR for that. Is it planned to be merged soon or I should create that PR for that branch?

Also to make that test fail there is one more step to reproduce: we have to add some model validation and fill invalid  attributes, because by default they are skipped and there is no validation, so we will never go into "bad path". 

But still I think that is a case to fix - pleace correct me if I am wrong.

pirj:
Ouch. Sorry, I forgot to point out that we have a branch for 6.1 fixes, `rails-6-1-dev`.
We've just one fix left to merge, https://github.com/rspec/rspec-rails/pull/2461, and hopefully, it is ready.
Your fix can be part of that branch. Please let me know if you can change the base branch:

or I can do it for you.

> and fill invalid attributes

I would suggest `gsub`'ing required files as we do here https://github.com/rspec/rspec-rails/blob/a78519850c8c8c3cf86319b2cc0b214bf16a6372/example_app_generator/generate_stuff.rb#L148

pirj:
That looks better! :+1: 
Just need a test to prove it fixes it for 6.1.1 and doesn't break for previous ones.

panSarin:
Test to prove it fixes it for 6.1.1 - what you mean by "test" there, some unit test, or more like recorded video where i generate app for 5.x and 6.0 versions and it generates old version ( which works ) and another for 6.1.1 where it works with new version ? 

I am not familiar with testing tools that generate tests ;]

pirj:
> rspec-rails/example_app_generator/generate_stuff.rb should generate a controller and its specs. Those specs are run afterwards. I would suggest gsub'ing those generated files

to include necessary validation for 422 to be triggered.
That's it!

panSarin:
Due to mess i had in commits and branches, i opened new clear PR so I am closing that and moving further conversation to https://github.com/rspec/rspec-rails/pull/2466

PR #2466, opened 2021-02-19
[WiP] Update of requests specs for invalid params for rails 6.1.1 and above

panSarin:
In order to make requests specs generated for rails 6.1.1 and above
working, we have to update expected behaviour of invalid params for create.

Since rails 6.1.1 is rendering 422 status instead of 302 like before, when
it re-renders form for invalid params, we changed expected behaviour in
request specs.

Also gstubbed files in example app so it will have proper validation
in widget that will be called in generated tests.

Issue related: https://github.com/rspec/rspec-rails/issues/2463

### DIFF
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -167,4 +167,17 @@ gsub_file 'spec/requests/gadgets_spec.rb',
 gsub_file 'spec/controllers/uploads_controller_spec.rb',
           'skip("Add a hash of attributes valid for your model")',
           '{}'
+
+# adds validation to widget so our test for happy and unhappy path will be ready to test validation
+gsub_file 'spec/requests/widgets_spec.rb',
+          'skip("Add a hash of attributes invalid for your model")',
+          '{ name: "" }'
+
+gsub_file 'spec/requests/widgets_spec.rb',
+          'skip("Add a hash of attributes valid for your model")',
+          '{ name: "lorem" }'
+
+gsub_file 'app/models/widget.rb',
+          'end',
+          'validates :name, presence: true; end'
 final_tasks
--- a/lib/generators/rspec/scaffold/templates/request_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/request_spec.rb
@@ -118,11 +118,19 @@ RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %
     end
 
     context "with invalid parameters" do
+    <% if Rails.version.to_f < 6.1 || Rails.version == '6.1.0' %>
       it "renders a successful response (i.e. to display the 'edit' template)" do
         <%= file_name %> = <%= class_name %>.create! valid_attributes
         patch <%= show_helper.tr('@', '') %>, params: { <%= singular_table_name %>: invalid_attributes }
         expect(response).to be_successful
       end
+     <% else %>
+      it "renders a response with 422 status - unporcessable entity" do
+        <%= file_name %> = <%= class_name %>.create! valid_attributes
+        patch <%= show_helper.tr('@', '') %>, params: { <%= singular_table_name %>: invalid_attributes }
+        expect(response.status).to eq(422)
+      end
+    <% end %>
     end
   end
 
--- a/lib/generators/rspec/scaffold/templates/request_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/request_spec.rb
@@ -82,10 +82,17 @@ RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %
         }.to change(<%= class_name %>, :count).by(0)
       end
 
+    <% if Rails.version.to_f < 6.1 || Rails.version == '6.1.0' %>
       it "renders a successful response (i.e. to display the 'new' template)" do
         post <%= index_helper %>_url, params: { <%= ns_file_name %>: invalid_attributes }
         expect(response).to be_successful
       end
+    <% else %>
+      it "renders a response with 422 status - unporcessable entity" do
+        post <%= index_helper %>_url, params: { <%= ns_file_name %>: invalid_attributes }
+        expect(response.status).to eq(422)
+      end
+    <% end %>
     end
   end
 


example_app_generator/generate_stuff.rb
```diff
@@ -167,4 +167,17 @@ def using_source_path(path)
...
+          'skip("Add a hash of attributes invalid for your model")',
+          '{ name: "" }'
+
+gsub_file 'spec/requests/widgets_spec.rb',
+          'skip("Add a hash of attributes valid for your model")',
+          '{ name: "lorem" }'
+
+gsub_file 'app/models/widget.rb',
+          'end',
+          'validates :name, presence: true; end'
```
panSarin, line 182: @pirj should I do it this way, or maybe I can just add validation to the widget model in the scaffold generator?

panSarin:
@pirj  I started that branch from rails-6-1-dev and since them regardless of CI or what RAILS_VERSION i will set , each time generator is generating rails scaffold for newest version.

```
def create
    @widget = Widget.new(widget_params)

    respond_to do |format|
      if @widget.save
        format.html { redirect_to @widget, notice: "Widget was successfully created." }
        format.json { render :show, status: :created, location: @widget }
      else
        format.html { render :new, status: :unprocessable_entity }
        format.json { render json: @widget.errors, status: :unprocessable_entity }
      end
    end
  end
  ```
  
  That code should only be generated in rails-6.1.1 but it is generated for each CI checks ( same locally ). Any ideas?

File: example_app_generator/generate_stuff.rb (hunk 1 above)
Comment (pirj) [reply example_app_generator/generate_stuff.rb:182]: It's quite good like this 👍

File: lib/generators/rspec/scaffold/templates/request_spec.rb [hunk 2]
```diff
@@ -82,10 +82,17 @@
         }.to change(<%= class_name %>, :count).by(0)
       end
 
+    <% if Rails.version.to_f < 6.1 || Rails.version == '6.1.0' %>
       it "renders a successful response (i.e. to display the 'new' template)" do
         post <%= index_helper %>_url, params: { <%= ns_file_name %>: invalid_attributes }
         expect(response).to be_successful
```
pirj, line 88: We're getting there.
```
  1) /widgets POST /create with invalid parameters renders a successful response (i.e. to display the 'new' template)
     Failure/Error: expect(response).to be_successful
       expected `#<ActionDispatch::TestResponse:0x0000564d388be7d8 @mon_mutex=#<Thread::Mutex:0x0000564d388be788>, @mo..., @method=nil, @request_method=nil, @remote_ip=nil, @original_fullpath=nil, @fullpath=nil, @ip=nil>>.successful?` to be truthy, got false
     # ./spec/requests/widgets_spec.rb:82:in `block (4 levels) in <top (required)>'

  2) /widgets PATCH /update with invalid parameters renders a successful response (i.e. to display the 'edit' template)
     Failure/Error: expect(response).to be_successful
       expected `#<ActionDispatch::TestResponse:0x0000564d38c3deb8 @mon_mutex=#<Thread::Mutex:0x0000564d38c3de68>, @mo..., @method=nil, @request_method=nil, @remote_ip=nil, @original_fullpath=nil, @fullpath=nil, @ip=nil>>.successful?` to be truthy, got false
     # ./spec/requests/widgets_spec.rb:114:in `block (4 levels) in <top (required)>'
```

I guess this should be
```
expect(response.status).to redirect_to(:new)
```
or something?

lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -82,10 +82,17 @@
...
       end
+    <% else %>
+      it "renders a response with 422 status - unporcessable entity" do
+        post <%= index_helper %>_url, params: { <%= ns_file_name %>: invalid_attributes }
+        expect(response.status).to eq(422)
+      end
+    <% end %>
```
pirj, line 95: That's quite odd for a 6.1 build 🤔 

```
   3) /widgets POST /create with invalid parameters renders a response with 422 status - unporcessable entity
     Failure/Error: expect(response.status).to eq(422)

       expected: 422
            got: 302
```

pirj:
What if you
```
export RAILS_VERSION='~> 6.0.0'
```
then remove your `Gemfile.lock` and `bundle install`?

I suggest adding to `Gemfile-custom`:
```ruby
gem 'pry-byebug'
```

Then, you would be able to add `binding.pry` in `lib/generators/rspec/scaffold/templates/request_spec.rb`:
```
<% binding.pry %>
<% if Rails.version.to_f < 6.1 || Rails.version == '6.1.0' %>
```
and see what `Rails.version` returns and if this condition is correct.

panSarin:
Hmm problem with rails version is not in ```lib/generators/rspec/scaffold/templates/request_spec.rb```.
There is rails version taken from RAILS_VERSION (checked it your way, but also errors from previous comments that you mention are specifying that).

 Problem is in generated 
```tmp/example_app/app/controllers/widgets_controller.rb```
and
```tmp/example_app/app/controllers/tickets_controller.rb```
that generates #create method with the 6.1.1 style ( even when rails --version => 6.0.3.5) which is:

```
def create    
    respond_to do |format|
      if @ticket.save
        format.html { redirect_to @ticket, notice: "Ticket was successfully created." }
        format.json { render :show, status: :created, location: @ticket }
      else
        format.html { render :new, status: :unprocessable_entity }
        format.json { render json: @ticket.errors, status: :unprocessable_entity }
      end
    end
end
```

and for rails versions lower than 6.1.1 it should be 

```
  # POST /tickets
  def create
    @ticket = Ticket.new(ticket_params)
    if @ticket.save
      redirect_to @ticket, notice: 'Ticket was successfully created.'
    else
      render :new
    end
  end
```

So I assume there is some issue in changing rails versions in example app generator from newest to the one selected in Gemfile.lock for rspec-rails. Or maybe I am doing something wrong ?

**Comment by pirj** [comment]
So what `Rails.version` evaluates to when you're at that breakpoint?

**Comment by panSarin** [comment]
> So what `Rails.version` evaluates to when you're at that breakpoint?

```ruby
   120:     context "with invalid parameters" do
   121:     <% byebug %>
=> 122:     <% if Rails.version.to_f < 6.1 || Rails.version == '6.1.0' %>
   123:       it "renders a successful response (i.e. to display the 'edit' template)" do
   124:         <%= file_name %> = <%= class_name %>.create! valid_attributes
   125:         patch <%= show_helper.tr('@', '') %>, params: { <%= singular_table_name %>: invalid_attributes }
   126:         expect(response).to be_successful
(byebug) Rails.version
"6.0.3.5"
(byebug) RUBY_VERSION
"2.7.2"
```

Funny thing is that i generated new applications for Rails 6.0.3.5 on two other machines, and in one of them there was "new 422" behaviou in generated controller, and in other one "the old one". But there was differences of ruby versions between them. So I will test that on same ruby, and probably report it as Rails issue.

PR #2467, opened 2021-02-20
Prevent collisions on let(:name) and let(:method_name)

JonRowe:
This is #2461 for main

### DIFF
--- a/snippets/avoid_fixture_name_collision.rb
+++ b/snippets/avoid_fixture_name_collision.rb
@@ -1,0 +1,57 @@
+if __FILE__ =~ /^snippets/
+  fail "Snippets are supposed to be run from their own directory to avoid side " \
+       "effects as e.g. the root `Gemfile`, or `spec/spec_helpers.rb` to be " \
+       "loaded by the root `.rspec`."
+end
+
+# We opt-out from using RubyGems, but `bundler/inline` requires it
+require 'rubygems'
+
+require "bundler/inline"
+
+# We pass `false` to `gemfile` to skip the installation of gems,
+# because it may install versions that would conflict with versions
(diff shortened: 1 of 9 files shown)


snippets/avoid_fixture_name_collision.rb
```diff
@@ -0,0 +1,57 @@
+if __FILE__ =~ /^snippets/
```
pirj, line 1: Just a note that we have no support for snippets in `main`, the code to run them is in `rails-6-1-dev` only.
But if CI passes on `rails-6-1-dev`, the patch should most possibly work for `main`, too.

pirj (approved):
Thanks!

benoittgt (approved):

snippets/avoid_fixture_name_collision.rb
JonRowe: Is it easy-ish to backport that support?

pirj: Should be easy. Looks like cherry-picking https://github.com/rspec/rspec-rails/commit/040fecc98894de42a433d94a15d19f07434b77f8 and accepting the removed lines from `Gemfile` should do it.

pirj:
Cherry-picked along with Changelog changes. Let's see if snippets are being run.

pirj:
Nice
```
============= Starting snippets ===============
Running avoid_fixture_name_collision.rb
.

Finished in 0.00988 seconds (files took 0.4626 seconds to load)
1 example, 0 failures

Running use_active_record_false.rb
.

Finished in 0.00749 seconds (files took 0.49572 seconds to load)
1 example, 0 failures

============= Ending snippets ===============
```

PR #2468, opened 2021-02-20
Update ammeter (to a temporary repo) to support RSpec 4

pirj:
`ammeter` is `rspec-rails`'s development dependency, used to test generators.
Its runtime contains code that is not compatible with the latest changes in RSpec 4 (unification of metadata multi-condition filtering), making `rspec-core`'s `rspec-rails` sub-build to fail.

```ruby
    c.include Ammeter::RSpec::Rails::GeneratorExampleGroup,
      :type          => :generator,
      :file_path     => lambda { |file_path, metadata|
        metadata[:type].nil? && generator_path_regex =~ file_path
      }
```

This is a temporary measure to use `ammeter`'s patched fork until the patch is accepted and a newer version is released.

See
- https://github.com/rspec/rspec-core/issues/1821
- https://github.com/rspec/rspec-core/pull/2874
- https://github.com/alexrothenberg/ammeter/pull/64
- example failure (Ammeter::RSpec::Rails::GeneratorExampleGroup that defines destination is not included) https://github.com/rspec/rspec-core/pull/2874/checks?check_run_id=1942170588

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,9 @@ end
 # Until 1.13.2 is released due to Rubygems usage
 gem 'ffi', '~> 1.12.0'
 
+# Until https://github.com/alexrothenberg/ammeter/pull/64 is merged and released
+gem 'ammeter', github: 'pirj/ammeter', branch: 'support-rspec-4'
+
 custom_gemfile = File.expand_path('Gemfile-custom', __dir__)
 eval_gemfile custom_gemfile if File.exist?(custom_gemfile)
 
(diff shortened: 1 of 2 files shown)


pirj:
Alternatively, we could just add the following to our `spec/spec_helper.rb` until `ammeter` is updated:
```ruby
    c.include Ammeter::RSpec::Rails::GeneratorExampleGroup,
      :type          => :generator
    c.include Ammeter::RSpec::Rails::GeneratorExampleGroup,
      :file_path     => lambda { |file_path, metadata|
        metadata[:type].nil? && generator_path_regex =~ file_path
      }
```

pirj:
I like the other approach better. And my suspicion is that we'll have to live longer with the patch, as `ammeter` doesn't seem to be actively maintained. Depending on my fork of it in the long run would be just weird.

PR #2469, opened 2021-02-20
Work around ammeter' incompatibility with RSpec 4

pirj:
See:

- rspec/rspec-core#1821
- rspec/rspec-core#2874
- alexrothenberg/ammeter#64
- example failure (Ammeter::RSpec::Rails::GeneratorExampleGroup that defines destination is not included) https://github.com/rspec/rspec-core/pull/2874/checks?check_run_id=1942170588

Less impactful alternative to #2468

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -74,4 +74,8 @@ RSpec.configure do |config|
       example.run
     end
   end
+
+  # TODO: Remove once https://github.com/alexrothenberg/ammeter/pull/64 is merged
+  # Work around ammeter's incompatibility with RSpec 4
+  config.include Ammeter::RSpec::Rails::GeneratorExampleGroup, type: :generator
 end


spec/spec_helper.rb
```diff
@@ -74,4 +74,14 @@ def self.run_all(reporter = nil)
...
+
+  # Work around ammeter's incompatibility with RSpec 4
+  # TODO: Remove once https://github.com/alexrothenberg/ammeter/pull/64 is merged
+  generator_path_regex = Regexp.compile(%w[spec generators].join('[\\\/]') + '[\\\/]')
+  config.include Ammeter::RSpec::Rails::GeneratorExampleGroup,
+                 type: :generator
+  config.include Ammeter::RSpec::Rails::GeneratorExampleGroup,
+                 file_path: lambda { |file_path, metadata|
+                   metadata[:type].nil? && generator_path_regex =~ file_path
+                 }
```
JonRowe, line 86: Do we even need this? The issue in ammeter is the use of both tags and file, we presumedly have one of them if we're getting failures?

pirj: Removed, works :+1: 
I'll squash the last commit.

JonRowe (approved):

PR #2470, opened 2021-02-24
Update ammeter dependency to support RSpec 4

pirj:
`ammeter` has been updated to support RSpec 4's changes related to metadata filtering, that allows us to remove the hack and use its mainline version.

See https://github.com/alexrothenberg/ammeter/pull/64#issuecomment-785280792


benoittgt (approved):
Good news!

PR #2473, opened 2021-03-02
Prevent using 3.9 with Rails 6.1

guigs:
There's an issue with Rails 6.1 that caused error `WrongScopeError`, which was only fixed in rspec-rails 4.0 by #2215.

This should warn the user that rspec-rails need to be upgraded.

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
     s.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]
   end
 
-  version_string = ['>= 3.0']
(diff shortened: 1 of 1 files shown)


JonRowe:
Sorry thats not possible, bundler will just go further down the dependency chain.

guigs:
What do you mean?

JonRowe:
We can't alter existing released versions of gems, and when you have some versions of a gem, say `3.9.1` and `3.9.2` and only `3.9.2` restricts Rails to `< 6.1` from bundlers point of view its ok to install 3.9.1 and Rails 6.1

So it is not possible to fix, you'll just get older versions of gems that don't have the restriction.

guigs:
Aha, I see. Makes totally sense. Thanks for the explanation!

PR #2474, opened 2021-03-03
Be conservative on which Rails version is supported

guigs:
`rspec-rails` 3.9 can be installed with Rails 6.1, because it does not limit which rails version is supported, but it actually does not support Rails 6.1. It is not possible to add restriction on the next point release (let's say 3.9.2), because bundler will then install previous version (3.9.1) that does not have the dependency restriction.

Moving forward with rspec-rails 4.1, I think we should be more conservative on which Rails version is supported. If it happens that the next rails version (6.2 or 7.0) is still compatible, it will be easy to make a point release of rspec-rails 4.1.x with the upper bound dependency on Rails updated.

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |s|
     s.cert_chain = [File.expand_path('~/.gem/rspec-gem-public_cert.pem')]
   end
 
-  version_string = ['>= 4.2']
+  version_string = ['>= 4.2', '< 6.2']
 
   s.add_runtime_dependency 'actionpack',    version_string
   s.add_runtime_dependency 'activesupport', version_string


JonRowe:
The same problem applies here, as in #2473, any released version of a gem that doesn't have this constraint can be installed by bundler, a new major version could have this constraint added but there would still be no indication for the previous version that anything is wrong.

guigs:
Yes, true, but if you have `gem 'rspec-rails', '~> 4.1'` in your Gemfile, it will work. I think at some point in future this will beneficial, so the sooner we add the better.

JonRowe:
I'm sorry I don't want to merge this, if the top version is constrained what happens is Bundler picks an older even less compatible version. I would rather that the most recent version is installed in the case of incompatibility, so we get bug reports and can fix and release versions.

PR #2475, opened 2021-03-06
remove all spec views on destroy controller

Naokimi:
**Actions**:
1. `rails g controller Authors index show`
2. `rails d controller Authors` 

**Current**
_ON GENERATE_
Rails generates controller and view files
Rspec generates controller and view files
_ON DESTROY_
Rails destroys controller and view files
Rspec destroys only controller

**Proposal**
_ON DESTROY_
Rails destroys controller and view files
Rspec destroys controller and view files

### DIFF
--- a/lib/generators/rspec/controller/controller_generator.rb
+++ b/lib/generators/rspec/controller/controller_generator.rb
@@ -27,7 +27,7 @@ module Rspec
       end
 
       def generate_view_specs
-        return if actions.empty?
+        return if actions.empty? && behavior == :invoke
         return unless options[:view_specs] && options[:template_engine]
 
         empty_directory File.join("spec", "views", file_path)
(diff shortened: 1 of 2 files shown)


pirj:
Awesome! Can you please add specs for this change?

Naokimi:
`run_generator %w[posts], behavior: :revoke` isn't working and I can't figure out why
they use it for rails: https://github.com/rails/rails/blob/main/railties/test/generators/task_generator_test.rb#L23

Naokimi:
@pirj found a workaround, ready to be reviewed again

benoittgt:
Hello @Naokimi  and thanks for you PR.

Tested locally and it is looks good to me.

Capture d’écran 2021-03-08 à 21 13 43

benoittgt (approved):

JonRowe:
Do we have this problem with other generators? Do we want to ensure we're consistent?

Naokimi:
seems like other generators work as intended
just controller one had a guard clause that would block expected behaviour

PR #2479, opened 2021-03-10
Fix support for Rails 4.2

### LINKED ISSUE #2477: RSpec 4.1 breaks on Rails 4.2

Ruby version: 2.4.10
Rails version: 4.2.11.3
RSpec version: 4.1.0

## Observed behaviour

RSpec 4.1 does not work start on Rails 4.2.11.3, while 4.0 does

```
     Failure/Error: example.run
     
     NameError:
       undefined local variable or method `use_transactional_tests' for #<RSpec::ExampleGroups::Welcome::Index:0x00007fcb64cdbf40>
       Did you mean?  use_transactional_fixtures
```

Tracked down with `git bisect` down to 073daad07027d7fb53a3f7ac3b06a0dd4af498da

## Expected behaviour

Keeps working with Rails 4.1 or `gemspec` specifies `>= 5.0` to prevent update to 4.1

https://github.com/rspec/rspec-rails/blob/main/rspec-rails.gemspec#L36

Travis CI failure: https://travis-ci.org/github/diowa/ruby2-rails4-bootstrap-heroku/builds/761888443

tagliala:
Rails 4.2 does not have `use_transactional_tests` defined

Fix #2477

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,3 +1,10 @@
+### 4.1.1 / 2021-03-09
+
+Bug Fixes:
+
+* Remove generated specs when destroying a generated controller.
+  (@Naokimi, #2475)
+
 ### 4.1.0 / 2021-03-06
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v4.0.2...v4.1.0)
 
--- a/lib/rspec/rails/version.rb
+++ b/lib/rspec/rails/version.rb
@@ -3,7 +3,7 @@ module RSpec
     # Version information for RSpec Rails.
     module Version
       # Current version of RSpec Rails, in semantic versioning format.
-      STRING = '4.1.0'
+      STRING = '4.1.1'
     end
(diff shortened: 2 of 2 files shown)


lib/rspec/rails/fixture_support.rb
```diff
@@ -13,7 +13,11 @@ module FixtureSupport
         # Monkey patched to avoid collisions with 'let(:name)' in Rails 6.1 and after
         # and let(:method_name) before Rails 6.1.
         def run_in_transaction?
-          use_transactional_tests && !self.class.uses_transaction?(self)
+          if ::Rails::VERSION::STRING > '5'
```
JonRowe, line 16: Generally, we define conditional methods rather than use run time checks, and this should probably check for the existence of `use_transactional_tests` directly rather than a version, although if you leave it on a version it should be `>= 5`

PR #2484, opened 2021-03-12
Fixed content_type comparison from eq to match

ham0215:
Since Rails 6.1, the return value of content-type has changed as follows, so the scaffold template has been modified to support Rails6.1 and others.

https://github.com/rails/rails/pull/36034/files#diff-f5ff9aa07f44111a79d56c09ec37d774b462d97aff68f32490c2e56e74c95783R137

### DIFF
--- a/lib/generators/rspec/scaffold/templates/api_request_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/api_request_spec.rb
@@ -80,7 +80,7 @@ RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %
         post <%= index_helper %>_url,
              params: { <%= ns_file_name %>: invalid_attributes }, headers: valid_headers, as: :json
         expect(response).to have_http_status(:unprocessable_entity)
-        expect(response.content_type).to eq("application/json")
+        expect(response.content_type).to match(a_string_including("application/json"))
       end
     end
   end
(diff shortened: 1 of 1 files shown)


pirj (approved):
Thank you!

I guess this is unrelated to Rails' change since they claim that previously:
```
resp.content_type #=> "text/csv; header=present"
```

And we're actually already [using `match(a_string_including` in a couple of places in the same file](https://github.com/rspec/rspec-rails/blob/cfc7b83188dbaed2c85bb30bd6d7b3ad829abbfb/lib/generators/rspec/scaffold/templates/api_request_spec.rb#L67).

PS

> now returned

🤯

JonRowe:
These are generator templates, so I'm happy for you to merge these as is if you want @pirj but I am curious why our tests didn't pick this up?

pirj:
@JonRowe This is due to the generated spec containing this:
```ruby
RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %> do
  let(:valid_attributes) {
    skip("Add a hash of attributes valid for your model")
  }
```
So we're skipping them when running `example_app_generator`'s app. It's something that we could improve.

pirj:
Thank you, @ham0215 !

JonRowe:
@pirj Ah a change broke the example app generator then, as we explictly gsub that to a hash... we need to fix that here, and then ensure it can't break like this again... https://github.com/rspec/rspec-rails/blob/main/example_app_generator/generate_stuff.rb#L165

PR #2490, opened 2021-04-01
Fix "render" typo on request specs generated via scaffold

lmatiolis:
As seen on other specs in the same file, the spec should read as:

`it "renders a successful response" do`

instead of:

`it "render a successful response" do`


pirj (approved):
Thanks!

PR #2495, opened 2021-04-19
Fix FixtureSupport#run_in_transaction?

st0012:
`ActiveRecord::TestFixture`'s `uses_transaction` is designed to be used like this:

```ruby
uses_transaction :the_test_method_name
```

And in RSpec, the method name would be the example's name.

```ruby
uses_transaction "does someting"

it "does someting" {}
```

But in the current implementation, it's passing the example object instead of its name, which would always fail the name comparison in https://github.com/rails/rails/blob/main/activerecord/lib/active_record/test_fixtures.rb#L94-L97

So this PR fixes the issue by passing the example's name instead of the example object.

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -13,7 +13,8 @@ module RSpec
         # Monkey patched to avoid collisions with 'let(:name)' in Rails 6.1 and after
         # and let(:method_name) before Rails 6.1.
         def run_in_transaction?
-          use_transactional_tests && !self.class.uses_transaction?(self)
+          current_example_name = (RSpec.current_example && RSpec.current_example.metadata[:description])
+          use_transactional_tests && !self.class.uses_transaction?(current_example_name)
         end
 
         included do
--- a/spec/rspec/rails/fixture_support_spec.rb
+++ b/spec/rspec/rails/fixture_support_spec.rb
@@ -12,6 +12,33 @@ module RSpec::Rails
       end
     end
 
+    context "with use_transactional_tests set to true" do
+      it "works with #uses_transaction helper" do
+        group = RSpec::Core::ExampleGroup.describe do
+          include FixtureSupport
+          self.use_transactional_tests = true
+
+          uses_transaction "doesn't run in transaction"
+
+          it "doesn't run in transaction" do
+            expect(ActiveRecord::Base.connection.transaction_open?).to eq(false)
+          end
+
+          it "runs in transaction" do
+            expect(ActiveRecord::Base.connection.transaction_open?).to eq(true)
+          end
+        end
+
+        expect_to_pass(group)
+      end
+
+      def expect_to_pass(group)
+        result = group.run(failure_reporter)
+        failure_reporter.exceptions.map { |e| raise e }
+        expect(result).to be true
+      end
+    end
+
     it "will allow #setup_fixture to run successfully", skip: Rails.version.to_f <= 6.0 do
       group = RSpec::Core::ExampleGroup.describe do
         include FixtureSupport
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,8 @@ Enhancements:
 
 * Make the API request scaffold template more consistent and compatible with
   Rails 6.1. (Naoto Hamada, #2484)
+* Provide support to turn off transactional tests for select examples.
+  (Stan Lo, #2495)
 
 ### 5.0.1 / 2021-03-18
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v5.0.0...v5.0.1)


lib/rspec/rails/fixture_support.rb
```diff
@@ -13,7 +13,8 @@ module FixtureSupport
         # Monkey patched to avoid collisions with 'let(:name)' in Rails 6.1 and after
         # and let(:method_name) before Rails 6.1.
         def run_in_transaction?
-          use_transactional_tests && !self.class.uses_transaction?(self)
+          current_example_name = (RSpec.current_example && RSpec.current_example.metadata[:description])
```
st0012, line 16: both `@example` and `method_name` works in my Rails project but not here. right now RSpec meta is the most reliable way for me to access the example's name. I know it's ugly so it'd be great if anyone can give me a pointer on this 🙂

pirj: Wondering what is `self` in this context? Is `self.metadata` or just `metadata` available?

pirj (commented):
Awesome, thank you!

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -12,6 +12,29 @@ module RSpec::Rails
       end
     end
 
+    context "with use_transactional_tests set to true" do
+      it "works with #uses_transaction helper" do
+        group = RSpec::Core::ExampleGroup.describe do
+          include FixtureSupport
+          self.use_transactional_tests = true
+
+          uses_transaction "doesn't run in transaction"
```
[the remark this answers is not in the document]
st0012: It's not documented. But yes, based on the code and the current behavior it turns that off for the given tests.

lib/rspec/rails/fixture_support.rb
st0012: calling either one returns 

```
`metadata` is not available from within an example (e.g. an `it` block) or from constructs that 
run in the scope of an example (e.g. `before`, `let`, etc). It is only available on an example group 
(e.g. a `describe` or `context` block).`
```

I do wish methods like `name` or `metadata` is available in an example though.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -12,6 +12,29 @@ module RSpec::Rails
...
+
+        expect_to_pass(group)
+      end
+
+      def expect_to_pass(group)
+        result = group.run(failure_reporter)
+        failure_reporter.exceptions.map { |e| raise e }
+        expect(result).to be true
+      end
+    end
```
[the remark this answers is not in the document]
st0012: added

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -12,6 +12,33 @@ module RSpec::Rails
...
+
+          uses_transaction "doesn't run in transaction"
+
+          it "doesn't run in transaction" do
+            expect(run_in_transaction?).to eq(false)
+          end
+
+          it "runs in transaction" do
+            expect(run_in_transaction?).to eq(true)
+          end
```
pirj, line 29: What do you think of changing those expectations to:
```ruby
expect(ActiveRecord::Base.connection.transaction_open?).to eq(...)
```
Worked for me.
It feels a bit more reassuring.

lib/rspec/rails/fixture_support.rb
pirj: Right.
`@example` seems to be coming from https://github.com/rspec/rspec-core/blob/ea8554afd1a2b63677c6593059fa8f2476181deb/lib/rspec/core/example.rb#L331. It feels that it's an undesirable effect that it is accessible in this scope.

I don't think there's anything wrong with using `current_example` here, even if it's not too elegant.

pirj (approved):
All looks good.
Thanks a lot for this contribution.

spec/rspec/rails/fixture_support_spec.rb
st0012: Good point 👍 have updated

st0012: it looks like the change caused some builds to fail but I can't see the log. can you gather some error messages here?

pirj:
```
Ruby: 2.5.8, Rails: ~> 5.2.0
GitHub Actions has encountered an internal error when running your job.
```
:man_shrugging:  

Re-running.

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -12,6 +12,33 @@ module RSpec::Rails
...
+    context "with use_transactional_tests set to true" do
+      it "works with #uses_transaction helper" do
+        group = RSpec::Core::ExampleGroup.describe do
+          include FixtureSupport
+          self.use_transactional_tests = true
+
+          uses_transaction "doesn't run in transaction"
+
+          it "doesn't run in transaction" do
+            expect(ActiveRecord::Base.connection.transaction_open?).to eq(false)
```
JonRowe, line 24: I'm having a big WTF moment at this, so `uses_transaction` turns off transactional tests?

spec/rspec/rails/fixture_support_spec.rb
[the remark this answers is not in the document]
JonRowe: If its not documented I'm hesitant to support it, if its public api fine but private api I'm a bit reluctant about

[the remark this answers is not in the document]
pirj: Public https://api.rubyonrails.org/classes/ActiveRecord/TestFixtures/ClassMethods.html#method-i-uses_transaction, but undocumented. Originally introduced [here](https://github.com/rails/rails/commit/2c0fa3208859aed2c1fd0ce54c8d5215d11d0c75).

Frankly, a couple of times I needed it desperately to test code that needs to have control of the outermost transaction or has a conditional statement with a check if the transaction is an outermost or a nested.

With DatabaseCleaner it's possible to turn off per-example transaction for select examples. With `use_transactional_fixtures` it's not, and it's a major inconvenience.
I was under the spell that it was, and it took me a while to realize that it didn't work like that on my last project. Our documentation lacks this moment.

JonRowe:
@pirj I'll leave any decisions on this up to you, my feedback is non blocking

spec/rspec/rails/fixture_support_spec.rb
pirj: Yes. No doubt `uses_transaction` is an ambiguous and confusing naming.

pirj:
Thank you for the contribution and for revealing the mystery behind `uses_transaction`, @st0012 !

st0012:
@pirj I've been using rspec-rails for years and it's finally my turn to make a small contribution 🙂

PR #2496, opened 2021-04-20
Implement check to prevent duplicate job naming

Liberatys:
# Description

When generating a new job with:
		    
	    rails g job mailer_job
Rails will not append another _job to the Job-Name or the file. 

This pr tries to align the creation of the spec file with the creation of the job file and checks for the file suffix before appending another _job.

# Example

## Setup

		rails g job sms_sending_job

## Before

		spec/jobs/sms_sending_job_job.rb

## After

		spec/jobs/sms_sending_job.rb


JonRowe (approved):

JonRowe:
Thanks!

PR #2505, opened 2021-06-04
Removed unnecessary whitespace from request spec scaffold template.

Madogiwa0124:
I removed unnecessary whitespace from the scaffold template in the request spec to match the others.

How about this PR?


pirj:
Thank you!

PR #2508, opened 2021-06-21
Respect RSpec `default_path` for generators

### LINKED ISSUE #2507: rails generators not respecting RSpec's `default_path`

Ruby version: 3.0
Rails version: 6.1.3.2
RSpec version: main

## Other RSpec configuration

RSpec is configured to use `behaviour` directory i.e.

```ruby
RSpec.configure do |c|
  c.default_path = "behavior"
end
```

## Observed behaviour

When running rspec generator by default `spec` directory got chosen

```
$ rails generate rspec:model user
    ...
    create  spec/models/user_spec.rb
    ...
```

## Expected behaviour

Generator should place file inside `behaviour` directory instead.

```
$ rails generate rspec:model user
    ...
    create  behaviour/models/user_spec.rb
    ...
```

I don't have any at the moment.

vivekmiyani:
Fixes #2507.

## TODO

- [x]  Replace hardcoded `spec` directory to `default_path`
- [x] Change log entry
- [ ] Add feature test for each generator listed below:
  - [x]  channel
  - [x] controller
  - [x] feature
  - [x] generator
  - [x] helper
  - [ ] install
  - [x] integration
  - [x] job
  - [x] mailbox
  - [x] mailer
  - [ ] model
  - [x] request
  - [ ] scaffold
  - [x] system
  - [x] view

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -73,6 +73,8 @@ Bug Fixes:
   `run_in_transaction?` method.  (Stan Lo, #2495)
 * Prevent keyword arguments being lost when methods are invoked dynamically
   in controller specs. (Josh Cheek, #2509, #2514)
+* Read configuration options from files to respect spec path setting.
+  (@vivekmiyani, #2508)
 
 ### 5.0.1 / 2021-03-18
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v5.0.0...v5.0.1)
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,10 @@
 ### Development
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v6.0.0.rc1...main)
 
+Enhancements:
+
+* Generators now respects default path configuration option. (@vivekmiyani, #2508)
+
 Breaking Changes:
 
 * Change the order of `after_teardown` from `after` to `around` in system
@@ -73,8 +77,6 @@ Bug Fixes:
   `run_in_transaction?` method.  (Stan Lo, #2495)
 * Prevent keyword arguments being lost when methods are invoked dynamically
   in controller specs. (Josh Cheek, #2509, #2514)
-* Read configuration options from files to respect spec path setting.
-  (@vivekmiyani, #2508)
 
 ### 5.0.1 / 2021-03-18
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v5.0.0...v5.0.1)
--- a/features/generator_specs/generator_specs.feature
+++ b/features/generator_specs/generator_specs.feature
@@ -15,8 +15,11 @@ Feature: Generator spec
                 create    spec/generator/my_generators_generator_spec.rb
           """
 
-    Scenario: Use custom generator with different default path
-        When I run `bundle exec rails generate rspec:install --default-path behaviour --force`
+    Scenario: Use custom generator with customized `default-path`
+        Given a file named ".rspec" with:
+          """
+          --default-path behaviour
+          """
         And I run `bundle exec rails generate generator my_generator`
         Then the features should pass
         Then the output should contain:
--- a/features/generator_specs/generator_specs.feature
+++ b/features/generator_specs/generator_specs.feature
@@ -14,3 +14,17 @@ Feature: Generator spec
                 invoke  rspec
                 create    spec/generator/my_generators_generator_spec.rb
           """
+
+    Scenario: Use custom generator with different default path
+        When I run `bundle exec rails generate rspec:install --default-path behaviour --force`
+        And I run `bundle exec rails generate generator my_generator`
+        Then the features should pass
+        Then the output should contain:
+          """
+                create  lib/generators/my_generator
+                create  lib/generators/my_generator/my_generator_generator.rb
+                create  lib/generators/my_generator/USAGE
+                create  lib/generators/my_generator/templates
+                invoke  rspec
+                create    behaviour/generator/my_generators_generator_spec.rb
+          """
--- a/lib/generators/rspec/install/install_generator.rb
+++ b/lib/generators/rspec/install/install_generator.rb
@@ -44,9 +44,11 @@ DESC
         replace_generator_command(spec_helper_path)
         remove_warnings_configuration(spec_helper_path)
 
-        dot_rspec_path = File.join(tmpdir, '.rspec')
+        unless default_path == "spec"
+          dot_rspec_path = File.join(tmpdir, '.rspec')
 
-        append_default_path(dot_rspec_path)
+          append_default_path(dot_rspec_path)
+        end
       end
 
       def replace_generator_command(spec_helper_path)
--- a/lib/generators/rspec/install/install_generator.rb
+++ b/lib/generators/rspec/install/install_generator.rb
@@ -12,6 +12,8 @@ Description:
     Copy rspec files to your application.
 DESC
 
+      class_option :default_path, type: :string, default: 'spec'
+
       def self.source_root
         @source_root ||= File.expand_path(File.join(File.dirname(__FILE__), 'templates'))
       end
@@ -20,12 +22,12 @@ DESC
         Dir.mktmpdir do |dir|
           generate_rspec_init dir
           template File.join(dir, '.rspec'), '.rspec'
-          directory File.join(dir, 'spec'), 'spec'
+          directory File.join(dir, 'spec'), default_path
         end
       end
 
       def copy_rails_files
-        template 'spec/rails_helper.rb'
+        template 'spec/rails_helper.rb', "#{default_path}/rails_helper.rb"
       end
 
     private
@@ -41,6 +43,10 @@ DESC
 
         replace_generator_command(spec_helper_path)
         remove_warnings_configuration(spec_helper_path)
+
+        dot_rspec_path = File.join(tmpdir, '.rspec')
+
+        append_default_path(dot_rspec_path)
       end
 
       def replace_generator_command(spec_helper_path)
@@ -57,6 +63,15 @@ DESC
                   /#{empty_line}(#{comment_line})+\s+config\.warnings = true\n/,
                   '',
                   verbose: false
+      end
+
+      def append_default_path(dot_rspec_path)
+        append_to_file dot_rspec_path,
+                       "--default-path #{default_path}"
+      end
+
+      def default_path
+        options[:default_path]
       end
     end
   end
--- a/features/README.md
+++ b/features/README.md
@@ -34,10 +34,14 @@ without having to type RAILS_ENV=test.
 
 Now you can run:
 
-    script/rails generate rspec:install
+    bundle exec rails generate rspec:install
 
 This adds the spec directory and some skeleton files, including a .rspec
 file.
+
+You can also customize the default spec path with `--default-path` option:
+
+    bundle exec rails generate rspec:install --default-path behaviour
 
 ## Issues
 
--- a/features/generator_specs/channel_specs.feature
+++ b/features/generator_specs/channel_specs.feature
@@ -1,0 +1,31 @@
+Feature: Channel generator spec
+
+    Scenario: Channel generator
+        When I run `bundle exec rails generate channel group`
+        Then the features should pass
+        Then the output should contain:
+          """
+                invoke  rspec
+                create    spec/channels/group_channel_spec.rb
+                create  app/channels/group_channel.rb
+             identical  app/javascript/channels/index.js
+             identical  app/javascript/channels/consumer.js
+                create  app/javascript/channels/group_channel.js
+          """
+
+    Scenario: Channel generator with customized `default-path`
+        Given a file named ".rspec" with:
+          """
+          --default-path behaviour
+          """
+        And I run `bundle exec rails generate channel group`
+        Then the features should pass
+        Then the output should contain:
+          """
+                invoke  rspec
+                create    behaviour/channels/group_channel_spec.rb
+                create  app/channels/group_channel.rb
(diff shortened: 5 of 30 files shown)


pirj:
Something failed

vivekmiyani:
Hii, I've updated the branch.

At the moment, rails installer does not support `--default_path`, So `lib/generators/rspec/install/install_generator.rb` remains untouched.
I think to change the `default_path` at the time of `rails generate rspec:install`, We need to do modify the `.rspec` file provided by `rspec-core`.

I'll work on it when get a chance.

pirj:
Looks great, thank you. You may (completely optionally) add a feature to `features/generator_specs/generator_specs.feature`.

Just noticed that our Generators documentation is [split into two](https://relishapp.com/rspec/rspec-rails/v/5-0/docs/generators), and the second one's name is misleading, [Generator specs](https://relishapp.com/rspec/rspec-rails/v/5-0/docs/generator-specs). It's a topic for another PR to bring them together, though.

vivekmiyani:
> You may (completely optionally) add a feature to features/generator_specs/generator_specs.feature.

@pirj Are we only going to test `features/generator_specs/generator_specs.feature` for different `default_path` and not others features. ?

Forgive me if I misunderstood anything 😅 .

pirj:
Yes, just your fix. Those features are published as documentation, and are end-to-end tests. No worries if you get stuck with it. Altering the default spec path is a pretty rare case. Also, with your fix it now works intuitively, so maybe documentation for it is redundant. Leaving it up to you to decide.

vivekmiyani:
Okay I will update the PR. But while implementing the feature I am running into one issue.:

I thought that `default_path` will be initialized from `.rspec` OR `spec_helper.rb` at the time of running `rails generate ...` but that is not the case with actual project.

It looks the blocker is the `ConfigurationOptions` which was **parsing options** when I ran `bundle exec rspec` but not for `rails generate ...`.

vivekmiyani:
We have to do something like, https://github.com/rspec/rspec-core/blob/main/lib/rspec/core/runner.rb#L132 (i.e. `RSpec::Core::Runner#configure`) is doing the same as we need here.

Any thoughts or suggestion?

pirj:
It's all kind of weird. If I'm not mistaken, the doc ‎https://relishapp.com/rspec/rspec-core/v/3-9/docs/configuration/setting-the-default-spec-path says that `default_path` is not supported on `RSpec.configuration`, but it seems that it is.The reasoning is not exactly clear.Let me check later when I get back to the computer.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        From: VivekSent: Saturday, 26 June 2021 10:31To: rspec/rspec-railsReply To: rspec/rspec-railsCc: Phil Pirozhkov; MentionSubject: Re: [rspec/rspec-rails] Respect RSpec `default_path` for generators (#2508)
We have to do something like, https://github.com/rspec/rspec-core/blob/main/lib/rspec/core/runner.rb#L132 (i.e. RSpec::Core::Runner#configure) is doing the same as we need here.
Any thoughts or suggestion?

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or unsubscribe.
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/rspec/rspec-rails/pull/2508#issuecomment-868962339",
"url": "https://github.com/rspec/rspec-rails/pull/2508#issuecomment-868962339",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

pirj:
> `--default-path` ... is not supported on RSpec.configuration, as it needs to be set before spec files are loaded

But https://github.com/rspec/rspec-core/blob/247d0a706ffa955718d600f3588aeb1e60cebde4/lib/rspec/core/configuration.rb#L35
```ruby
    #     RSpec.configure do |c|
    #       c.default_path = 'behavior'
```
and
```
      # Path to use if no path is provided to the `rspec` command (default:
      # `"spec"`). Allows you to just type `rspec` instead of `rspec spec` to
      # run all the examples in the `spec` directory.
```

and what's more important:
```ruby
      # @note Other scripts invoking `rspec` indirectly will ignore this
      #   setting.
```
which is exactly the case with generators.

Though `.rspec` should support setting the default path with:
```
--default-path behaviour
```

Can you please confirm that defining the default path in `.rspec` works with this fix for generators?
Wondering if it gets loaded when you require 'rspec/core` as you've added.
And why `spec/spec_helper.rb` is not loaded, even though a typical `.rspec` file defines `--require spec/spec_helper.rb`. 🤔 

If it works, I suggest you to write the feature test in such a way that the default path is set in the `.rspec` file, not `spec/spec_helper.rb`.

pirj:
I think I have bad news. Options from files, `.rspec` etc are loaded by [`ConfigurationOptions`](https://github.com/rspec/rspec-core/blob/247d0a706ffa955718d600f3588aeb1e60cebde4/lib/rspec/core/configuration_options.rb#L121) that is exclusively called by `Runner` that is responsible for running specs, and only [immediately before running specs](https://github.com/rspec/rspec-core/blob/247d0a706ffa955718d600f3588aeb1e60cebde4/lib/rspec/core/runner.rb#L66).
`spec/spec_helper.rb` is required after that if `.rspec` has a `--require spec_helper` option.
So `require 'rspec/core'` only loads up the default `Configuration` with [`default_path` set to `spec`](https://github.com/rspec/rspec-core/blob/247d0a706ffa955718d600f3588aeb1e60cebde4/lib/rspec/core/configuration.rb#L537). Alas.

vivekmiyani:
Yes that's what I exactly found. See my https://github.com/rspec/rspec-rails/pull/2508#issuecomment-868962339 I've mentioned the line no. where `ConfigurationOptions#configure` updates the `Configuration`.

I will try to find more and will update the PR.

JonRowe:
I'd like to see a feature file demonstrating this behaviour so that its both documented and tested, the changes to the generator look fine but I'm not sure its sufficient to load `rspec/core` before accessing configuration, the config must have been loaded / set before being accessed.

vivekmiyani:
@JonRowe @pirj  I will implement feature file.
but please check the ca907cf commit, I've added temporary workaround to parse the .rspec file (Extracted from the `rspec/core/runner.rb`).

Can we initialize `ConfigurationOptions` when `Configuration` initializes in `rspec-core`, So separate initialization of `ConfigurationOptions` can be removed.?

pirj:
I believe with those discoveries, adding a feature test becomes a necessity.

vivekmiyani:
Hey,
I have added rspec install option:

```sh
# Install rspec at custom directory (this will append `--default-path` option to `.rspec` file)
$ rails generate rspec:install --default-path behaviour
```

Also look at 5f7b640 which implements the feature you've asked for!!

vivekmiyani:
@pirj @JonRowe Do you want me to write the feature for all other generators (like: controller, model, etc.. etc..)?

I've only added feature for `generator` generator, as I mentioned above.

features/generator_specs/generator_specs.feature
```diff
@@ -14,3 +14,17 @@ Feature: Generator spec
                 invoke  rspec
                 create    spec/generator/my_generators_generator_spec.rb
           """
+
+    Scenario: Use custom generator with different default path
+        When I run `bundle exec rails generate rspec:install --default-path behaviour --force`
```
pirj, line 19: Features are used for two purposes, to generate the [documentation](https://relishapp.com/rspec/rspec-rails/docs) and for end-to-end testing. So we strive to keep the balance between the two. This scenario seems to be more end-to-end centred.
From the documentation standpoint, it's not clear where the default path setting is persisted.

As far as I see, you've improved the current code in two aspects:
1. It's now able to persist `--default-path` option passed to `rspec:install` into `.rspec`
2. Generators now pick up the default path option from `.rspec`

I suggest separating documentation for those two improvements.
Does it make sense from your standpoint?

If you'd agree, I suggest adding `rspec:install` example as a note to `features/README.md`'s "Configure" section, something like:
```
You can also customize the default spec path with `--default-path` option:

bundle exec rails generate rspec:install --default-path behaviour --force
```
I can't think of a better place this note would fit.

Side note - I really appreciate if you take care of `script/rails generate` along the way, I believe it's a legacy form, and doesn't even work anymore.

For `generator_specs.feature`, instead of running `rspec:install` in a project directory that already has RSpec installed, I'd suggest to directly specify the contents of `.rspec`. You can copy-paste this piece from [this feature](https://github.com/rspec/rspec-core/blob/9de11e52d078696c01c3515351e1e634aada1115/features/configuration/default_path.feature#L25).

And yes, one feature is sufficient. It is a generic document, that was supposed to be a root documentation file for all generators. The table of contents is just broken, and this doc appears separate from other generator docs.

vivekmiyani: Updated!

pirj (approved):
The code is copied over from `rspec-core`. It might be exposed as a public interface, but this may be postponed to a later version, and we can do this in `rspec-core` 4.0.1, and set the minimum dependent version to it.

Thank you!

vivekmiyani:
@pirj Can we initialize `ConfigurationOptions` in `Configuration` constructor directly in `rspec-core`, So separate initialization of `ConfigurationOptions` can be completely removed from codebase.?

pirj:
That's an option to consider, too. `RSpec.configuration` is memoized, though:
```ruby
  def self.configuration
    @configuration ||= RSpec::Core::Configuration.new
  end
```
and there could be cases when it's necessary to run a suite with different options, e.g. filters. It would make `ConfigurationOptions` class redundant, and some may make use of it. It could be done in a major `rspec-core` version, but we'd prefer not to do this in RSpec 4.

I don't feel that this code duplication is a big issue, it just hints that something can be optimized.

lib/generators/rspec/install/install_generator.rb
```diff
@@ -41,6 +43,10 @@ def generate_rspec_init(tmpdir)
 
         replace_generator_command(spec_helper_path)
         remove_warnings_configuration(spec_helper_path)
+
+        dot_rspec_path = File.join(tmpdir, '.rspec')
+
+        append_default_path(dot_rspec_path)
```
JonRowe, line 49: Its not necessary to always set this, perhaps:
```suggestion
        unless default_path == "spec"
          dot_rspec_path = File.join(tmpdir, '.rspec')

          append_default_path(dot_rspec_path)
        end
```

Although it would be nice if we could determine if it was set when generating.

Changelog.md
```diff
@@ -13,6 +13,8 @@ Bug Fixes:
 * Fix `ActiveRecord::TestFixture#uses_transaction` by using example description
   to replace example name rather than example in our monkey patched
   `run_in_transaction?` method.  (Stan Lo, #2495)
+* Read configuration options from files to respect spec path setting.
+  (@vivekmiyani, #2508)
```
JonRowe, line 18:
```suggestion
```

This isn't what this PR does, instead you need a section on the enhancements section saying that generators now support this flag.

JonRowe (changes requested):
Sorry for the delayed feedback, looking good but some more tweaks needed.

features/generator_specs/generator_specs.feature
```diff
@@ -14,3 +14,20 @@ Feature: Generator spec
...
+        Then the features should pass
+        Then the output should contain:
+          """
+                create  lib/generators/my_generator
+                create  lib/generators/my_generator/my_generator_generator.rb
+                create  lib/generators/my_generator/USAGE
+                create  lib/generators/my_generator/templates
+                invoke  rspec
+                create    behaviour/generator/my_generators_generator_spec.rb
+          """
```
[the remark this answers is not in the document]
vivekmiyani: Hey @JonRowe, We are only testing generator `generator` even for `spec` directory. So can we make this topic for some another PR?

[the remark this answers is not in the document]
pirj: Sounds reasonable to fix separately. I recall there were a couple of problems with the structure of those features breaking the table of contents etc.

JonRowe (changes requested):
This changes behaviour for all generators, yet is only tested against one generator. This needs the additional tests and scenarios.

vivekmiyani:
Hello guys, Sorry for delayed response. I was super busy in some stuffs.

As @JonRowe asked, I am writing feature tests for below items (List grabbed from `spec/generators/rspec/**/*_spec.rb`):

- channel
- controller
- feature
- ~generator~
- helper
- install
- integration
- job
- mailbox
- mailer
- model
- request
- scaffold
- system
- view

vivekmiyani:
@pirj @JonRowe Check channel generator feature, is it correct OR any modification needed?
If it is correct I'll repeat for others.

pirj:
> channel generator feature, is it correct

Looks good to me 👍 
Thanks for getting back to this.

vivekmiyani:
Hey @pirj @JonRowe, Added much more feature tests 🎉 .

Below are skipped as I couldn't figure how to tackle them 🤕 :
- **Install** generator feature, Because I' not able to run the installation using `When I run bundle exec rails generate rspec:install` in feature, as our directory has already installed rspec structure.
- **Model** generator feature, Because I don't how to match the migration timestamp output 🤔 .
- **Scaffold** generator feature, Same I don't how to match the migration timestamp output.

JonRowe:
@pirj can you review this?

vivekmiyani:
@JonRowe @pirj Can you please review this?

pirj:
Thank you!

PR #2509, opened 2021-07-15
Fix keyword arguments in dynamic matchers

JoshCheek:
Ruby 3 changed the way keyword arguments work. Currently, to forward arguments, the only way that works in Ruby 2.6, 2.7, and 3.0, is to call `ruby2_keywords :method_name` for the method that receives the arguments.

This will cause Ruby to remember that the final hash was created by keyword arguments, and then when expanding out the arguments later, eg [here](https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/built_in/has.rb#L67), it will see that the last element was created by keywords, and it will pass that into the keywords area of a method infocation.

If we don't do this, it will consider it an ordinal argument and not line it up with the keywords. This typically causes a "wrong number of arguments" error, because it's passing an extra ordinal argument.

Here is another example from within RSpec, where they have arrived at this solution for the same problem: [rspec/matchers/dsl.rb#L538-L540](https://github.com/rspec/rspec-expectations/blob/43bf64b01f8356979ffbc373b2e81d2ab1389b29/lib/rspec/matchers/dsl.rb#L538-L540)

Currently, my team is hacking around this with:

```ruby
# What we are currently doing without this patch (we're on Ruby 3.0.1)
def has_updated?(model, ruby2_attrs = {}, **ruby3_attrs)
  attrs = ruby2_attrs.merge(ruby3_attrs)
  # ...
end

# What we will do with this patch
def has_updated?(model, **attrs)
  # ...
end
```

### DIFF
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -21,9 +21,8 @@ module RSpec::Rails
 
     it "handles methods invoked via `method_missing` that use keywords" do
       example = group.new
-      example.define_singleton_method(:has_val?) { |val:| val == 1 }
-      expect(example).to example.have_val(val: 1)
-      expect(example).to_not example.have_val(val: 2)
+      example.define_singleton_method(:keyword_value) { |value:| value }
+      expect(example.keyword_value(value: 42)).to eq 42
     end
 
     context "with implicit subject" do
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -19,6 +19,13 @@ module RSpec::Rails
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles dynamic matchers with keywords (big change from ruby 2 to ruby 3)" do
+      example = group.new
+      example.define_singleton_method(:has_val?) { |val:| val == 1 }
+      expect(example).to example.have_val(val: 1)
+      expect(example).to_not example.have_val(val: 2)
+    end
+
     context "with implicit subject" do
       it "uses the controller as the subject" do
         controller = double('controller')
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -19,7 +19,7 @@ module RSpec::Rails
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
-    it "handles dynamic matchers with keywords (big change from ruby 2 to ruby 3)" do
+    it "handles dynamic matchers with keywords" do
       example = group.new
       example.define_singleton_method(:has_val?) { |val:| val == 1 }
       expect(example).to example.have_val(val: 1)
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -21,8 +21,9 @@ module RSpec::Rails
 
     it "handles methods invoked via `method_missing` that use keywords" do
       example = group.new
-      example.define_singleton_method(:keyword_value) { |value:| value }
-      expect(example.keyword_value(value: 42)).to eq 42
+      example.define_singleton_method(:has_val?) { |val:| val == 1 }
+      expect(example).to example.have_val(val: 1)
+      expect(example).to_not example.have_val(val: 2)
     end
 
     context "with implicit subject" do
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -19,7 +19,7 @@ module RSpec::Rails
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
-    it "handles dynamic matchers with keywords" do
+    it "handles methods invoked via `method_missing` that use keywords" do
       example = group.new
       example.define_singleton_method(:has_val?) { |val:| val == 1 }
       expect(example).to example.have_val(val: 1)
--- a/lib/rspec/rails/example/controller_example_group.rb
+++ b/lib/rspec/rails/example/controller_example_group.rb
@@ -176,6 +176,7 @@ module RSpec
           super
         end
       end
+      ruby2_keywords :method_missing if respond_to?(:ruby2_keywords, true)
 
       included do
         subject { controller }


spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -19,6 +19,13 @@ def group_for(klass)
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles dynamic matchers with keywords (big change from ruby 2 to ruby 3)" do
+      example = group.new
+      example.define_singleton_method(:has_val?) { |val:| val == 1 }
```
pirj, line 24: The code comment to `method_missing` claims:
>      # If method is a named_route, delegates to the RouteSet associated with
>      # this controller.

Do you think this statement still stands?
You mention dynamic matchers, what is the real case scenario that failed for you?

pirj (approved):
Looks good, thank you.
Do you mind adding a Changelog entry?

The discrepancy between the code comment and example description might be a bit confusing. Do you think it makes sense to align them?

JonRowe (changes requested):
:wave: So we're aware of this change as you can see from other code, and have tests in place to ensure they work correctly, I'm not sure what exactly you're changing here but it would only affect controller specs? rspec-rails doesn't implement dynamic matchers at all so this fix is at best badly named...

spec/rspec/rails/example/controller_example_group_spec.rb
JoshCheek: > Do you think this statement still stands?

Yeah, that's what the `if` statement does ([here](https://github.com/rspec/rspec-rails/blob/e4b3953561ae5193459685ec009e91ae3705debb/lib/rspec/rails/example/controller_example_group.rb#L173-L174)). The method did not intend to change the arguments, and in older versions of Ruby, it didn't. But with the new keyword change, this method accidentally moved the keyword arguments into the ordinal arguments. In Ruby 3, the way to do what this method used to do is to accept an ellipsis instead of an asterisk. Or to accept an asterisk for the ordinals and two asterisks for the keywords. But you can't use those if you also need to support Ruby 2.

Anyway here's an example showing how the intermediate `method_missing` accidentally changes the structure of the arguments:

```sh
# You *must* be on Ruby 3 to see the issue
ruby -v -W0 -e '
 # The real method_missing, this looks at the name, realizes it matches the `has_whatever`
 # matcher, and constructs / returns RSpec::Matchers::Has, out of the arguments
 def method_missing(name, a:)
   "SUCCESS! `#{name}(a: #{a.inspect})` was invoked"
 end

 # ControllerExampleGroup\'s method_missing. This checks if it should do stuff
 # related to route set (or whatever), decides it shouldn\'t, and passes it up
 # to the overwritten `method_missing`, thinking it has had no impact, but in
 # reality, it moved the keyword arguments into the ordinal arguments
 def self.method_missing(*)
   super
 end

 def self.HEADER(h) = puts("\n\e[33m#{h}\e[0m")

 HEADER "Invocation will fail because intermediate method_missing changes args"
 have_failure a: 111111111 rescue p $!

 HEADER "Now we will fix the args"
 singleton_class.class_eval { ruby2_keywords :method_missing }

 HEADER "Now invocation will succeed"
 puts have_success a: 222222222
'
```

Output:
image


> You mention dynamic matchers, what is the real case scenario that failed for you?

The same as what is shown. I had a helper object that made assertions, people thought it wasn't "RSpecy" enough, so I changed `obj.assert_whatever arg: val` to `expect(obj).to have_whatever(arg: val)`, which failed in the controller spec, because the controller spec included this module.

Here is a standalone example, I tried to make it slightly more realistic, since my test apparently didn't communicate it. Run it in Ruby 3 and it will fail, uncomment line 7 and it will pass:

```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'actionpack', '= 6.1.4', require: 'action_controller/railtie'
  gem 'rspec-rails', '= 5.0.1', require: 'rspec/rails'
end

Class.new Rails::Application

module RSpec::Rails::ControllerExampleGroup
  # ruby2_keywords :method_missing  # <-- Uncomment this line and it will pass
end

DomainObject = Struct.new :string do
  def has_attribute?(encoding:) = string.encoding == encoding
end

RSpec.describe 'some controller', type: :controller do
  it 'whatever' do
    obj = DomainObject.new 'abc'.encode('utf-8')
    expect(obj).to have_attribute encoding: Encoding::UTF_8
  end
end
```

JoshCheek:
>👋 So we're aware of this change as you can see from other code, and have tests in place to ensure they work correctly, I'm not sure what exactly you're changing here but it would only affect controller specs? rspec-rails doesn't implement dynamic matchers at all so this fix is at best badly named...

Perhaps something more technical, such as `it "does not alter keyword missing's keyword arguments"`?

spec/rspec/rails/example/controller_example_group_spec.rb
JoshCheek: Hey, I just changed that first code example. I forgot that bash doesn't understand escaping in single quotes. I added a dollar sign in front of the argument to `ruby -e`, and now it works in bash. Sorry for not verifying that before posting it 😖

pirj: Yes, I clearly understand why `ruby2_keywords` is needed, and that it is the only approach that works in all Ruby versions that we strive to support. We have it on our roadmap to improve that for all RSpec libraries, but the effort is beyond our current capacity and focus. See https://github.com/rspec/rspec-expectations/issues/1241#issuecomment-749621016

My note was about an entirely different thing. 

The `method_missing` is catching and delegating calls to named routes, and falls back to superclass's implementation otherwise.
But your example is titled "handles dynamic matchers", you define a predicate matcher `has_val?` and `have_val`.
This mismatch caught my eye, as it is confusing.

Do you mind changing the example so that it refers to a named route instead? Does it make sense?

Or was the original problem that you've experienced related to dynamic matchers?
If so, it might be `rspec-core` or `rspec-expectations` that need to be patched instead or in addition to.

pirj: Alright, I might have misunderstood the whole thing. Your last example clearly indicates the problem.

So, the kwargs error is raised when there's no route, and we delegate to `super`. But when `ruby2_keywords` isn't there, it fails. This case with `super` with no explicit arguments slipped my mind, please accept my apologies.

So this has nothing to do with handling routes specifically.

Let me take a closer look at an earlier time of the day this weekend. Presumably everything is in order.

JoshCheek: > Alright, I might have misunderstood the whole thing. Your last example clearly indicates the problem.
>
> So, the kwargs error is raised when there's no route, and we delegate to super. But when ruby2_keywords isn't there, it fails. This case with super with no explicit arguments slipped my mind, please accept my apologies.

No worries, it's exceedingly tricky to understand, IME, b/c the spots with the problems can go undetected for a long time and then explode somewhere else completely. And the mental model to understand what the problem is and what the solution is is also very nuanced. I probably put 15 hours of pretty intense analysis into a similar bug in a different library before I actually understood why things weren't working and what to do about it (and I still only partially understand why we're doing any of this 🤷).

> So this has nothing to do with handling routes specifically.

Correct, the `else` clause, which was intended to be a noop, has the unintentional side-effect of changing the invocation signature in Ruby 3.

> Let me take a closer look at an earlier time of the day this weekend. Presumably everything is in order.

👍

JoshCheek:
Okay, I changed the name to be more technical, and then felt like it wasn't very obvious how the test was an example of the description. So I changed it to not depend on the group at all. But IDK if this is better or worse, so figure I'll get y'all's feedback before pushing it.

```ruby
    it "does not alter method_missing's keyword arguments" do
      klass = Class.new.prepend described_class
      klass.define_method(:method_missing) { |a, b, c:, &d| {a: a, b: b, c: c, d: d} }
      block = -> { }
      expect(klass.new.a(:b, c: :c, &block)).to eq a: :a, b: :b, c: :c, d: block
    end
```

JoshCheek:
Like for me, it makes sense, but I'm a little worried it might be too terse/magical for other people to read.

pirj:
I suggest keeping the current example code. It is clear that the method is not a named route. Dynamic matcher is one of completely legitimate cases from all possible options.

The title, though, is too specific. It mentions dynamic matchers, while it can be anything else.
What do you think of "it correctly delegates keyword arguments to the superclass implementation"? Does it make sense to clarify "when the method is not a named route helper"?

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -19,6 +19,13 @@ def group_for(klass)
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles dynamic matchers with keywords" do
```
JonRowe, line 22:
```suggestion
    it "handles methods invoked via `method_missing` that use keywords" do
```

JonRowe (changes requested):

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -19,6 +19,13 @@ def group_for(klass)
...
+      example = group.new
+      example.define_singleton_method(:has_val?) { |val:| val == 1 }
+      expect(example).to example.have_val(val: 1)
+      expect(example).to_not example.have_val(val: 2)
```
JonRowe, line 26:
```suggestion
      example.define_singleton_method(:keyword_value) { |value:| value }
      expect(example.keyword_value(value: 42)).to eq 42
```

JonRowe:
My understanding here is that the `method_missing` in controller specs is dropping keywords without this check, but it doesn't actually have anything to do with matchers, so I've requested some changes to reflect that

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -21,9 +21,8 @@ def group_for(klass)
 
     it "handles methods invoked via `method_missing` that use keywords" do
       example = group.new
-      example.define_singleton_method(:has_val?) { |val:| val == 1 }
-      expect(example).to example.have_val(val: 1)
-      expect(example).to_not example.have_val(val: 2)
+      example.define_singleton_method(:keyword_value) { |value:| value }
+      expect(example.keyword_value(value: 42)).to eq 42
```
pirj, line 25: I'll roll this change back, it fails to protect from regression. The reason is that `method_missing` is not called, because we call the same method that we define. Previously, we were defining `has_val?` and were calling `have_val`. `super` then calls a correct method thanks to dynamic matchers.

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -19,6 +19,13 @@ def group_for(klass)
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles methods invoked via `method_missing` that use keywords" do
+      example = group.new
+      example.define_singleton_method(:has_val?) { |val:| val == 1 }
+      expect(example).to example.have_val(val: 1)
```
pirj, line 25: I'd suggest changing this to:
```ruby
expect { example.have_val(val: 1) }.not_to raise_error
```
as `expect(example)` doesn't participate much and isn't even called when there's this failure. As the exception is coming from `example.have_val` call.
Also, `example.have_val(val: 1)` isn't a valid matcher you can pass to `.to`.

Please let me know what are your thoughts on this.

pirj: > Also, `example.have_val(val: 1)` isn't a valid matcher you can pass to `.to`

My mistake. It is a matcher, since `have_val` is backed by dynamic matchers.

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: I'd prefer a clear example method here not something imitating a matcher as thats not what is the issue.

pirj: This test is not hitting the `method_missing`. We defined `keyword_value` and call it.

The current `expect(example).to example.have_val(val: 1)` works, but it's incredibly confusing.
```ruby
expect { example.have_val(val: 1) }.not_to raise_error
```
failed to catch the regression, it doesn't fail if `ruby2_keywords` is removed.

I'll figure out a self-explanatory spec to cover this. I intend to rely on dynamic matchers, but will try to avoid the `expect(example).to` part.

spec/rspec/rails/example/controller_example_group_spec.rb
pirj: A couple of findings:

1. `super` for `method_missing` is (`actionpack-6.0.4/lib/action_dispatch/testing/assertions/routing.rb`):
```ruby
      # ROUTES TODO: These assertions should really work in an integration context
      def method_missing(selector, *args, &block)
        if defined?(@controller) && @controller && defined?(@routes) && @routes && @routes.named_routes.route_defined?(selector)
          @controller.send(selector, *args, &block)
        else
          super
        end
      end
```

And as of Rails 6.0.4 **it doesn't have** `ruby2_keywords` to it.

2. What triggers the error

```ruby
        def predicate_result
          @predicate_result = actual.__send__(predicate, *@args, &@block)
        end
```

the error happens `after` `method_missing` has called `super`:
```ruby
      example = group.new
      example.define_singleton_method(:has_val?) { |val:| val == 1 }
      dynamic_matcher = example.have_val(val: 1)
      require 'pry'; binding.pry # 
      dynamic_matcher.matches?(example)
```

`matches?`:
```ruby
predicate_accessible? && predicate_matches?
```

`predicate_matches?`:
```ruby
value == predicate_result
```

`predicate_result`:
```ruby
def predicate_result
  @predicate_result = actual.__send__(predicate, *@args, &@block)
end
```

and this `__send__` does trigger.

I barely understand how `ruby2_keywords` on a `method_missing` fixes it.
`predicate` evaluates to `:has_val?`.
```ruby
@args.inspect # => "[{:val=>1}]"
```

Adding `ruby2_keywords :predicate_result` doesn't help.

We do already have `ruby2_keywords` before `method_missing` in dynamic matcher:
```ruby
    def method_missing(method, *args, &block)
      case method.to_s
      when BE_PREDICATE_REGEX
        BuiltIn::BePredicate.new(method, *args, &block)
      when HAS_REGEX
        BuiltIn::Has.new(method, *args, &block)
      else
        super
      end
    end
    ruby2_keywords :method_missing if respond_to?(:ruby2_keywords, true)
```

and `initialize`, too:
```ruby
        def initialize(method_name, *args, &block)
          @method_name, @args, @block = method_name, args, block
        end
        ruby2_keywords :initialize if respond_to?(:ruby2_keywords, true)
```

If I add `ruby2_keywords` to `method_missing` in `controller_example_group.rb`, `@args.inspect` and `predicate` don't change.

`actual` is:
```ruby
> actual # => #<RSpec::ExampleGroups::ApplicationController_2 (no description provided)>
```

and the difference between having and not having `ruby2_keywords before `method_missing` in `ControllerExampleGroup` boils down to:
```ruby
(#<RSpec::Matchers::BuiltIn::Has>)> actual.has_val?(*@args)
ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: val)
```
vs
```ruby
(#<RSpec::Matchers::BuiltIn::Has>)> actual.has_val?(*@args)
=> true
```

while `method_missing` is **not** involved in this call.

I'm out of ideas why this happens, and I tend to think it's some weird side effect, and it's time to summon Jeremy and Benoit for help.

pirj: Please don't get me wrong, I appreciate that the fix worked, and not opposed to it. My Ruby knowledge of kwargs clearly fails to provide me with understanding why the fix works.

I've expanded the example you came up with, and it works, and prevents from regression, but what it is very suspicious.

If something fishy is going on, and we're addressing the issue incorrectly, or some additional Ruby fix is needed. It's not something that didn't happen in the past.

pirj: In both cases,
```ruby
> actual.method(:has_val?).parameters
=> [[:keyreq, :val]]
```

pirj: Bingo!

This is especially confusing (`ruby2_keywords` is present before `method_missing`):
```ruby
    57: def predicate_result
    58:   require 'pry'; binding.pry
 => 59:   @predicate_result = actual.__send__(predicate, *@args, &@block)
    60: end

> @args
=> [{:val=>1}]
> actual.has_val?(*@args)
=> true

> a = [{val: 1}]
=> [{:val=>1}]
> actual.has_val?(*a)
ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: val)
from /Users/pirj/source/rspec-dev/repos/rspec-rails/spec/rspec/rails/example/controller_example_group_spec.rb:24:in `block (3 levels) in <module:Rails>'
```

I vaguely recall that an Array has some internal meta-attribute to it that indicates that it might be passed and expanded to keywords.
And this where it gets ridiculously hard to understand.

So, apparently, when we initially pass `val: 1` to `have_value`, it's tagged as such. We then save it to `@args` in dynamic matchers. And then, when dynamic matchers call the real predicate, `*@args` expand it to kwargs properly.

Without `ruby2_keywords`, this tagging doesn't happen.

And, obviously, to write a regression-preventing spec around it that is not perplexing at the same time.

@eregon Can you please confirm this theory? Do I understand it correctly that there is a special tag that allows an Array to be expanded to kwargs?

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: I just don't want a `have_` anything style example, its confusing, its the method dispatch thats the bug here, not matchers.

spec/rspec/rails/example/controller_example_group_spec.rb
eregon: See https://docs.ruby-lang.org/en/master/Module.html#method-i-ruby2_keywords and https://docs.ruby-lang.org/en/master/Hash.html#method-c-ruby2_keywords_hash-3F
The flag is on the Hash itself, and it only matters for calls with a splat, where the flagged Hash is the last element.

pirj: Thank you!
TIL `ruby2_keywords_hash?` 👍

spec/rspec/rails/example/controller_example_group_spec.rb
pirj: It's a bit of both.

To reproduce the issue:
 - it is not sufficient for us to just call the same method that we define, as `method_missing` won't be called
 - there should be some code that handles `method_missing` below `ControllerExampleGroup` and below Rails
 - this code should accept arguments and pass it to a method that accepts keyword arguments

We could write a spec with something like `prepend Module.new { def method_missing ... }`, but this might be excessively complicated.

There are dynamic matchers. They do exactly what we want - they have a `method_missing`, and they can delegate to the method we define.

It's still a brain-teaser how to express the case in a spec. The best I could think of so far:
```ruby
expect {
  example.have_val(val: 1).matches?(example)
}.not_to raise_error
```

`example.have_val` goes through `ControllerExampleGroup`'s `method_missing`, then through Rails' one, and lands in dynamic matchers. They save the kwarg to `@args`, and the flag that it's a kwarg is already set on the object.
`matches?` calls `has_val?` on the `example`, and passes `*@args` to it.

It's not necessary to call `have_val` on `example`, and we could:
```ruby
expect(example).to have_val(val: 1)
```
but in this case `ControllerExampleGroup`'s `method_missing` is never called, and it's dynamic matchers' `method_missing` who sets the flag. I.e. this fix would not be covered by such a test.

Frankly speaking, it seems like it doesn't matter much what exactly the spec will be, we'll have to add a chunk of code comments along. And I have strong doubts that in a month from now it will take me less than half an hour to recall all the details about this tangled case.

I'm out of good ideas.

spec/rspec/rails/example/controller_example_group_spec.rb
JoshCheek: > `predicate_result`:
> ```ruby
> def predicate_result
>   @predicate_result = actual.__send__(predicate, *@args, &@block)
> end
> ```
>
> and this `__send__` does trigger.
>
> I barely understand how `ruby2_keywords` on a `method_missing` fixes it.
> `predicate` evaluates to `:has_val?`.
> ```ruby
> @args.inspect # => "[{:val=>1}]"
> ```
>
> Adding `ruby2_keywords :predicate_result` doesn't help.
>
> -- https://github.com/rspec/rspec-rails/pull/2509#discussion_r678474813

Yeah, it needs to called where the hash is created, because it's not about how our code calls the method or passes the arguments (with the `ruby2_keywords` solution, anyway).

It looks like you figured it out, but I wrote up this example before I read that, so I'll post it anyway. This is what finally made it click for me, b/c my natural implementation expectations were incorrect and making it harder for me to reason about:

```ruby
def call_this(val:)
  "val is #{val}"
end

def record_args(*args)
  args
end

args = record_args val: 1 # => [{:val=>1}]

# It's a normal hash, Ruby won't pass it into the keywords slot
Hash.ruby2_keywords_hash? args.last # => false
call_this *args rescue $!           # => #<ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: val)>

# We can explicitly turn it into a keywords hash
keywords = args[-1]
Hash.ruby2_keywords_hash? keywords           # => false
keywords = Hash.ruby2_keywords_hash keywords # => {:val=>1}
Hash.ruby2_keywords_hash? keywords           # => true

# Which will indicate to Ruby that it should pass it into the keywords slot
call_this *args rescue $! # => #<ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: val)>
args[-1] = keywords
call_this *args rescue $! # => "val is 1"

# `ruby2_keywords` modifies the method to implicitly do what we did above
Hash.ruby2_keywords_hash? record_args(val: 1).last # => false
ruby2_keywords :record_args
Hash.ruby2_keywords_hash? record_args(val: 1).last # => true

# So now any method anywhere can splat that array into args
call_this *record_args(val: 1) # => "val is 1"
```

JoshCheek: Playing with it a bit more, I figured out how they do it. It's a flag on the hash. https://gist.github.com/JoshCheek/43b15b9123e337ee6a118cb6ff4c7e7a

pirj: Bitwise precision aim at the root cause! 👏

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: If my change doesn't trigger method missing neither does the original, they both call the method directly I just skip the expect to. So my original point stands here, please construct an example that doesn't look like a matcher.

pirj: @JonRowe There is a significant difference. You call a method that we've just defined, `keyword_value`, while the original example calls a non-existent method, `have_val`, while we define `has_val?`.

JonRowe: Ah tired eyes etc, ok fine, implement a method missing and remove the has / have names. This is a lot of conversation for a simple fix, but it needs to not be confused with matchers. Its method_missing and the method names that are causing this.

pirj: Right. Dynamic matchers are a victim here.
And they are the only known victim. And they apparently are the only way to test the fix without involving `in_sub_process { ControllerExampleGroup.prepend Module.new { def method_missing ... end } }`.

JonRowe:
Closing in favour of #2514 which has a spec demonstrating the issue in isolation, thanks for the help @JoshCheek

PR #2510, opened 2021-07-19
Add .circleci/config.yml

iamharish:


pirj:
Thank you, we're quite fine with GHA.

PR #2511, opened 2021-07-30
Use RSpec.current_scope

pirj:
It was introduced in https://github.com/rspec/rspec-core/pull/2895

When `RSpec.current_scope` is released in [3.11](https://github.com/rspec/rspec-core/pull/2895) and [4.0](https://github.com/rspec/rspec-core/pull/2910), we're fine with merging this PR to `main` and to `5-0-maintenance`.

TODO:
 - [x] remove the `temp!` commit

### DIFF
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -46,7 +46,7 @@ module RSpec
             def proxy_method_warning_if_called_in_before_context_scope(method_name)
               orig_implementation = instance_method(method_name)
               define_method(method_name) do |*args, &blk|
-                if inspect.include?("before(:context)")
+                if RSpec.current_scope == :before_context_hook
                   RSpec.warn_with("Calling fixture method in before :context ")
                 else
                   orig_implementation.bind(self).call(*args, &blk)


JonRowe (approved):
For when the other PR is merged, this gets a thumbs up.

JonRowe (approved):

pirj:
If I merge now, I'll surely forget to cherry-pick to `5-0-maintenance`.

If I cherry-pick it to `5-0-maintenance`, I would have to bump `expected_rspec_version` in `gemspec` to the yet unreleased 3.11.0.

In theory, pointing to 3.11.0 might break someone's setup if they prefer using `5-0-maintenance` with the latest unreleased fixes to 5.0.x.
In practice, though, [there are no changes since 5.0.2](https://github.com/rspec/rspec-rails/commits/5-0-maintenance/), so for those using `github: ..., branch ...` there's a workaround to use `~> 5.0`.

If I forget to cherry-pick to `5-0-maintenance`, it will only be available for the future 6.0 release. Not a big deal, though.

This process requires thinking and having good memory.
I intend to hold off merging this until 3.11.0 is released, does this sound reasonable?

JonRowe:
I don't think this should be merged to `5-0-maintenance` as I don't think its a good idea to bump a dependency in a patch version, only a minor. However this also would gate any release of `rspec-rails` `5.1` on `3.11` so its probably still worth holding off merging.

PR #2514, opened 2021-08-13
Fix keyword arguments in dynamic methods in controller specs

JonRowe:
Supersedes #2509 by demonstrating the issue in isolation from the rest of RSpec, many thanks to the original reporter @joshcheek

### DIFF
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -25,6 +25,10 @@ module RSpec::Rails
           def a_method(value:); value; end
           def method_missing(_name, *_args, **kwargs, &_block); a_method(**kwargs); end
 
+          # This example requires prepend so that the `method_missing` definition
+          # from `ControllerExampleGroup` bubbles up to our artificial one, in reality
+          # this is likely to be either an internal RSpec dispatch or one from a 3rd
+          # party gem.
           prepend ControllerExampleGroup
         end
       example = group.new
--- a/spec/rspec/rails/example/controller_example_group_spec.rb
+++ b/spec/rspec/rails/example/controller_example_group_spec.rb
@@ -19,6 +19,19 @@ module RSpec::Rails
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles methods invoked via `method_missing` that use keywords" do
+      group =
+        RSpec::Core::ExampleGroup.describe ApplicationController do
+          def a_method(value:); value; end
+          def method_missing(_name, *_args, **kwargs, &_block); a_method(**kwargs); end
+
+          prepend ControllerExampleGroup
+        end
+      example = group.new
+
+      expect(example.call_a_method(value: :value)).to eq :value
+    end
+
     context "with implicit subject" do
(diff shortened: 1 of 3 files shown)


benoittgt (approved):
LGTM.

spec/rspec/rails/example/controller_example_group_spec.rb
```diff
@@ -19,6 +19,23 @@ def group_for(klass)
       expect(group.included_modules).to include(RSpec::Rails::Matchers::RoutingMatchers)
     end
 
+    it "handles methods invoked via `method_missing` that use keywords" do
+      group =
+        RSpec::Core::ExampleGroup.describe ApplicationController do
+          def a_method(value:); value; end
+          def method_missing(_name, *_args, **kwargs, &_block); a_method(**kwargs); end
```
pirj, line 26: WDYT of writing this like this instead:
```diff
- def method_missing(_name, *_args, **kwargs, &_block); a_method(**kwargs); end
+ def method_missing(_name, *args, &_block); a_method(*args); end
+ ruby2_keywords :method_missing if respond_to?(:ruby2_keywords, true)
```

The error message (without the fix) is:
```
     ArgumentError:
       wrong number of arguments (given 1, expected 0; required keyword: value)
 ```
instead of:
```
     ArgumentError:
       missing keyword: :value
```

File: spec/rspec/rails/example/controller_example_group_spec.rb [hunk 2]
```diff
@@ -19,6 +19,23 @@ def group_for(klass)
...
+
+          # This example requires prepend so that the `method_missing` definition
+          # from `ControllerExampleGroup` bubbles up to our artificial one, in reality
+          # this is likely to be either an internal RSpec dispatch or one from a 3rd 
+          # party gem.
+          prepend ControllerExampleGroup
```
pirj, line 32: Is this statement purposefully put to the end of the block?
Moving it above own's `method_missing` definition makes no difference, and includes/prepends are commonly at the beginning of the class definition.

pirj (approved):
The test looks cleaner and more isolated.

Subjectively, I would understand the `method_missing` if it was defined with a proper way of delegating. Thus, not putting a shade of doubt on its participation in the failure.

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: Its purposefully demonstrating the mis match between keyword arguments, on Ruby 3.0.2 on my machine the error is:

```
Failure/Error: def a_method(value:); value; end

     ArgumentError:
       missing keyword: :value
```

spec/rspec/rails/example/controller_example_group_spec.rb
JonRowe: This was the implementation of a failing spec I was happy with

spec/rspec/rails/example/controller_example_group_spec.rb
pirj: Agree. Tried a couple of different cases (dafault kwargs, do passing an argument in addition to kwargs), and with your approach, the failure message is consistent.

pirj:
Thanks!

PR #2515, opened 2021-08-29
Remove redundunt include Capybara::DSL

YusukeIwaki:
## Overivew

Both Capybara and rspec-rails includes Capybara::DSL and Capybara::RSpecMatchers for feature/system specs.

```
begin
  require 'capybara/rspec'
rescue LoadError
end

begin
  require 'capybara/rails'
rescue LoadError
end

if defined?(Capybara)
  RSpec.configure do |c|

    binding.pry ### BREAKPOINT HERE
    if defined?(Capybara::DSL)
      c.include Capybara::DSL, type: :feature
      if defined?(ActionPack) && ActionPack::VERSION::STRING >= "5.1"
        c.include Capybara::DSL, type: :system
```

When I watch the configuration by setting a breakpoint above with an actual Rails application, configuration instance already has Capybara::DSL and RSpecMatchers

```
 @include_modules=
  #<RSpec::Core::FilterableItemRepository::QueryOptimized:0x00007f7f7aad00d8
   @applicable_keys=#<RSpec::Core::Set:0x00007f7f7aad0038 @values={:capybara_feature=>true, :type=>true}>,
   @applies_predicate=:any?,
   @items_and_filters=
    [[#<RSpec::Core::SharedExampleGroupModule "Capybara Features">, {:capybara_feature=>true}],
     [Capybara::DSL, {:type=>:feature}],
     [Capybara::RSpecMatchers, {:type=>:feature}],
     [Capybara::DSL, {:type=>:system}],
     [Capybara::RSpecMatchers, {:type=>:system}],
     [Capybara::RSpecMatchers, {:type=>:view}]],
   @memoized_lookups={},
```

So no need to include Capybara::DSL for feature/system specs.

## Direction

Remove redundunt includes, and leave the setup of feature/system spec to Capybara.

### DIFF
--- a/lib/rspec/rails/vendor/capybara.rb
+++ b/lib/rspec/rails/vendor/capybara.rb
@@ -10,20 +10,18 @@ end
 
 if defined?(Capybara)
   RSpec.configure do |c|
-    if defined?(Capybara::DSL)
-      c.include Capybara::DSL, type: :feature
-      if defined?(ActionPack) && ActionPack::VERSION::STRING >= "5.1"
-        c.include Capybara::DSL, type: :system
-      end
-    end
+    # 'capybara/rspec' defines configurations below.
+    #
+    # config.include Capybara::DSL, type: :feature
+    # config.include Capybara::RSpecMatchers, type: :feature
+    # config.include Capybara::DSL, type: :system
+    # config.include Capybara::RSpecMatchers, type: :system
+    # config.include Capybara::RSpecMatchers, type: :view
 
     if defined?(Capybara::RSpecMatchers)
-      c.include Capybara::RSpecMatchers, type: :view
       c.include Capybara::RSpecMatchers, type: :helper
       c.include Capybara::RSpecMatchers, type: :mailer
       c.include Capybara::RSpecMatchers, type: :controller
-      c.include Capybara::RSpecMatchers, type: :feature
-      c.include Capybara::RSpecMatchers, type: :system
     end
 
     unless defined?(Capybara::RSpecMatchers) || defined?(Capybara::DSL)


YusukeIwaki:
Another solution for duplicated including Capybara::DSL is to add a condition in Capybara side like below:

https://github.com/teamcapybara/capybara/blob/master/lib/capybara/rspec.rb

```
RSpec.configure do |config|
  unless defined?(RSpec::Rails) ## ADD THIS
    config.include Capybara::DSL, type: :feature
    config.include Capybara::RSpecMatchers, type: :feature
    config.include Capybara::DSL, type: :system
    config.include Capybara::RSpecMatchers, type: :system
    config.include Capybara::RSpecMatchers, type: :view
  end
```

Which library should setup Capybara::DSL and RSpecMatchers...?

pirj:
What is the problem exactly you're dealing with?

Are you certain that earlier versions (< 3.34.0) of `capybara` include `Capybara::DSL` and `Capybara::RSpecMatchers` to examples marked as `type: :system`? I see [this commit](https://github.com/teamcapybara/capybara/commit/5b68043516451c3831fcc9833442a9feb4f2415d), and it may result in a surprise for those who update `rspec-rails`, but keep their old `capybara` version.
I didn't dig deep, but this may be the case for other `include`s as well.

YusukeIwaki:
Hi, thank you for you kind feedback.

Actually this PR is not originated from concrete usecases. I just noticed the duplicated includes when I studied how to use RSpec and Capybara without Capybara DSL.
The duplicated includes doesn't affect in actual use, but I was a little confused why they are duplicated and which library is really responsible for including Capybara::DSL.

> Are you certain that earlier versions (< 3.34.0) of capybara include Capybara::DSL and Capybara::RSpecMatchers to examples marked as type: :system? 

I didn't know this, and the advice really helped me, thank you. :)
Now I understand rspec-rails have to include Capybara::DSL for supporting older version of Capybara.
The direction of this PR is incorrect, and I will close this PR.

PR #2516, opened 2021-09-10
Fix parameter matching with mail delivery job and `ActionMailer::MailDeliveryJob`

### LINKED ISSUE #2351: have_enqueued_mail with options doesn't match
Because of https://github.com/rails/rails/commit/f5050d998def98563f8fa4b381c09f563681f159
have_enqueued_mail with options is broken
for example, we have a Mailer service:
```
class MailerTestService
  def save
    MailtestMailer.mailtest(option: { foo: 'bar' }).deliver_later
  end
end
```

and test
```
RSpec.describe MailerTestService, type: :service do
  it do
    expect { described_class.new.save }
      .to have_enqueued_mail(MailtestMailer, :mailtest)
      .with(option: { foo: 'bar' })
  end
end
```

it will not match because arguments serialization in https://github.com/rails/rails/commit/f5050d998def98563f8fa4b381c09f563681f159#diff-e264d3f0104c3e213b6402459897ee61R144 updated, but not updated yet in rspec-rails

Already working on a solution, will push PR soon.

related #2184

fabn:
This should fix #2351. 

Maybe it's a breaking change for Rails 6 users, I don't know but it will allow users to easily upgrade to rails 6 without changing tons of specs.

Let me know if it's ok, it's my first PR in rspec.

### DIFF
--- a/features/matchers/have_enqueued_mail_matcher.feature
+++ b/features/matchers/have_enqueued_mail_matcher.feature
@@ -38,3 +38,40 @@ Feature: have_enqueued_mail matcher
       """
     When I run `rspec spec/mailers/user_mailer_spec.rb`
     Then the examples should all pass
+
+  Scenario: Checking mailer arguments
+    Given a file named "app/mailers/my_mailer.rb" with:
+      """ruby
+      class MyMailer < ApplicationMailer
+
+        def signup(user = nil)
+          @user = user
+
+          mail to: "to@example.org"
+        end
+      end
+      """
+    Given a file named "spec/mailers/my_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe MyMailer do
+        it "matches with enqueued mailer" do
+          ActiveJob::Base.queue_adapter = :test
+          # Works with plain args
+          expect {
+            MyMailer.signup('user').deliver_later
+          }.to have_enqueued_mail(MyMailer, :signup).with('user')
+          # Works with named parameters
+          expect {
+            MyMailer.with('foo' => 'bar').signup.deliver_later
+          }.to have_enqueued_mail(MyMailer, :signup).with('foo' => 'bar')
+          # Works also with both, named parameters match first argument
+          expect {
+            MyMailer.with('foo' => 'bar').signup('user').deliver_later
+          }.to have_enqueued_mail(MyMailer, :signup).with({'foo' => 'bar'}, 'user')
+        end
+      end
+      """
+    When I run `rspec spec/mailers/my_mailer_spec.rb`
+    Then the examples should all pass
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -76,7 +76,7 @@ module RSpec
         def arguments_match?(job)
           @args =
             if @mail_args.any?
-              base_mailer_args + @mail_args
+              base_mailer_args + process_arguments(job, @mail_args)
             elsif @mailer_class && @method_name
               base_mailer_args + [any_args]
             elsif @mailer_class
@@ -86,6 +86,18 @@ module RSpec
             end
 
           super(job)
+        end
+
+        def process_arguments(job, given_mail_args)
+          if job[:job] == ActionMailer::MailDeliveryJob
+            if given_mail_args.first.is_a?(Hash) && job[:args][3]['params'].present?
+              [hash_including(params: given_mail_args[0], args: given_mail_args.drop(1))]
+            else
+              [hash_including(args: given_mail_args)]
+            end
+          else
+            given_mail_args
+          end
         end
 
         def base_mailer_args
@@ -120,7 +132,6 @@ module RSpec
 
         def mail_job_message(job)
           mailer_method = job[:args][0..1].join('.')
-
           mailer_args = job[:args][3..-1]
           msg_parts = []
           msg_parts << "with #{mailer_args}" if mailer_args.any?
--- a/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
+++ b/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
@@ -393,18 +393,22 @@ RSpec.describe "HaveEnqueuedMail matchers", skip: !RSpec::Rails::FeatureCheck.ha
         }.to have_enqueued_mail(UnifiedMailer, :test_email).and have_enqueued_mail(UnifiedMailer, :email_with_args)
       end
 
-      it "passes with provided argument matchers" do
+      it "matches arguments when mailer has only args" do
+        expect {
+          UnifiedMailer.email_with_args(1, 2).deliver_later
+        }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with(1, 2)
+      end
+
+      it "matches arguments when mailer is parameterized" do
         expect {
           UnifiedMailer.with('foo' => 'bar').test_email.deliver_later
-        }.to have_enqueued_mail(UnifiedMailer, :test_email).with(
-          a_hash_including(params: {'foo' => 'bar'})
-        )
+        }.to have_enqueued_mail(UnifiedMailer, :test_email).with('foo' => 'bar')
+      end
 
+      it "matches arguments when mixing parameterized and non-parameterized emails" do
         expect {
           UnifiedMailer.with('foo' => 'bar').email_with_args(1, 2).deliver_later
-        }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with(
-          a_hash_including(params: {'foo' => 'bar'}, args: [1, 2])
-        )
+        }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with({'foo' => 'bar'}, 1, 2)
       end
 
       it "passes when using a mailer with `delivery_job` set to a sub class of `ActionMailer::DeliveryJob`" do


lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -88,6 +88,18 @@ def arguments_match?(job)
           super(job)
         end
 
+        def process_arguments(job, given_mail_args)
+          if job[:job] == ActionMailer::MailDeliveryJob
+            if given_mail_args.first.is_a?(Hash) && job[:args][3]['params'].present?
```
pirj, line 93: Is it necessarily a `[3]` or it may depend on something?

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -88,6 +88,18 @@ def arguments_match?(job)
           super(job)
         end
 
+        def process_arguments(job, given_mail_args)
+          if job[:job] == ActionMailer::MailDeliveryJob
```
pirj, line 92: What if it's a custom delivery job set with [`config.action_mailer.delivery_job`](https://guides.rubyonrails.org/configuring.html)?

lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -88,6 +88,18 @@ def arguments_match?(job)
...
+            if given_mail_args.first.is_a?(Hash) && job[:args][3]['params'].present?
+              [hash_including(params: given_mail_args[0], args: given_mail_args.drop(1))]
+            else
+              [hash_including(args: given_mail_args)]
+            end
+          else
```
pirj, line 98: What the `job[:job]` is in this case?

pirj (commented):
Please feel free to locally disable `Metrics/ClassLength` for this class.

lib/rspec/rails/matchers/have_enqueued_mail.rb
fabn: I think is guaranteed that `[3]` always point to mail job args, especially when using one of builtin mailers:

In these example you have:

```
 :args=>["UnifiedMailer", "test_email", "deliver_now", {"params"=>{"foo"=>"bar", "_aj_symbol_keys"=>[]}, "args"=>[], "_aj_symbol_keys"=>["params", "args"]}],
```

the same assumption for args is present in the above code when building `base_mailer_args`

lib/rspec/rails/matchers/have_enqueued_mail.rb
fabn: If job is a subclass of `ActionMailer::MailDeliveryJob` the code should work, maybe I can replace with `job[:job] <= ActionMailer::MailDeliveryJob` to also catch subclasses.

Also I'll add a `defined?(ActionMailer::MailDeliveryJob)` to avoid to break when running in rails < 6.0.

lib/rspec/rails/matchers/have_enqueued_mail.rb
fabn: Any other delivery job that's not `ActionMailer::MailDeliveryJob`, if it behaves like the existing one, then current code is fine, else the matcher would be broken also without this PR.

pirj (approved):
Good job! Thank you for the contribution.

features/matchers/have_enqueued_mail_matcher.feature
```diff
@@ -38,3 +38,40 @@ Feature: have_enqueued_mail matcher
...
+      RSpec.describe MyMailer do
+        it "matches with enqueued mailer" do
+          ActiveJob::Base.queue_adapter = :test
+          # Works with plain args
+          expect {
+            MyMailer.signup('user').deliver_later
+          }.to have_enqueued_mail(MyMailer, :signup).with('user')
+          # Works with named parameters
+          expect {
+            MyMailer.with('foo' => 'bar').signup.deliver_later
```
pirj, line 67: Should it be like that? Rails API docs suggest using a symbol for the key:
```ruby
InvitationsMailer.with(inviter: person_a, invitee: person_b).account_invitation.deliver_later
```

features/matchers/have_enqueued_mail_matcher.feature
```diff
@@ -38,3 +38,40 @@ Feature: have_enqueued_mail matcher
...
+          }.to have_enqueued_mail(MyMailer, :signup).with('foo' => 'bar')
+          # Works also with both, named parameters match first argument
```
pirj, line 69: WDYT of "Parameterize and pass an argument to the mailer"? And "Parameterize the mailer" for an example above.

features/matchers/have_enqueued_mail_matcher.feature
pirj: I think using a hash may induce confusion.
[Rails API example](https://api.rubyonrails.org/classes/ActionMailer/Parameterized.html):
```ruby
class InvitationsMailer < ApplicationMailer
  def account_invitation(inviter, invitee)
...
InvitationsMailer.account_invitation(person_a, person_b).deliver_later
```
Does it make sense to pass a non-hash argument? Is there a difference for the documentation purpose?
Was there a special case for named parameters that you've fixed in this PR? Is it covered sufficiently in specs?

features/matchers/have_enqueued_mail_matcher.feature
fabn: That's not bad.

features/matchers/have_enqueued_mail_matcher.feature
fabn: I added these examples only because they was missing in relish doc, there's no special case addressed.

fabn:
> Should it be like that? Rails API docs suggest using a symbol for the key:

Copy pasted from existing examples, I can change them

fabn:
@pirj Pushed a new commit to have different examples for each syntax

features/matchers/have_enqueued_mail_matcher.feature
```diff
@@ -38,3 +38,90 @@ Feature: have_enqueued_mail matcher
...
+    When I run `rspec spec/mailers/my_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Parameterize the mailer
+    Given a file named "app/mailers/my_mailer.rb" with:
+      """ruby
+      class MyMailer < ApplicationMailer
+
+        def signup(user = nil)
+          @user = user
```
pirj, line 77: Should this be:
```ruby
def signup
  @foo = params[:foo]
```
to match the spec?

features/matchers/have_enqueued_mail_matcher.feature
```diff
@@ -38,3 +38,90 @@ Feature: have_enqueued_mail matcher
...
+    When I run `rspec spec/mailers/my_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Parameterize and pass an argument to the mailer
+    Given a file named "app/mailers/my_mailer.rb" with:
+      """ruby
+      class MyMailer < ApplicationMailer
+
+        def signup(user = nil)
+          @user = user
```
pirj, line 106: Subjectively, for me it would be less confusing if it was
```ruby
        def signup(user) # no default value
          @user = user
          @foo = params[:foo]
```

I understand that those examples are not Rails code examples, but still.
What do you think?

features/matchers/have_enqueued_mail_matcher.feature
```diff
@@ -38,3 +38,90 @@ Feature: have_enqueued_mail matcher
...
+    Given a file named "spec/mailers/my_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe MyMailer do
+        it "matches with enqueued mailer" do
+          ActiveJob::Base.queue_adapter = :test
+          # Works also with both, named parameters match first argument
+          expect {
+            MyMailer.with('foo' => 'bar').signup('user').deliver_later
```
pirj, line 121: A failed scenario suggests this should be a symbol.

pirj (approved):
Looks good. Thanks for moving this forward.
A couple of small fixes/clarifications and it's good to go.

features/matchers/have_enqueued_mail_matcher.feature
fabn: Done

pirj:
@mico @dmitry Ping. Does this PR fix your cases?

pirj:
@fabn Thanks a lot, looks flawless to me.

@JonRowe Any objections to merging this?

Haegin:
I've just run in to this when upgrading to Rails 6 myself - is there anything I can do to help get this over the line?

pirj:
Sure, just take a look at the PR, and make it work locally for Rails 5.2, 6.0 and 7.0. You can send a patch, or open a pr basing on what has been done, your choice. That would be very helpful as as  7.0 was just released.

pirj:
@Haegin Please accept my apologies, I replied without looking.
I suggest you to use this branch in your Gemfile while the PR is not merged.

dmitry:
@pirj on the next week going to check this branch in the project.

dmitry:
Just tested and seems it works in the following cases:

```ruby
expect {
          expect {
            request!
          }.to have_enqueued_mail(LeadMailer, :user_created)
                 .with(
                   a_hash_including(
                     params: {
                       user: instance_of(User)
                     }
                   )
                 )
        }.to have_enqueued_mail(DeviseMailer, :confirmation_instructions)
               .with(
                 args: [
                   instance_of(User),
                   instance_of(String),
                   {}
                 ]
               )
```

Now it looks like this:

```ruby
expect {
          expect {
            request!
          }.to have_enqueued_mail(LeadMailer, :user_created)
                 .with(
                   user: instance_of(User)
                 )
        }.to have_enqueued_mail(DeviseMailer, :confirmation_instructions)
               .with(
                 instance_of(User),
                 instance_of(String),
                 {}
               )
```

Means it still works.

pirj:
I had to make [this change](https://github.com/rspec/rspec-rails/pull/2546/commits/b0dc2eae87011e623bb5e771217301d4ea9857f9) to make this change pass on both Rails 7 and Rails 6.0. Kindly appreciate your feedback.

pirj:
Fixed in #2546. Thanks for your contribuion, @fabn !

fabn:
You're welcome

PR #2458, opened 2021-02-14
Stop using --color option

pirj:
Similar to https://github.com/rspec/rspec-support/pull/496

`--color` removed in https://github.com/rspec/rspec-core/pull/2864 (`4.0-dev` branch for upcoming RSpec 4)

PR #2519, opened 2021-09-18
Remove incompatible --color CLI option

pirj:
The option was removed in RSpec 4, see https://github.com/rspec/rspec-core/pull/2864
Follow-up to https://github.com/rspec/rspec-rails/pull/2458


pirj:
Merging, as there's no external behaviour change, it only affects our builds. To my best knowledge, `--color` is a no-op, as `color_mode` is `:automatic` by default ("the output will be in color if the output is a TTY").

PR #2520, opened 2021-09-27
Add a scenario for using subdomain constrained routes

mirkode:
Just some documentation on how to use `host!` in before block for testing subdomain constrained requests.

Credits for this go to @gferrarocamus for pointing this out.

### DIFF
--- a/features/request_specs/request_spec.feature
+++ b/features/request_specs/request_spec.feature
@@ -152,3 +152,23 @@ Feature: request spec
       """
     When I run `rspec spec`
     Then the example should pass
+
+  Scenario: testing subdomain constrained requests
+    Given a file named "spec/requests/widgets_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe "Widget management", :type => :request do
+        before { host! "api.example.com" }
+
+        it "creates a Widget" do
+          headers = { "ACCEPT" => "application/json" }
+          post "/widgets", :params => { :widget => {:name => "My Widget"} }, :headers => headers
+
+          expect(response.content_type).to eq("application/json; charset=utf-8")
+          expect(response).to have_http_status(:created)
+        end
+      end
+      """
+    When I run `rspec spec`
+    Then the example should pass


features/request_specs/request_spec.feature
```diff
@@ -152,3 +152,23 @@ Feature: request spec
...
     When I run `rspec spec`
     Then the example should pass
+
+  Scenario: testing subdomain constrained requests
+    Given a file named "spec/requests/widgets_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe "Widget management", :type => :request do
+        before { host! "api.example.com" }
```
pirj, line 162: This test will pass without `host!`, too, won't it?

What do you think of drawing routes so that it fails without `host!`?
Something like
```ruby
Rails.application.routes.draw do
  resources :widgets, constraints: { subdomain: "api" }
end
```
should do.

features/request_specs/request_spec.feature
```diff
@@ -152,3 +152,23 @@ Feature: request spec
...
+
+        it "creates a Widget" do
+          headers = { "ACCEPT" => "application/json" }
+          post "/widgets", :params => { :widget => {:name => "My Widget"} }, :headers => headers
+
+          expect(response.content_type).to eq("application/json; charset=utf-8")
```
pirj, line 168: As you may have noticed, `content_type` depends on the Rails version. We have different examples, and they are either run or skipped depending on CI job matrix.
I suggest `.to start_with("application/json")` so that this example passes with all Rails versions from our matrix.

pirj (commented):
A couple of notes.
Looks good otherwise, thank you!

features/request_specs/request_spec.feature
mirkode: Hi @pirj ,

Thanks for your feedback. I added the requested changes. 
However the test without the `host!` is still passing, so I wanna look into this again. I will let you know once the PR is ready for review again.
Thanks for taking the time!

JonRowe: I checked on Rails 6 that this fails without `host!`

PR #2115, opened 2019-04-30
Improve path parsing in _default_render_options

jhawthorn:
This improves the path parsing here to be more strict, understand variants, and should support the additional validations around format in Rails 6.0.

I added a few previously failing specs, which this makes pass.

Fixes #2114

cc @samphippen @benoittgt @jaredmoody

PR #2521, opened 2021-09-30
Fix the type of handlers options passed to render

alpaca-tc:
When calling #render without arguments, #_render_options infers default
options. Correctly, the type of handlers should be symbols, but now it is
strings.

In Rails7, if the type of handlers isn't symbols, action_view's template
resolver can't find the template. This is because the implementation of template search has been changed to match handlers exactly.

related changes: https://github.com/rails/rails/pull/42210/files#diff-b356f21d70a4c088415a9c4f315bbccbc979c2ddeda6b8cc5d76770084e5bc6bR31


JonRowe:
:wave: Thanks for this! Our versioning scheme means it'd be a major version bump to support Rails 7, so I might not merge this to `main`. Do you have a scenario that would trigger this? As I'm surprised there was no change required to our scenarios for this.

alpaca-tc:
> so I might not merge this to `main`.

👌 😄 

> Do you have a scenario that would trigger this? As I'm surprised there was no change required to our scenarios for this.

Sorry I can't reproduce this bug in one file. Please check the following 🐕 

```sh
$ rails --version #=> 7.0.0.alpha1
$ rails new test-app --minimal

...and applied changes. 
https://gist.github.com/alpaca-tc/9ce5c09bae477be96534338933627df8
```

I've also created a repository, but I'm going to delete it later.

```sh
$ git clone https://github.com/alpaca-tc/test-app
$ cd test-app
$ bundle install
$ bundle exec rspec
```

pirj:
Probably something fails, but [we've allowed failures for Rails master](https://github.com/rspec/rspec-rails/blob/211d7d990e9762e229d8a86249b88c2a7604e8b0/.github/workflows/ci.yml#L21), and GitHub deceptively shows it as green, while it's actually:
```
NameError:
  uninitialized constant Rack::Request::Env
# ./spec/spec_helper.rb:27:in `<top (required)>'
```

pirj:
The problem with `Rack::Request::Env` is due to this https://github.com/rack/rack/issues/1768 in `rack` `master`.
When we test against Rails ~`master`~ `main`, we also [use the latest `rack` from git](https://github.com/rspec/rspec-rails/blob/211d7d990e9762e229d8a86249b88c2a7604e8b0/Gemfile-rails-dependencies#L11).

CI fix https://github.com/rspec/rspec-rails/pull/2523:
 - 🍏 specs for Rails `main` pass so far
 - 🍏 dependencies for features resolve
 - ❌ sprockets complain about missing [`manifest.js`](https://github.com/rails/sprockets/blob/070fc01947c111d35bb4c836e9bb71962a8e0595/UPGRADING.md#manifestjs)

@alpaca-tc I'm off until Sunday, do you want to take over to push that other PR further?

pirj:
Alright, here's the failing test we were looking for:
```
Failures:

  1) admin/accounts/index renders a list of admin/accounts
     Failure/Error: <%= render @admin_accounts %>

     ActionView::Template::Error:
       Missing partial admin/accounts/_account with {:locale=>[:en], :formats=>[:html, :text, :js, :css, :ics, :csv, :vcf, :vtt, :png, :jpeg, :gif, :bmp, :tiff, :svg, :mpeg, :mp3, :ogg, :m4a, :webm, :mp4, :otf, :ttf, :woff, :woff2, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip, :gzip], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :coffee, :jbuilder]}.

       Searched in:
         * "/home/runner/work/rspec-rails/rspec-rails/tmp/example_app/app/views"
         * "/home/runner/work/rspec-rails/bundle/ruby/2.7.0/bundler/gems/rails-c2b083df913f/actiontext/app/views"
         * "/home/runner/work/rspec-rails/bundle/ruby/2.7.0/bundler/gems/rails-c2b083df913f/actionmailbox/app/views"
     # ./app/views/admin/accounts/index.html.erb:6:in `_app_views_admin_accounts_index_html_erb___1899867805406451531_30760'
     # ./spec/views/admin/accounts/index.html.erb_spec.rb:16:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # ActionView::MissingTemplate:
     #   Missing partial admin/accounts/_account with {:locale=>[:en], :formats=>[:html, :text, :js, :css, :ics, :csv, :vcf, :vtt, :png, :jpeg, :gif, :bmp, :tiff, :svg, :mpeg, :mp3, :ogg, :m4a, :webm, :mp4, :otf, :ttf, :woff, :woff2, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip, :gzip], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :coffee, :jbuilder]}.
     #   
     #   Searched in:
```

@alpaca-tc I suggest you to rebase on top of `fix-rails-7-build` branch and see if your changes fixed the failure.

pirj:
Sadly enough, this fix doesn't fix the failing spec, something seems to still be missing.
Should we `.to_sym` `match[:locale]`, too?

I've rebased and force-pushed this branch.

Can you try running `cucumber` locally, @alpaca-tc ?
To run against Rails' `main`:
```
rm Gemfile.lock
export RAILS_VERSION='master'
bundle
rm -rf tmp
cucumber
```

alpaca-tc:
I see, it is a cucumber scenario 🥲 

---

It seems that the content of `tmp/example_app/app/views/admin/accounts/index.html.erb` created by rails7 is different from that of the template created by rails6.
Moreover, the template created by rails7 contains the bug. `render @admin_accounts` doesn't work in every rails versions.

I am not familiar with the behavior of cucumber, how can I fix this?

```erb
# generated by rail6
<p id="notice"><%= notice %></p>

<h1>Admin Accounts</h1>

<table>
  <thead>
    <tr>
      <th>Name</th>
      <th colspan="3"></th>
    </tr>
  </thead>

  <tbody>
    <% @admin_accounts.each do |admin_account| %>
      <tr>
        <td><%= admin_account.name %></td>
        <td><%= link_to 'Show', admin_account %></td>
        <td><%= link_to 'Edit', edit_admin_account_path(admin_account) %></td>
        <td><%= link_to 'Destroy', admin_account, method: :delete, data: { confirm: 'Are you sure?' } %></td>
      </tr>
    <% end %>
  </tbody>
</table>

<br>

<%= link_to 'New Admin Account', new_admin_account_path %>
```

```erb
# generated by rail7
<p id="notice"><%= notice %></p>

<h1>Accounts</h1>

<div id="admin_accounts">
  <%= render @admin_accounts %>
</div>

<%= link_to "New account", new_admin_account_path %>
```

pirj:
Your Rails bugfix PR is merged, but there's now an inconsitency due to the recently removed `ActionMailer::DeliveryJob` #2531.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -151,19 +151,19 @@ def _default_file_to_render; end # Stub method
...
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], formats: [:html], handlers: ['erb'], variants: ['fancy']}, {}, nil])
+          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: [:en], formats: [:html], handlers: [:erb], variants: [:fancy]}, {}, nil])
         end
 
         it "converts the filename without format into render options" do
           allow(view_spec).to receive(:_default_file_to_render) { "widgets/new.en.erb" }
           view_spec.render
-          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: ['en'], handlers: ['erb']}, {}, nil])
+          expect(view_spec.received.first).to eq([{template: "widgets/new", locales: [:en], handlers: [:erb]}, {}, nil])
```
pirj, line 166: I would feel confident with merging this to `main` if we had a confirmation that with `to_sym` all supported Rails versions (from soft-supported 5.0 and 5.1) up to 6.1 still work fine.

Can you recall, what was the failing cucumber scenario for 7.0 without `.to_sym` exactly?

alpaca-tc: I can't recall but I think that calling `render` without arguments raises `ActionView::MissingTemplate`.

alpaca-tc: It seems that an exception is raised if the path described in the description contains handler, locales, etc. as shown below. Sorry I don't know that there is such a test in rspec-rails' cucumber.

```ruby
RSpec.describe "root/index.html.erb", type: :view do # also xxx.html.slim, xxx.en.html.erb 
  it { render } #=> raise ActionView::MissingTemplate
end
```

pirj: My concern is that the change would break on an older version of Rails. As this is something that was not thought forward to be changed in the future, there might not be a test for it. Sorry for being extra cautious.

jcoyne: @pirj In that case would it be possible to release this as a new major version

pirj: Only if we drop support for all versions that have a chance not to support this. In `rspec-rails` 6 we will most probably still keep support for Rails 6.0, and this means we have to make sure the approach with symbols works for it, too.

This is purely theoretical. I have a gut feeling that with Symbols it will work with 5.2 and probably even [earlier versions](https://github.com/rspec/rspec-rails/pull/2115/files#diff-104fd0e14717644c4d9887ea03b002a615a7f41020003a78ab6992fa425e7beeR131), too. But to be on the same side, I'd love some confirmation that `requested.handlers` in Rails 6.0 would accept a Symbol.
Otherwise, we'll have to add a conditional Rails version check in the runtime and [specs](https://github.com/rspec/rspec-rails/pull/2115/files#diff-a5c90dfedaef6ce2253b56400927211db833791f14651c2aab7b7a33df40cb15R140).

Frankly speaking, I don't have a clear plan how to proceed with this. @JonRowe ?
#2537 is standing in our way to get a green Rails 7 build. I hope to get to it next week.

alpaca-tc:
Some commits(https://github.com/rspec/rspec-rails/pull/2521/commits/2a2c0650dbf2dcb1af67eaf5bea4d6974b275911 https://github.com/rspec/rspec-rails/pull/2521/commits/0f133af725ccae0dfaac124565f0799679232c8b) are conflicted with #2552 , so I removed the commits and force-pushed.

spec/rspec/rails/example/view_example_group_spec.rb
petergoldstein: @pirj So I just saw this discussion after making a similar commit to the PR I opened a few days ago.  I'm a little confused as to what the concern is.

We know symbols work on Rails versions in CI because the relevant Cucumber feature (`features/view_specs/view_spec.feature:124`) passes for those versions.  If symbols didn't work on a particular Rails version that feature would fail.

pirj: > features/view_specs/view_spec.feature:124

Fair enough. For some reason, I failed to find this feature.

Wondering if it works equally fine with `variant` and `locale`.

For the reference, a [discussion about `format.to_sym`](https://github.com/rspec/rspec-rails/pull/1452).
This discussion has an argument "if they don't support symbols, it's a Rails bug, and people using older versions of Rails, they should stick to older versions of `rspec-rails`". We've started adhering to semver since then, and currently we officially support Rails 5.2 and soft-support 5.0 and 5.1 (please correct if I'm wrong with this again). From my perspective, it would be quite irresponsible to break the current `rspec-rails` 5.0.x for users of Rails 5.
Even though we claim that:
> According to RSpec Rails new versioning strategy use:
>    rspec-rails 5.x for Rails 6.x.
>    rspec-rails 4.x for Rails from 5.x or 6.x.

and 
> ### 5.0.0 / 2021-03-09
> * Drop support for Rails below 5.2.

We have a release strategy @JonRowe has described [here](https://github.com/rspec/rspec-rails/pull/2546#issuecomment-1009785284) that I've recently violated with the merge of that PR, hopefully without any breaking impact for the users.
For the future researchers, it might be surprising that Rails 7.0 compatibility support appears in `rspec-rails` 5.0.3/5.1.0 releases that target Rails 6.0 and 6.1.

I suggest branching `rails-7-0-dev` from `main`, and merging this PR and #2552 to it.
When we release 5.0.3 and 5.1.0, it can be merged back to `main`, and it will be time for `rspec-rails` 6.0.

@JonRowe Does this make sense?

petergoldstein: For what it's worth, that approach sounds reasonable to me.

That said I'm not sure why there's a "soft support" for Rails 5 and 5.1.  If you want to support them, why not just add them to CI?  If they can't be added to CI, are they actually supported?

It's worth noting that PR #2552 + #2546 has only 2 code changes of very limited scope that could end users of spec-rails:

1. The symbol change for the handler argument ("erb" becomes :erb before being passed to ActionView::Base).  This only becomes relevant in a particular type of view spec.
2. The changes to argument handling for Action Mailer matches (in my view this is the far more likely to break item between Rails versions)

On top of these changes, this PR symbolizes a few other values passed to ActionView::Base.

That's a very small, tightly scoped set of changes.  So the potential blast radius and risk of impact on end users is very limited.

pirj: > if you want to support them, why not just add them to CI

We don't want to support them, and we don't want to break them. The reason is to drop some burden, mostly branching in specs:
 - [drop support for Rails < 5.2](https://github.com/tagliala/rspec-rails/commit/686bf845ad17704b757aafdd8388058563546b0c)
 - [drop support for Rails < 5.0](https://github.com/rspec/rspec-rails/pull/2226)

Actually, by looking at the former, there are [two changes](https://github.com/tagliala/rspec-rails/commit/686bf845ad17704b757aafdd8388058563546b0c#diff-cf41d523f2c21002e68799050ceb4c8e781e1d291680537457a3bebdf801161bL10) that presumably break Rails 5.0 support. But 5.1 may still work.

>  PR #2552 + #2546 has only 2 code changes of very limited scope

Right 👍
Please don't get me wrong with my rant above, I'm all for merging both ASAP and releasing this as 6.0.0-pre without bothering with 5.0.3 and 5.1.0.

petergoldstein: I guess I just consider "soft support" very harmful because it adds ambiguity and discussion to almost any change, without any kind of authoritative answer.  If it's in CI, then you know you didn't break it (or you broke it and need to fix it).  Saving a few branches is in my view not worth the ambiguity on nearly every material PR.  In my experience it sows up work (and open source work) substantially.

But I'm not a maintainer here, so I'll of course defer to the library maintainers.

pirj:
Thank you, @alpaca-tc !

PR #2523, opened 2021-10-01
Fix Rails 7 build

pirj:
This change:

1. Removes explicit dependencies on gems independent, but required by Rails (`i18n`, `rack`)
2. Removes explicit dependencies on Rails-swallowed gems (`journey`, `arel`)
3. Minor fixes to `tmp/example_app/Gemfile` generation to avoid conflicts with `Gemfile-rails-dependencies`
4. Added sprockets manifest generation

These changes only affect our build, and only against Rails' `main`.
See https://github.com/rspec/rspec-rails/pull/2521#issuecomment-932412247 and https://github.com/rspec/rspec-rails/pull/2521/checks?check_run_id=3757830550

### `rack`

Using rack from git opens up opportunities to fail due to unreleased rack issues, e.g. https://github.com/rack/rack/issues/1768

No need to explicitly depend on rack, Rails does that: https://github.com/rails/rails/blob/c2b083df913f44e664576aafe24c41ec657f8eb5/actionpack/actionpack.gemspec#L37

### `i18n`

Depending on i18n's latest git version opens up opportunities to:
1. fail due to i18n's unstable changes
2. fail due to version conflict (their `master` can flip to 2.0, while `rails` would depend on `< 2`)

Rails explicitly depend on `i18n` anyway: https://github.com/rails/rails/blob/c2b083df913f44e664576aafe24c41ec657f8eb5/activesupport/activesupport.gemspec#L36

### `journey`

https://github.com/rails/journey

> This gem was merged on Rails 4.0 and will only receive security fixes

### `arel`

https://github.com/rails/arel

Arel was swallowed by Rails back in 2018

### DIFF
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -6,7 +6,6 @@ when /master/
   gem "activerecord-deprecated_finders", :git => "https://github.com/rails/activerecord-deprecated_finders.git"
   gem "web-console", :git => "https://github.com/rails/web-console", :group => :development
   gem 'coffee-rails', :git => "https://github.com/rails/coffee-rails.git"
-  gem 'i18n', :git => 'https://github.com/svenfuchs/i18n.git', :branch => 'master'
   gem 'sprockets', :git => 'https://github.com/rails/sprockets.git', :branch => 'master'
   gem 'sprockets-rails', :git => 'https://github.com/rails/sprockets-rails.git', :branch => 'master'
   gem 'puma', "3.12.1"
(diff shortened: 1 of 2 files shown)


JonRowe:
Removing mainline dependencies on `arel` and `journey` is fine, as the Rails mainline has since swallowed them; `rack` and `i18n` is a bit more hand-wavey, they are there as previous mainline versions of `rails` have relied on unreleased versions, we can remove them for now but we might have to add them back in future. Feel free to merge with that in mind.

JonRowe (approved):

PR #2529, opened 2021-11-16
Fix type of be_routable scenario in controllers

eduardoj:


pirj:
Failures don't seem to be related - apparently something in Cucumber 7.1.0 breaks our features.
I'll look into this later.

pirj:
Thank you for the contribution!

Issue #2532, opened 2021-11-18
Parameter name is not equal the model name when executing `scaffold_controller` with namespace


kenzo-tanaka:
Hi👋
First of all thank you for the very helpful gem!

## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.6.3p62
Rails version: 6.1.4.1
RSpec core version: 3.10.1
RSpec rails version: 5.0.2

## Observed behaviour

When the request spec is created by executing the `scaffold_controller` command with namespace, the parameter name is not equal the model name.

Here is a pull request that reproduces the behaviour
https://github.com/kenzo-tanaka/rails_sandbox/pull/14

```shell
$ rails g model Product name:string price:integer
$ rails db:migrate
$ rails g scaffold_controller admin/products --model-name=product --no-helper --no-assets --no-controller-specs --no-routing-specs --no-view-specs
```

```rb
describe "POST /create" do
  context "with valid parameters" do
    it "creates a new Product" do
      expect {
        # Here, I expected `{ product: valid_attributes }` like the test of update action
        post admin_products_url, params: { admin_product: valid_attributes } 
      }.to change(Product, :count).by(1)
    end
    # ...
end
```

## Expected behaviour

I expected that the autogenerated code is like this:

```rb
describe "POST /create" do
  context "with valid parameters" do
    it "creates a new Product" do
      expect {
        post admin_products_url, params: { product: valid_attributes } 
      }.to change(Product, :count).by(1)
    end
    # ...
end
```

## Can you provide an example app?

https://github.com/kenzo-tanaka/rails_sandbox/pull/14/files

---

I'm sorry if I misunderstood something about this behaviour🙏

pirj:
Seems related to #2491 and #2492. @PedroAugustoRamalhoDuarte What do you think?

PedroAugustoRamalhoDuarte:
Yes, its wrong. In [create action](https://github.com/rspec/rspec-rails/blob/d3e7b85877fcbcec63f8a76434d8750e7f3b7aef/lib/generators/rspec/scaffold/templates/request_spec.rb#L68) uses ns_file_name method and in [update_action](https://github.com/rspec/rspec-rails/blob/d3e7b85877fcbcec63f8a76434d8750e7f3b7aef/lib/generators/rspec/scaffold/templates/request_spec.rb#L109) uses file_name method.

In my pull request i have changed only model name and I hadn't seen this inconsistency in params

[...]

PR #2534, opened 2021-11-19
Fix scaffold template with namespace and `--model-name` option

kenzo-tanaka:
close https://github.com/rspec/rspec-rails/issues/2532

Fixed the behavior of scaffold template when namespace and `--model-name` option passed.

```shell
$ rails g scaffold_controller admin/products --model-name=product
```

```rb
describe "POST /create" do
  context "with valid parameters" do
    it "creates a new Product" do
      expect {
        # Here, I expected `{ product: valid_attributes }` like the test of update action
        post admin_products_url, params: { admin_product: valid_attributes } 
      }.to change(Product, :count).by(1)
    end
    # ...
end
```

This is because `ns_file_name` method that is called in scaffold template returns `admin_product` when namespaced.

https://github.com/rspec/rspec-rails/blob/ba2d66abf275fe1f490c3ed219459e42698936d6/lib/generators/rspec/scaffold/scaffold_generator.rb#L79-L83

So, I replaced `ns_file_name` to `singular_table_name` in scaffold templates.

Also, I added some tests to ensure that templates works in the old way when the `--model-name` option is not passed.

### DIFF
--- a/lib/generators/rspec/scaffold/templates/api_controller_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/api_controller_spec.rb
@@ -64,21 +64,21 @@ RSpec.describe <%= controller_class_name %>Controller, <%= type_metatag(:control
     context "with valid params" do
       it "creates a new <%= class_name %>" do
         expect {
-          post :create, params: {<%= ns_file_name %>: valid_attributes}, session: valid_session
+          post :create, params: {<%= singular_table_name %>: valid_attributes}, session: valid_session
         }.to change(<%= class_name %>, :count).by(1)
       end
 
-      it "renders a JSON response with the new <%= ns_file_name %>" do
-        post :create, params: {<%= ns_file_name %>: valid_attributes}, session: valid_session
+      it "renders a JSON response with the new <%= singular_table_name %>" do
+        post :create, params: {<%= singular_table_name %>: valid_attributes}, session: valid_session
         expect(response).to have_http_status(:created)
         expect(response.content_type).to eq('application/json')
-        expect(response.location).to eq(<%= ns_file_name %>_url(<%= class_name %>.last))
+        expect(response.location).to eq(<%= singular_table_name %>_url(<%= class_name %>.last))
       end
     end
 
     context "with invalid params" do
-      it "renders a JSON response with errors for the new <%= ns_file_name %>" do
-        post :create, params: {<%= ns_file_name %>: invalid_attributes}, session: valid_session
+      it "renders a JSON response with errors for the new <%= singular_table_name %>" do
+        post :create, params: {<%= singular_table_name %>: invalid_attributes}, session: valid_session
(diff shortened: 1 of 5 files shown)


pirj:
Nice, thank you!

What happens if you don't specify ` --model-name=card` in `rails g scaffold_controller admin/cards`, will it behave in the old way, will it use `admin_cards` (`admin_products` from the ticket's example)?

In case no specs fail, would you please add some examples to prevent from regression to `spec/generators/rspec/scaffold/scaffold_generator_spec.rb`?

kenzo-tanaka:
@pirj Thank you for quick response!

> What happens if you don't specify --model-name=card in rails g scaffold_controller admin/cards, will it behave in the old way, will it use admin_cards (admin_products from the ticket's example)?

Thanks for pointing that out. If I remove `--model-name=card`, it behaved in the old way like below:

```rb
describe "POST /create" do
  context "with valid parameters" do
    it "creates a new Admin::Card" do
      expect {
        post admin_cards_url, params: { admin_card: valid_attributes }
      }.to change(Admin::Card, :count).by(1)
    end
   # ...
  end
end
```

Is this the behavior that you expected?

---

> In case no specs fail, would you please add some examples to prevent from regression to spec/generators/rspec/scaffold/scaffold_generator_spec.rb?

OK! I'll try it.

pirj:
> Is this the behavior that you expected?

Yes, perfect. Thank you!

pirj (approved):
Perfect, thank you!

kenzo-tanaka:
@pirj 
When I run `rails g scaffold_controller admin/cards --model-name=card  --controller_specs` in a example app, the controller spec may have a same situation of this issue.

```rb
describe "POST #create" do
  context "with valid params" do
    it "creates a new Card" do
      expect {
        post :create, params: {admin_card: valid_attributes}, session: valid_session
      }.to change(Card, :count).by(1)
    end
    # ...
  end
end
```

Should this file be fixed as well?

https://github.com/rspec/rspec-rails/blob/d3e7b85877fcbcec63f8a76434d8750e7f3b7aef/lib/generators/rspec/scaffold/templates/controller_spec.rb#L82

pirj:
It would be great if you're up for it.

PedroAugustoRamalhoDuarte:
We have to change templates/api_request_spec.rb too, there is the same problem.

kenzo-tanaka:
@pirj @PedroAugustoRamalhoDuarte 
Finally, I fixed the following files because they had the same problem.

- `templates/request_spec.rb`
- `templates/api_request_spec.rb`
- `templates/controller_spec.rb`
- `templates/api_controller_spec.rb`

And also, I added some examples to prevent from regression. Please check🙏

pirj (approved):
Great job! Thanks a lot for taking care of the whole thing.

JonRowe:
Thanks for tackling this

PR #2535, opened 2021-11-21
Fix Rails integration branch name

pirj:
I'm not aware of the magic how `bundle install` worked when we pointed to `rails`' `master` branch: `git fetch origin master` in Rails source fails.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,15 +16,15 @@ jobs:
       fail-fast: false
       matrix:
        include:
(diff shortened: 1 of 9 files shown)


pirj:
@JonRowe The fix from `master` to `main` triggered TravisCI JRuby head build jobs. What is our plan for JRuby, remove it from the build matrix?

JonRowe:
They must have changed their `main` branch name finally :joy: Mostly I had no plan, I'm not actively working on replacing them but we do support JRuby

PR #2543, opened 2021-12-14
Changed message in have_broadcasted_to test

Lykos:
Previously it was "successfully subscribes" which has very little to do with this particular test. Probably it's an artifact from copying the more general action cable tests somewhere. I found this because it shows up on documentation when googling the have_broascasted_for matcher. The message I chose is consistent with the others in this file: all existing similar tests have this message.


pirj:
@palkan Would you agree with the change?

palkan:
Yeah, LGTM

PR #2537, opened 2021-11-23
Add compatibility for have_enqueued_mail (rails 7)

mhenrixon:
1. ActionMailer::DeliveryJob does not exist anymore
2. ActionMailer::Parameterized::DeliveryJob

According to my research they have both been superseeded by ActionMailer::MailDeliveryJob

Closes #2531

PR #2547, opened 2021-12-30
Stop using the deceiving continue-on-error

pirj:
Build jobs appear as green even though they actually fail, e.g. https://github.com/rspec/rspec-rails/pull/2537#issuecomment-977641695

A better approach is to mark jobs as "required" in branch protection.
image

This way, failed jobs are red, but this doesn't prevent the PR from being merged upon at maintainers' discretion.
image

@JonRowe Can you please set up branch protection so it has 6.1 jobs and below as "Required"?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,25 +18,20 @@ jobs:
        include:
          # Edge Rails (7.1) builds >= 2.7
          - ruby: '3.0'
-           allow_failure: true
            env:
              RAILS_VERSION: 'main'
          - ruby: 2.7
-           allow_failure: true
            env:
              RAILS_VERSION: 'main'
 
          # Rails 7.0 builds >= 2.7
          - ruby: 3.1
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 7.0.0'
          - ruby: '3.0'
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 7.0.0'
          - ruby: 2.7
-           allow_failure: true
            env:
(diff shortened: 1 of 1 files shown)


pirj:
Green for Rails 6.1 and below, red for Rails 7.0 and edge, as expected.

JonRowe:
The problem with this, and we making status checks required, is it prevents us from having an "allow failures" mode, which as you can see by the red builds, we need for edge rails and currently unsupported rails versions

pirj:
> setting the builds to required, would actually prevent this PR from being merged for example.

Yes. I suggest marking those jobs that *we know* should pass, e.g. currently Rails 6.1 on Ruby 2.7.
Others, like Rails 7 on Ruby 3.1 should not be marked as such until we make sure it's stable.

In any case, you can still merge a PR with a red "required" build job, but GH will ask for an additional confirmation:
image

A benefit, as I see it, is that non-required builds are shown as red.
This may save us some trouble with regressions in Rails 7 support when we get to it. With the current approach, the build will appear as "all green", even if the PR broke Rails 7 support that was previously green.
An easy way to check this it to unfold jobs with "Show all checks", and check build times for suspiciously low "Successful in 2m" where the baseline is ~15m.
image

JonRowe:
> A benefit, as I see it, is that non-required builds are shown as red.

Thats not a benefit to me, we want builds to be green if required builds pass, github doesn't give us the option of having optional failures without this, those interested in non-required builds (mostly us as we develop compatibility for things) can investigate individual builds when we care, it would be best if @Github allowed us to mark non required builds but thats not a feature they've built yet.

PR #2548, opened 2021-12-30
Add Rails 7.0 and Ruby 3.1 to build matrix

pirj:
https://www.ruby-lang.org/en/news/2021/12/25/ruby-3-1-0-released/
https://rubyonrails.org/2021/12/15/Rails-7-fulfilling-a-vision

PR #2549, opened 2021-12-30
Get RubyGems/Bundler update CI step back

pirj:
https://github.com/rubygems/bundler/pull/6728

Example failure (https://github.com/rspec/rspec-rails/pull/2548):
```
Failure/Error: expect(`bundle exec #{script} 2>&1`).to be_empty
       expected `"Your RubyGems version (2.7.6.3)) has a bug that prevents `required_ruby_version` from working for Bundler. Any scripts that use `gem install bundler` will break as soon as Bundler drops support for your Ruby version. Please upgrade RubyGems to avoid future breakage and silence this warning by running `gem update --system 3.2.3`\n".empty?` to be truthy, got false
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby }}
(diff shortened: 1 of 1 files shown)


pirj:
image

PR #2546, opened 2021-12-20
Fix have_enqueued_mail for Rails 7 and argument matching

pirj:
This is a combo of #2516 and #2537, supposed to fix #2351 and #2531 with a slight fix for argument matching and a small fix for our system spec spec on Rails 7.

PR #2552, opened 2022-01-09
Add Ruby 3.1 to CI and fix Rails 7

petergoldstein:
The bulk of the changes in this PR are test changes.  These include:

1. Updating the example app configuration for Rails 7, including disabling eager class loading on CI
2. Updating specs to filter out additional lines that may be generated when running commands
3. Removing ".html.erb" and ".xml.erb" suffixes in render calls
4. Updating specs to accomodate differences in Rails view scaffolding before and after Rails 7

Material code changes include:

1. Adding additional logic to the ActionMailer argument parsing to accomodate for differences under Ruby 3.1/Rails 6.1
2. Symbolizing the handler argument parsed from the spec description for view specs with an empty render

With the application of these changes to main all CI entries (including the new Ruby 3.1 entries and existing Rails 7 entries) run green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,10 @@ jobs:
       matrix:
        include:
          # Edge Rails (7.1) builds >= 2.7
+         - ruby: 3.1
+           allow_failure: true
+           env:
+             RAILS_VERSION: 'main'
          - ruby: '3.0'
            allow_failure: true
            env:
@@ -41,19 +45,19 @@ jobs:
 
          # Rails 7.0 builds >= 2.7
          - ruby: 3.1
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 7.0.0'
          - ruby: '3.0'
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 7.0.0'
          - ruby: 2.7
-           allow_failure: true
            env:
              RAILS_VERSION: '~> 7.0.0'
 
          # Rails 6.1 builds >= 2.5
+         - ruby: 3.1
+           env:
+             RAILS_VERSION: '~> 6.1.0'
          - ruby: '3.0'
            env:
              RAILS_VERSION: '~> 6.1.0'
--- a/example_app_generator/generate_action_mailer_specs.rb
+++ b/example_app_generator/generate_action_mailer_specs.rb
@@ -31,6 +31,7 @@ using_source_path(File.expand_path(__dir__)) do
   if skip_active_record?
     comment_lines 'spec/support/default_preview_path', /active_record/
     comment_lines 'spec/support/default_preview_path', /active_storage/
+    comment_lines 'spec/support/default_preview_path', /action_mailbox/
   end
   copy_file 'spec/verify_mailer_preview_path_spec.rb'
 end
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -167,4 +167,10 @@ gsub_file 'spec/requests/gadgets_spec.rb',
 gsub_file 'spec/controllers/uploads_controller_spec.rb',
           'skip("Add a hash of attributes valid for your model")',
           '{}'
+
+if Rails.version >= '7.0.0'
+  # Some gems (ActionMailBox, ActionCable, etc.) are not used when running `example_app_generator/spec/verify_mailer_preview_path_spec.rb`, so `eager_load` must be false.
+  gsub_file "config/environments/test.rb", 'ENV["CI"].present?', "false"
+end
+
 final_tasks
--- a/example_app_generator/generate_stuff.rb
+++ b/example_app_generator/generate_stuff.rb
@@ -168,9 +168,4 @@ gsub_file 'spec/controllers/uploads_controller_spec.rb',
           'skip("Add a hash of attributes valid for your model")',
           '{}'
 
-if Rails.version >= '7.0.0'
-  # Some gems (ActionMailBox, ActionCable, etc.) are not used when running `example_app_generator/spec/verify_mailer_preview_path_spec.rb`, so `eager_load` must be false.
-  gsub_file "config/environments/test.rb", 'ENV["CI"].present?', "false"
-end
-
 final_tasks
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -28,7 +28,7 @@ require_file_stub 'config/environment' do
     require "action_view/railtie"
     if Rails::VERSION::STRING >= '6'
       require "action_cable/engine"
-      require "action_mailbox/engine"
+      require "active_job/railtie"
     end
 
     # Require the gems listed in Gemfile, including any gems
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -26,6 +26,10 @@ require_file_stub 'config/environment' do
     require "action_controller/railtie"
     require "action_mailer/railtie" unless ENV['NO_ACTION_MAILER']
     require "action_view/railtie"
+    if Rails::VERSION::STRING >= '6'
+      require "action_cable/engine"
+      require "action_mailbox/engine"
+    end
 
     # Require the gems listed in Gemfile, including any gems
     # you've limited to :test, :development, or :production.
@@ -44,6 +48,8 @@ require_file_stub 'config/environment' do
         if ENV['SHOW_PREVIEWS']
           config.action_mailer.show_previews = (ENV['SHOW_PREVIEWS'] == 'true')
         end
+
+        config.active_record.legacy_connection_handling = false if Rails::VERSION::STRING >= '7'
       end
     end
 
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -29,6 +29,7 @@ require_file_stub 'config/environment' do
     if Rails::VERSION::STRING >= '6'
       require "action_cable/engine"
       require "active_job/railtie"
+      require "action_mailbox/engine"
     end
 
     # Require the gems listed in Gemfile, including any gems
--- a/example_app_generator/spec/verify_mailer_preview_path_spec.rb
+++ b/example_app_generator/spec/verify_mailer_preview_path_spec.rb
@@ -22,6 +22,9 @@ RSpec.describe 'Action Mailer railtie hook' do
     out =  io.readlines
               .reject { |line| line =~ /warning: circular argument reference/ }
               .reject { |line| line =~ /Gem::Specification#rubyforge_project=/ }
+              .reject { |line| line =~ /DEPRECATION WARNING/ }
+              .reject { |line| line =~ /warning: previous/ }
+              .reject { |line| line =~ /warning: already/ }
               .join
               .chomp
     CaptureExec.new(out, $?.exitstatus)
--- a/example_app_generator/spec/verify_mailer_preview_path_spec.rb
+++ b/example_app_generator/spec/verify_mailer_preview_path_spec.rb
@@ -111,6 +111,8 @@ RSpec.describe 'Action Mailer railtie hook' do
       end
 
       it 'handles action mailer not being available' do
+        skip "Rails 7 forces eager loading on CI, loads app/mailers and fails" if Rails::VERSION::STRING.to_f >= 7.0
+
         expect(
           capture_exec(
             custom_env.merge('NO_ACTION_MAILER' => 'true'),
--- a/lib/generators/rspec/scaffold/templates/index_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/index_spec.rb
@@ -18,8 +18,9 @@ RSpec.describe "<%= ns_table_name %>/index", <%= type_metatag(:view) %> do
 
   it "renders a list of <%= ns_table_name %>" do
     render
+    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
 <% for attribute in output_attributes -%>
-    assert_select "tr>td", text: <%= value_for(attribute) %>.to_s, count: 2
+    assert_select cell_selector, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
 <% end -%>
   end
 end
--- a/lib/rspec/rails/feature_check.rb
+++ b/lib/rspec/rails/feature_check.rb
@@ -43,8 +43,11 @@ module RSpec
         defined?(::ActionMailbox)
       end
 
-      def ruby_3_1?
-        RUBY_VERSION >= "3.1"
+      # TODO: add a self-explanatory method name. See https://github.com/rspec/rspec-rails/pull/2552#issuecomment-1009508693 for hints
+      def rails_6_1_and_ruby_3_1?
+        return false if RUBY_VERSION < "3.1"
+
+        ::Rails::VERSION::STRING >= "6.1" && ::Rails::VERSION::STRING < "7"
       end
 
       def type_metatag(type)
--- a/spec/generators/rspec/scaffold/scaffold_generator_spec.rb
+++ b/spec/generators/rspec/scaffold/scaffold_generator_spec.rb
@@ -208,16 +208,16 @@ RSpec.describe Rspec::Generators::ScaffoldGenerator, type: :generator do
       before { run_generator %w[posts upvotes:integer downvotes:integer] }
       subject { file("spec/views/posts/index.html.erb_spec.rb") }
       it { is_expected.to exist }
-      it { is_expected.to contain('assert_select "tr>td", text: 2.to_s, count: 2') }
-      it { is_expected.to contain('assert_select "tr>td", text: 3.to_s, count: 2') }
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(2.to_s), count: 2') }
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(3.to_s), count: 2') }
     end
 
     describe 'with multiple float attributes index' do
       before { run_generator %w[posts upvotes:float downvotes:float] }
       subject { file("spec/views/posts/index.html.erb_spec.rb") }
       it { is_expected.to exist }
-      it { is_expected.to contain('assert_select "tr>td", text: 2.5.to_s, count: 2') }
-      it { is_expected.to contain('assert_select "tr>td", text: 3.5.to_s, count: 2') }
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(2.5.to_s), count: 2') }
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(3.5.to_s), count: 2') }
     end
 
     describe 'with reference attribute' do
--- a/Rakefile
+++ b/Rakefile
@@ -160,7 +160,7 @@ namespace :no_active_record do
         sh "rm -f #{bindir}/rails"
         sh "bundle exec rails new #{example_app_dir} --no-rc --skip-active-record --skip-javascript --skip-bootsnap " \
            "--skip-sprockets --skip-git --skip-test-unit --skip-listen --skip-bundle --skip-spring " \
-           "--template=example_app_generator/generate_app.rb"
+           "--skip-action-text --template=example_app_generator/generate_app.rb"
 
         in_example_app(app_dir: example_app_dir) do
           sh "./ci_retry_bundle_install.sh 2>&1"
--- a/example_app_generator/no_active_record/app/models/in_memory/model.rb
+++ b/example_app_generator/no_active_record/app/models/in_memory/model.rb
@@ -1,5 +1,7 @@
 raise "ActiveRecord is defined but should not be!" if defined?(::ActiveRecord)
 
+require 'active_model'
+
 module InMemory
   module Persistence
     def all
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -102,9 +102,17 @@ module RSpec
 
         def use_given_mail_args?(job)
           return true if FeatureCheck.has_action_mailer_parameterized? && job[:job] <= ActionMailer::Parameterized::DeliveryJob
-          return false if FeatureCheck.ruby_3_1?
+          return false if rails_6_1_and_ruby_3_1?
 
           !(FeatureCheck.has_action_mailer_unified_delivery? && job[:job] <= ActionMailer::MailDeliveryJob)
+        end
+
+        # TODO: move to FeatureCheck
+        def rails_6_1_and_ruby_3_1?
+          return false unless RUBY_VERSION >= "3.1"
+          return false unless ::Rails::VERSION::STRING >= '6.1'
+
+          ::Rails::VERSION::STRING < '7'
         end
 
         def base_mailer_args
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -102,17 +102,9 @@ module RSpec
 
         def use_given_mail_args?(job)
           return true if FeatureCheck.has_action_mailer_parameterized? && job[:job] <= ActionMailer::Parameterized::DeliveryJob
-          return false if rails_6_1_and_ruby_3_1?
+          return false if FeatureCheck.rails_6_1_and_ruby_3_1?
 
           !(FeatureCheck.has_action_mailer_unified_delivery? && job[:job] <= ActionMailer::MailDeliveryJob)
-        end
-
-        # TODO: move to FeatureCheck
-        def rails_6_1_and_ruby_3_1?
-          return false unless RUBY_VERSION >= "3.1"
-          return false unless ::Rails::VERSION::STRING >= '6.1'
-
-          ::Rails::VERSION::STRING < '7'
         end
 
         def base_mailer_args


.github/workflows/ci.yml
```diff
@@ -41,6 +45,9 @@ jobs:
              RAILS_VERSION: '~> 7.0.0'
 
          # Rails 6.1 builds >= 2.5
```
pirj, line 47: I'd suggest removing `allow_failures` for Rails 7.0/Ruby 3.1 job:
```
         # Rails 7.0 builds >= 2.7
         - ruby: 3.1
           allow_failure: true
```

But this is unlikely to go green until #2546 is merged

petergoldstein:
@pirj Agreed.  I'm trying to work that into a combined branch and deal with the remaining failures.

pirj:
I'll go ahead and merge #2546 as soon as the build is green, so you could rebase and work on feature failures.

pirj:
I've rebased this PR on top of `main`. And dared to replace `'3.1'` with just `3.1` in `ci.yml`. I think the problem is just with `3.0` being interpreted as `3` and later on taken as "latest 3.x" by `ruby-setup`.

petergoldstein:
3.1 without quotes is fine.  Yes, that's the issue for the use of an unquoted 3.0 - it rounds to 3 and gets interpreted as latest 3.

As you noted our force-pushes collided.

pirj:
> our force-pushes collided

🤜 🤛

Unfortunately, `--force-with-lease` doesn't work for contributor's repos (when I push to PRs).

pirj:
Cool, now failures in this PR are indicative of what are the incompatibilities with Rails 7 are, not some random `net-` issues and RuboCop parser version mismatches 🎉

petergoldstein:
I've got a handle on the first few failures. The default ERB template changed - https://github.com/rails/rails/commit/164c2f625742da95407c44c70a381e2c80e4e118#diff-f00cb6b8a5dac4a72299b7c0fc2962b5fc554a44e5452a8c0a7fb47b97080ad2 - and a table layout is no longer being used, meaning the matchers are failing.  Working on updating the specs now.

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
```diff
@@ -208,16 +208,16 @@
       before { run_generator %w[posts upvotes:integer downvotes:integer] }
       subject { file("spec/views/posts/index.html.erb_spec.rb") }
       it { is_expected.to exist }
-      it { is_expected.to contain('assert_select "tr>td", text: 2.to_s, count: 2') }
-      it { is_expected.to contain('assert_select "tr>td", text: 3.to_s, count: 2') }
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(2.to_s), count: 2') }
```
pirj, line 211: ❤️

lib/generators/rspec/scaffold/templates/index_spec.rb
```diff
@@ -18,8 +18,9 @@
 
   it "renders a list of <%= ns_table_name %>" do
     render
+    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
```
pirj, line 21: I suggest moving this inside `<% %>` so it doesn't appear in the resulting spec.

spec/generators/rspec/scaffold/scaffold_generator_spec.rb
```diff
@@ -208,16 +208,16 @@
...
+      it { is_expected.to contain('assert_select cell_selector, text: Regexp.new(3.to_s), count: 2') }
```
pirj, line 212: How `cell_selector` is available in this context? Where is it defined?

petergoldstein: `cell_selector` is a variable defined in the scaffold tool.  Honestly, this spec is a little weird, as it basically just tests if the test generator generates the test it expects.  I'm not sure I see the point.

lib/generators/rspec/scaffold/templates/index_spec.rb
petergoldstein: That won't help much.  Something needs to appear in the resulting spec, and that something is going to be different for Rails 7.0 and for previous versions of Rails.  So either we duplicate the logic in the spec (which seems kind of pointless) or we put it in a single variable here.

petergoldstein:
Ok, the Rails 6.1/Ruby 3.1 spec failures are...interesting.  In the Ruby 3.1 case the arguments are being treated as keyword arguments.  They look like:

`{"args"=>["required"], "_aj_ruby2_keywords"=>["args"]}`

If you run the same Rails version (6.1) under Ruby 3.0, the arguments passed to the jobs ar3:

`"required"`

So basically we need to deconstruct this.  I can fix the failing specs pretty easily, but I'm not sure if this indicates a wider issue with ActiveJob arguments in this combination.

lib/generators/rspec/scaffold/templates/index_spec.rb
pirj: I imagine the flow as:
1. generate a scaffold
2. upgrade Rails from 6 to 7

I can't imagine a failure if we've hardcoded `tr>td`. Because both the `index_spec` and the `index` have it like that being generated by Rails 6.


But, if we keep this ternary, after Rails 7 upgrade, `index_spec` will start failing, because `index` has `td` inside `tr`, while the spec would attempt to select `div>p`.

pirj:
>  Rails 6.1/Ruby 3.1 spec failures are...interesting

Indeed. And so is the fact the same spec passed with Rails 7.
Wondering if it's [this line](https://github.com/rspec/rspec-rails/pull/2546/files#diff-0b3ae675c5eb55ae78893e4ec44706c07219596036765950ad8b12c13764c4a4R111) that makes it fail.

lib/generators/rspec/scaffold/templates/index_spec.rb
petergoldstein: No, that's exactly backwards.

The scaffold is generated dynamically as part of the spec run.  With the version of Rails that is currently active.  So for you get:

Rails 6.x - tr>td
Rails 7.x - div>p

Hardcoding either one in the scaffold spec causes the scaffold spec to fail in at least one case.  

This is easy to confirm on the results on the current branch.  With the current code the scaffold and the various index specs pass across all Ruby/Rails combos.  Hardcoding will break it.

pirj: > scaffold is generated dynamically as part of the spec run

You think RSpec CI, I mean a real app developer's local machine.

RSpec CI would be green both ways. Real app developer's local test run will fail with the way it is now.

Please don't get me wrong, I suggest to:
```ruby
    render
    
<% for attribute in output_attributes -%>
    assert_select <% Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td' %>, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
<% end -%>
  end
```

On Rails 6, it will generate a spec with lines like:
```ruby
    assert_select 'tr>td', text: Regexp.new(3.to_s), count: 2
```

I think given the HTML table template Rails 6 used to generate, on developer's machine after update to Rails 7 this code will remain working.

What makes this code problematic?

pirj: > we've hardcoded

Sorry for being ambiguous.
By we I didn't mean me or you, I meant the template, `lib/generators/rspec/scaffold/templates/index_spec.rb`.
And by the product that will undergo hardcoding I meant the resulting `index_spec.rb` spec file on the real project's developer machine, not the template in this git repository.

petergoldstein: So the suggested approach doesn't work.  It breaks in CI and breaks on a developer's box.

* CI - for Rails 7 your approach will work for the generated index spec you reference (because the `tr>td` will be replaced in the generated spec.  But the scaffold spec - `spec/generators/rspec/scaffold/scaffold_generator_spec.rb ` - will fail, because it will have a hard-coded `tr>td`

* Developer's Box - We have the same issue for the scaffold spec if we generate the app at one Rails version and switch to another.  If a developer builds the smoke app under Rails 7, then the specs now fail.  Why does Rails 6 have a preferential position over Rails 7?  What happens if I generate the app on Rails 7 and move to Rails 6.

When a developer changes RAILS_VERSION then they should run `bundle exec rake clobber generate:app generate:stuff`.  That solves the upgrade problem entirely.

petergoldstein:
> > Rails 6.1/Ruby 3.1 spec failures are...interesting
> 
> Indeed. And so is the fact the same spec passed with Rails 7. Wondering if it's [this line](https://github.com/rspec/rspec-rails/pull/2546/files#diff-0b3ae675c5eb55ae78893e4ec44706c07219596036765950ad8b12c13764c4a4R111) that makes it fail.

I think I've almost got this.  I've fixed 2 out of the 3 failures in this case.

And I'm making headway on the remaining Rails 7 issues - I've basically got to get through the `verify_mailer_preview_path_spec.rb` fixes and that should be that.

pirj:
Fantastic! :zap: Thank you!

petergoldstein:
A quick update on this PR / Rails 7 issues in general.

I think our issues with Rails 7 fall into a few categories:

1. Scaffold template changes - at this point well understood, even if there's debate on how to handle
2. New output lines when running commands - these need to be filtered out, and for the most part that's done at this point.  Things like new deprecation warnings
3. Class loading changes with zeitwerk - this is still not fully understood

The 3rd is the one we're still wrestling with.  I think it's ultimately the cause of the two-ish remaining issues on the Rails 7 runs:

1. Not requiring `action_mailer/railtie` causes specs to fail - zeitwerk appears to be eager loading the app, and fails when it loads the existing mailer class.  `NotificationsMailer` is loaded and fails
2. The No ActiveRecord smoke test fails - I think zeitwerk is loading ActiveRecord at some point, even though the test app doesn't have it by default.

@pirj I think we've had trouble reproducing this locally because the `config/environments/test.rb` file in the example app has the following line:

`config.eager_load = ENV["CI"].present?`

petergoldstein:
@pirj By changing the value of `config.eager_load` I'm not able to reproduce locally.

petergoldstein:
Ok, I've got the actual spec failure in Ruby 3.1 / Rails 6.1 fixed.  But that branch is still failing because Rubocop with 3.1 is running into an error.

@pirj and @JonRowe is there any objection to me pulling out the rubocop run into a single, dedicated job - similar to what was just done on the `rspec-core` repo?

lib/generators/rspec/scaffold/templates/index_spec.rb
pirj: https://github.com/rspec/rspec-rails/commit/c68b5680de1e6828caa152f1485e889c09e5cf4b#diff-94efcbcbb995cd8f60c1956dbb889ec2af5999f3aef85e686278dd96b9a3be16R21

A competitive commit coincidentally implementing the approach I suggest.

> Why does Rails 6 have a preferential position over Rails 7? What happens if I generate the app on Rails 7 and move to Rails 6.

There is no preferential approach. If you were on Rails 7, generated those specs with `index` with `<p><strong>`. 
If you move to Rails 6 after, `index_spec` that will now start asserting that `index` should have `tr>td`. But it doesn't, it still has `<p><strong>`.

There seem to be a massive misunderstanding on this.

petergoldstein: There is no misunderstanding.  You've implemented a version of the alternative that I described [here](https://github.com/rspec/rspec-rails/pull/2552#discussion_r781685831).  Specifically:

> So either we duplicate the logic in the spec (which seems kind of pointless) or we put it in a single variable here.

You've duplicated the logic from the index generator in the previously static scaffold spec.  Note there is no mention of such a change in [this comment](https://github.com/rspec/rspec-rails/pull/2552#discussion_r781787417).  As I noted, the issue is we have dynamically generated things (index specs) and one statically generated thing (scaffold spec).  Without the modification of the scaffold spec your specs would fail.

To me it seems obvious that duplicating the logic in two places (as you do in the referenced commit) is less ideal, but it's not a hill I want to die on.  If there's a consensus that this approach is "better" I'm fine with that.  As noted elsewhere, my goal is to get things working.

Keep in mind that regardless of this spec's behavior, switching Rails versions without regenerating the smoke app would count as a bad idea.  There's no guarantee that the generated app would i) run at all or ii) run consistently because there may be Rails version dependent changes.  My earlier comment about blowing away and regenerating the Rails app when switching Rails versions still holds.

petergoldstein:
I've rebased in the externalized Rubocop run.  That should allow the Ruby 3.1/Rails6.1 branch to pass.

Tomorrow I'm going to look at the ActionMailer issue and see if I can wrap the mailer classes in `if` statements that are aware of the environment variable.  I think that will get the previews running completely green on the standard smoke app.  Then we just need to figure out where ActiveRecord is being pulled in for the no AR app.

pirj:
Thanks for your hard work!

pirj:
@petergoldstein A hint from a competitive PR fixing Rails 7/Ruby 3.1 https://github.com/rspec/rspec-rails/pull/2521/commits/f4736f6f789223d0ad54ebc738067f2f548c42ef#diff-2609fe753d96200fae508b4ac54ff71b7cfc1186f8bc226d55055b96ddf05413R177

@alpaca-tc I suggest you to cooperate and communicate on this here.

petergoldstein:
> @petergoldstein A hint from a competitive PR fixing Rails 7/Ruby 3.1 [f4736f6#diff-2609fe753d96200fae508b4ac54ff71b7cfc1186f8bc226d55055b96ddf05413R177](https://github.com/rspec/rspec-rails/commit/f4736f6f789223d0ad54ebc738067f2f548c42ef#diff-2609fe753d96200fae508b4ac54ff71b7cfc1186f8bc226d55055b96ddf05413R177)
> 
> @alpaca-tc I suggest you to cooperate and communicate on this here.

@alpaca-tc Happy to cooperate.

I do have some concerns about disabling eager loading based on Rails version.  I don't think that's an accurate representation of what a real app would look like, and so may give us misleading results.  I'd rather do something a bit more surgical.  That's why I avoided that solution.

What's the motivation behind the `unshift` for those files?

alpaca-tc:
> @ alpaca-tcここで協力してコミュニケーションすることをお勧めします。

I just noticed your comment. :pray: 

> I do have some concerns about disabling eager loading based on Rails version. I don't think that's an accurate representation of what a real app would look like, and so may give us misleading results. I'd rather do something a bit more surgical. That's why I avoided that solution.
> 
> What's the motivation behind the unshift for those files?

Since the test was originally intended to have `eager_load = false`, I fixed it to work as expected. https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/example_app_generator/spec/support/default_preview_path#L36

pirj:
Is there a good reason to set `eager_load` to true on CI? To my best memory, `config/environments/test.rb` has `config.eager_load = false` by default.

petergoldstein:
@pirj Typically you want eager_load to be true in a CI environment because that's the behavior you have in production environments.  If you have eager_load false it can hide issues that arise in production.  I have run into that a few times and it is considered by many to eager load in the CI environment for that reason.

That said, the rspec case is a little different than a free-standing app.  RSpec's "production environment" is an application or library's CI environment.  So it may be less of an issue here.

alpaca-tc:
> That said, the rspec case is a little different than a free-standing app. RSpec's "production environment" is an application or library's CI environment. So it may be less of an issue here.

I agree.
By the way, https://github.com/rails/rails/pull/43508 is the commit with eager_load set to `ENV["CI"].present?` in the Rails 7 CI test environment.

petergoldstein:
@alpaca-tc @pirj After thinking about it a bit more, I agree that disabling eager load shouldn't be an issue in this case.  Both because that's the way it works for Rails versions prior to 7.0 and because conceptually I think the rspec situation is different.

There were a few Cucumber failures which I fixed with a set of changes very similar to those here - https://github.com/rspec/rspec-rails/pull/2521 .  Sorry @alpaca-tc, I didn't see the PR until after I'd already committed.  I commented on the discussion in that PR.

While there's probably a little clean up required (including determining how to approach the scaffolding spec), I think this will come out passing at the end.

pirj:
I might be missing something essential and talking nonsense. Be eager loading on or off, why would `active_record` be required if it's not required in `config/application.rb`?
Is this because `gem 'rails'` doesn't have `require: false` in `Gemfile`?
Does this mean that if someone only needs a small fraction of what Rails has to offer, in production the whole thing is eager loaded anyway, and eats up memory?
That would be ridiculous, wouldn't it?

petergoldstein:
@pirj Figuring that out is essentially what we sidestepped by setting eager load to false.  I agree, it's not clear why it's getting loaded in that case.

pirj:
I couldn't reproduce the `no_ar_example_app` failure by setting `CI` env var to `true` or even changing this "app"'s `config/environments/test.rb` `config.eager_load` to `true`.

Can you reproduce the issue?

I'd say - let's keep the current `gsub_file "config/environments/test.rb", 'ENV["CI"].present?', "false"` hack in place, mark it as such just in case it's a Rails bug of some kind and it's eventually fixed.

Judging just by build times, there's no issues left for Rails 7. Do you think there's anything that needs attention?

@JonRowe How do you feel about:
 - reverting #2546
 - releasing what's left for 5.0.3 and 5.1.0
 - merge #2546 and this PR and releasing 6.0.0?
Can I help with anything on this front?

petergoldstein:
@pirj The following steps are reproducible in my local environment:

1. Select Ruby 3.1 or 3.0
2. Check out this branch
3. Run `RAILS_VERSION="~> 7.0.0" bundle exec rake no_active_record:smoke:app` .  This will wipe out any pre-existing version of the no Active Record app, and run the specs.  The specs will all pass.
4. Open `./tmp/no_ar_example_app/config/environments/test.rb` and change the value of `config.eager_load` from `false` to `true`.  Save that change.
5. Run `RAILS_VERSION="~> 7.0.0" bundle exec rake no_active_record:smoke` which ONLY runs the specs and doesn't rebuild the no_ar_example_app.  The specs fail with something like:

Screen Shot 2022-01-17 at 3 57 57 PM

pirj:
```
     98:       it 'respects the setting from `show_previews`' do
     99:         x= capture_exec(custom_env.merge('SHOW_PREVIEWS' => 'true'), exec_script)
    100:         require 'pry'; binding.pry
 => 101:         expect(
    102:           x
    103:         ).to eq("#{::Rails.root}/spec/mailers/previews")
    104:       end
    105:
    106:       it 'allows initializers to set options' do

[1] pry(#<RSpec::ExampleGroups::ActionMailerRailtieHook::InANonDevelopmentEnvironment>)> puts x.io

/Users/pirj/source/rspec-dev/repos/bundle/ruby/3.1.0/gems/activerecord-7.0.1/lib/active_record.rb:26:in `<top (required)>': ololololo (RuntimeError)
        from /Users/pirj/source/rspec-dev/repos/bundle/ruby/3.1.0/gems/activerecord-7.0.1/lib/active_record/railtie.rb:3:in `require'
        from /Users/pirj/source/rspec-dev/repos/bundle/ruby/3.1.0/gems/activerecord-7.0.1/lib/active_record/railtie.rb:3:in `<top (required)>'
        from /Users/pirj/source/rspec-dev/repos/bundle/ruby/3.1.0/gems/actionmailbox-7.0.1/lib/action_mailbox/engine.rb:6:in `require'
        from /Users/pirj/source/rspec-dev/repos/bundle/ruby/3.1.0/gems/actionmailbox-7.0.1/lib/action_mailbox/engine.rb:6:in `<top (required)>'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:31:in `require'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:31:in `block (2 levels) in <main>'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:4:in `require_file_stub'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:17:in `block in <main>'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:4:in `require_file_stub'
        from /Users/pirj/source/rspec-dev/repos/rspec-rails/tmp/no_ar_example_app/spec/support/default_preview_path:15:in `<main>'
```
:suspect:

example_app_generator/spec/support/default_preview_path
```diff
@@ -26,6 +26,10 @@ require_file_stub 'config/environment' do
     require "action_controller/railtie"
     require "action_mailer/railtie" unless ENV['NO_ACTION_MAILER']
     require "action_view/railtie"
+    if Rails::VERSION::STRING >= '6'
+      require "action_cable/engine"
+      require "action_mailbox/engine"
```
pirj, line 31: It's this line that's causing the "ActiveRecord is defined but should not be!" message in `no_ar_example_app`.

[`action_mailbox` requires `active_record`](https://github.com/rails/rails/blame/main/actionmailbox/lib/action_mailbox/engine.rb#L6).

🎉

petergoldstein: Good work.  I think the fix is slightly more complex than just removing that line.

Some work on my machine indicates that these changes may do it - https://github.com/petergoldstein/rspec-rails/pull/5 .  I want to ensure they run green on CI before pushing to this branch.

petergoldstein:
Ok, this last commit ran green.  And it fixes everything but the ActionMailer dependency on my side, even when eager loading.

To fix the ActionMailer dependency we'll need to wrap the definitions of the mailer classes in conditionals on that environment variable.

JonRowe:
No need to revert, see #2560, I pulled out `5-1-maintenance`, main is now essentially `6-0-dev`.

JonRowe:
We do need Ruby 3.1 fixes backported seperately to Rails 7 fixes.

pirj:
I've merged #2521, so those parts are no longer needed.

pirj:
I have merged conflicting changes in #2563 to `main` and am going to rebase this PR.

All the `have_enqueued_mail` changes done here will remain, but I intend to extract some code to `FeatureCheck` [like in #2563](https://github.com/rspec/rspec-rails/pull/2563/commits/948b67be61e09e3225eb6d5dbb47c3e4ed7a8065#diff-ecdb788624767749b0bc8033987cfb94242212240d1746241196dc3ec715cfb6R46).

example_app_generator/generate_stuff.rb
```diff
@@ -167,4 +167,10 @@ def using_source_path(path)
 gsub_file 'spec/controllers/uploads_controller_spec.rb',
           'skip("Add a hash of attributes valid for your model")',
           '{}'
+
+if Rails.version >= '7.0.0'
+  # Some gems (ActionMailBox, ActionCable, etc.) are not used when running `example_app_generator/spec/verify_mailer_preview_path_spec.rb`, so `eager_load` must be false.
+  gsub_file "config/environments/test.rb", 'ENV["CI"].present?', "false"
```
pirj, line 173: Is this hack still needed? I intend to try removing it now when we're not requiring `action_mailbox` from `no_ar_example_app`.

pirj:
@petergoldstein I dared approaching the problem with AR loaded in `no_ar_example_app` slightly differently.
And changed a couple of other things. Appreciate if you glance over.

I didn't change the way you approach scaffold spec generator, though, as I didn't take time to gather arguments to convince you with my approach.

example_app_generator/spec/verify_mailer_preview_path_spec.rb
```diff
@@ -108,6 +111,8 @@ def have_no_preview
       end
 
       it 'handles action mailer not being available' do
+        skip "Rails 7 forces eager loading on CI, loads app/mailers and fails" if Rails::VERSION::STRING.to_f >= 7.0
```
petergoldstein, line 114: @pirj Sure, this is a reasonable way to address the issue.

petergoldstein:
@pirj I think this all looks fine.  Thanks for pulling it together.

lib/rspec/rails/feature_check.rb
```diff
@@ -43,8 +43,11 @@ def has_action_mailbox?
         defined?(::ActionMailbox)
       end
 
-      def ruby_3_1?
-        RUBY_VERSION >= "3.1"
+      # TODO: add a self-explanatory method name. See https://github.com/rspec/rspec-rails/pull/2552#issuecomment-1009508693 for hints
+      def rails_6_1_and_ruby_3_1?
+        return false if RUBY_VERSION < "3.1"
+
+        ::Rails::VERSION::STRING >= "6.1" && ::Rails::VERSION::STRING < "7"
```
pirj, line 50: I haven't seen this in practice, but it's tempting to write this as:
```ruby
"6.1" <= ::Rails::VERSION::STRING < "7"
```

petergoldstein: I'd probably avoid that.

Simpler is usually better than more concise.  I'd want a developer to be able to look at it and immediately know what it means.  And with the proposed change I don't think that'd be true.

lib/rspec/rails/feature_check.rb
```diff
@@ -43,8 +43,11 @@ def has_action_mailbox?
         defined?(::ActionMailbox)
       end
 
-      def ruby_3_1?
-        RUBY_VERSION >= "3.1"
+      # TODO: add a self-explanatory method name. See https://github.com/rspec/rspec-rails/pull/2552#issuecomment-1009508693 for hints
```
pirj, line 46: I couldn't. Can you please help me? Is there some explanation, like "the last hash argument is now treated as a positional argument instead of being auto-expanded into kwargs"?
I searched in https://rubyreferences.github.io/rubychanges/3.1.html and couldn't find anything related.

example_app_generator/generate_action_mailer_specs.rb
```diff
@@ -31,6 +31,7 @@
   if skip_active_record?
     comment_lines 'spec/support/default_preview_path', /active_record/
     comment_lines 'spec/support/default_preview_path', /active_storage/
+    comment_lines 'spec/support/default_preview_path', /action_mailbox/
```
pirj, line 34: This is the key change to skip loading AR in `no_ar_example_app`, it's patching `default_preview_path`.

lib/rspec/rails/feature_check.rb
petergoldstein: So I'm not sure there's a simple description other than "this is how it works for this combo".  The Argument processing in the Rails codebase is pretty complex.  There's an intersection of several changes happening at once, combined with an goal to keep support working from Ruby 2.5 to Ruby 3.1.  That's a lot of moving parts, and I didn't try and dissect exactly what causes the change on input to rspec-rails.

Then rspec-rails is post-processing the arguments with its own conditions, because of how it processes arguments to the matchers.  So making that all line up is complicated, and I basically looked at the discrepancies and found the simplest resolution for the failing case.

JonRowe:
Thanks for this, I've merged the initial CI work bar the mailer stuff that clashes

pirj:
It appears that all of this has been merged in parts.
Thanks a lot, @petergoldstein !

PR #2554, opened 2022-01-11
Fix Ruby 3.1 on CI for dependent builds (this time for real)

pirj:
Follow-up to #2553 I've quite missed the fact that dependent builds specify no `RAILS_VERSION`, and another `case` branch was used, causing `net-` dependency failures.

Also, add Ruby 3.1/Rails 6.x to CI matrix, since Rails 6.0 is what our dependent builds use.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,9 @@ jobs:
              RAILS_VERSION: '~> 7.0.0'
 
          # Rails 6.1 builds >= 2.5
+         - ruby: 3.1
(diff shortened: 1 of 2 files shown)


pirj:
This depends on [this fix](https://github.com/rspec/rspec-rails/pull/2552/commits/9c9eb3986a22b5b1e4210052067e88cd493edf3f#diff-0b3ae675c5eb55ae78893e4ec44706c07219596036765950ad8b12c13764c4a4R94) from #2552

PR #2556, opened 2022-01-12
This updates Rubocop and uses the .rubocop_rspec_base.yml for consistency

petergoldstein:


.rubocop_rspec_base.yml
```diff
@@ -0,0 +1,316 @@
...
+Metrics/CyclomaticComplexity:
+  Max: 10
+
+# We use YARD to enforce documentation. It works better than rubocop's
+# enforcement...rubocop complains about the places we re-open
+# `RSpec::Expectations` and `RSpec::Matchers` w/o having doc commments.
+Style/Documentation:
+  Enabled: false
+
+# We still support 1.8.7 which requires trailing dots
```
pirj, line 39: We don't really in `rspec-rails`, it's just the 3.9 of other `rspec-*` repos.

petergoldstein: Fair enough.  And we don't have to use the same config.

That said, I just think it makes sense to use a consistent .rubocop_rspec_base.yml across the rspec codebases.  Same with the Hash rocket syntax changes - not needed here, but still needed for the other repos which support earlier Ruby versions.

lib/generators/rspec/scaffold/scaffold_generator.rb
```diff
@@ -7,18 +7,18 @@ module Generators
     class ScaffoldGenerator < Base
       include ::Rails::Generators::ResourceHelpers
       source_paths << File.expand_path('../helper/templates', __dir__)
-      argument :attributes, type: :array, default: [], banner: "field:type field:type"
+      argument :attributes, :type => :array, :default => [], :banner => "field:type field:type"
```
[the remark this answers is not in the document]
petergoldstein: So we could move the cop that requires hash rocket out of the common base, and put it in todo for the repos that still need to support older Rubies.  That would work.  Then we could just revert the hash rocket changes here - this repo was using JSON-style syntax.

.rubocop_rspec_base.yml
pirj: Makes total sense to me. When we release RSpec 4, we could update those base rules for other RSpec repos and align then with `rspec-rails` 👍 
Let's keep this base file 👍

lib/generators/rspec/scaffold/scaffold_generator.rb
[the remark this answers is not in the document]
pirj: What if we keep it in base, but set `Enabled: false` in todo? Will todo take precedence over base?
That way, base is still common across all repos and intact. And we reduce churn for this PR.

spec/rspec/rails/matchers/route_to_spec.rb
```diff
@@ -13,50 +13,50 @@ def assert_recognizes(*)
...
+        { 'queryitem' => 'queryvalue' }
       )
       expect(get("path?queryitem=queryvalue")).to route_to("controller#action", queryitem: 'queryvalue')
     end
 
     it "routes with multiple query parameters" do
       expect(self).to receive(:assert_recognizes).with(
-        {controller: "controller", action: "action", queryitem: "queryvalue", qi2: 'qv2'},
-        {method: :get, path: "path"},
-        {'queryitem' => 'queryvalue', 'qi2' => 'qv2'}
```
pirj, line 48: Maybe disable this cop, too? This code looks good with or without spaces.

petergoldstein: I think consistency wins on this Cop.  This is a cop that's enabled on the base, and unlike the others which are necessary to support older Rubies, this is a style issue.  So either disable it on the base (which is fine) or make these changes.

pirj:
Thank you, @petergoldstein !

PR #2557, opened 2022-01-12
Abort with argument

arzezak:
It condenses two lines into one being that `abort` exits with 1.

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -28,8 +28,7 @@ require 'rspec/rails'
 begin
   ActiveRecord::Migration.maintain_test_schema!
 rescue ActiveRecord::PendingMigrationError => e
-  puts e.to_s.strip
-  exit 1
+  abort e.to_s.strip
 end
 <% end -%>
 RSpec.configure do |config|


lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -28,8 +28,7 @@
 begin
   ActiveRecord::Migration.maintain_test_schema!
 rescue ActiveRecord::PendingMigrationError => e
-  puts e.to_s.strip
-  exit 1
+  abort "#{e}".strip
```
pirj, line 31: There is a difference, though. `abort` outputs to STDERR, while `puts` to STDOUT.
I tend to think that STDERR is more appropriate here.

Cosmetic: I'd prefer to keep `e.to_s`. Is there a reason to use string interpolation?

pirj (approved):
Thanks!

lib/generators/rspec/install/templates/spec/rails_helper.rb
arzezak: I agree that STDERR seems to be more appropriate. No reason to use string interpolation. I'll go ahead and update the PR. Thank you very much.

pirj:
@JonRowe Do you have any objections on merging this small change?

JonRowe:
I'm on the fence due to the change in stdout to stderr

arzezak:
It'd be the same approach as the one already taken on line #6.

pirj:
Would you wait a bit? We can merge this to the soon-to-be-released RSpec 4.

JonRowe:
Having reviewed this further, I don't mind merging this as is, the intent of the original PR was to produce a better output for this error, and that it was to stdout probably didn't occur. @pirj I leave this decision up to you.

JonRowe (approved):

PR #2562, opened 2022-01-19
Setup 5-1-maintenance without Rails 7 but with Ruby 3.1

JonRowe:
Rails 7 will be supported in rspec-rails 6 via `main`, this just tracks Ruby 3.1 support into the upcoming 5.1.x release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,31 +29,10 @@ jobs:
       fail-fast: false
       matrix:
        include:
-         # Edge Rails (7.1) builds >= 2.7
-         - ruby: '3.0'
-           allow_failure: true
-           env:
-             RAILS_VERSION: 'main'
-         - ruby: 2.7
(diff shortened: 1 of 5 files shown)


pirj:
Note to self - don't call branches ending with `*-maintenance` to prevent double the builds 😄

pirj:
@JonRowe Green. Does it look good to you?
Fixes for Rails 7 and Ruby 3.1 got interleaved. I extracted the part that I could remember was specifically for Rails 7.

pirj:
I'll go ahead and merge this.

JonRowe:
@pirj I'm going to be unpicking this a bit, it includes changes not needed, I was working on this on my machine but just hadn't responded.

pirj:
Please let me know what you plan to remove from `5-1-maintenance`. I may work on making the same changes on `5-0-maintenance` in parallel.

PR #2563, opened 2022-01-22
Add Ruby 3.1 (without Rails 7 support) to `main`

pirj:


.github/workflows/ci.yml
```diff
@@ -29,31 +29,10 @@ jobs:
...
-             RAILS_VERSION: '~> 7.0.0'
-         - ruby: '3.0'
-           allow_failure: true
-           env:
-             RAILS_VERSION: '~> 7.0.0'
-         - ruby: 2.7
-           allow_failure: true
-           env:
-             RAILS_VERSION: '~> 7.0.0'
-
```
JonRowe, line 55: I've reverted this part because main supports Rails 7.

PR #2566, opened 2022-01-25
Improve mailer argument deserialisation for 6.1 on Ruby 3.1

JonRowe:
@pirj this is a simpler version of the changes required to fix mailers solely on Ruby 3.1 for Rails 6.1

The problem was not the different mailers but simply that Rails changed (only on 6.1) how it deserialized arguments, this logic restores that original implementation for our matcher and is the change I want to ship in the next 5.0 and 5.1 releases.

Arguably this could be in the jobs matcher and I'm surprised our specs pass as is for the jobs matcher, but its probably because that has less explicit requirements around deserialising.

I'm going to merge this to 5-1-maintenance and cherry pick to 5-0-maintenance on green, so you don't have to worry about that.

### DIFF
--- a/lib/rspec/rails/feature_check.rb
+++ b/lib/rspec/rails/feature_check.rb
@@ -39,10 +39,6 @@ module RSpec
         defined?(::ActionMailbox)
       end
 
-      def ruby_3_1?
-        RUBY_VERSION >= "3.1"
-      end
-
       def type_metatag(type)
         "type: :#{type}"
       end
--- a/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
+++ b/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
@@ -399,16 +399,18 @@ RSpec.describe "HaveEnqueuedMail matchers", skip: !RSpec::Rails::FeatureCheck.ha
         }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with(1, 2)
       end
 
-      it "matches arguments when mailer is parameterized" do
+      it "passes with provided argument matchers" do
         expect {
           UnifiedMailer.with('foo' => 'bar').test_email.deliver_later
-        }.to have_enqueued_mail(UnifiedMailer, :test_email).with('foo' => 'bar')
-      end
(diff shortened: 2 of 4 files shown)


lib/rspec/rails/feature_check.rb
```diff
@@ -39,10 +39,6 @@ def has_action_mailbox?
         defined?(::ActionMailbox)
       end
 
-      def ruby_3_1?
-        RUBY_VERSION >= "3.1"
-      end
-
```
JonRowe, line 45: If we want this it should be in RSpec Support, its a later improvement that could be made once 3.11 is released.

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -393,22 +393,18 @@ def self.name; "NonMailerJob"; end
...
+          a_hash_including(params: {'foo' => 'bar'})
+        )
 
-      it "matches arguments when mixing parameterized and non-parameterized emails" do
         expect {
           UnifiedMailer.with('foo' => 'bar').email_with_args(1, 2).deliver_later
-        }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with({'foo' => 'bar'}, 1, 2)
+        }.to have_enqueued_mail(UnifiedMailer, :email_with_args).with(
+          a_hash_including(params: {'foo' => 'bar'}, args: [1, 2])
+        )
```
JonRowe, line 407: This just unpicks the changes to the behaviour

pirj: I wonder if both:
```ruby
have_enqueued_mail(UnifiedMailer, :email_with_args).with(
          a_hash_including(params: {'foo' => 'bar'}, args: [1, 2])
        )
```
and 
```ruby
have_enqueued_mail(UnifiedMailer, :email_with_args).with({'foo' => 'bar'}, 1, 2)
```
would match.
I was believing that the change from #2516 made both match, but it's not confirmed in specs.

pirj:
I really like how you've structured the code here.
Do you want me to attempt to adjust how we handle this for 6.0 in #2554?

PR #2569, opened 2022-01-26
Fix undefined Rails version for other repos sub-builds

pirj:
https://github.com/rspec/rspec-support/runs/4956870915?check_suite_focus=true

### DIFF
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -1,5 +1,16 @@
 version_file = File.expand_path("../.rails-version", __FILE__)
 
+# This is required for Ruby 3.1, because in Ruby 3.1 these gems were moved to
+# bundled gems from default gems. This issue was fixed in Rails Rails 7.0.1.
+# Discussion can be found here - https://github.com/mikel/mail/pull/1439
+def add_net_gems_dependency
+  if RUBY_VERSION >= '3.1'
+    gem 'net-smtp', require: false
+    gem 'net-imap', require: false
+    gem 'net-pop', require: false
+  end
+end
+
 case version = ENV['RAILS_VERSION'] || (File.exist?(version_file) && File.read(version_file).chomp) || ''
 when /main/
   gem "rails", :git => "https://github.com/rails/rails.git"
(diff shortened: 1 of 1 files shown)


Gemfile-rails-dependencies
```diff
@@ -32,7 +32,7 @@ end
 # automatically included.  This issue was fixed on the Rails
 # side in Rails 7.0.1, but is not yet fixed in the Rails 6.1.x
 # branch.  Discussion can be found here - https://github.com/mikel/mail/pull/1439
-if RUBY_VERSION >= '3.1' && version.split(' ').last < '7.0'
```
JonRowe, line 35: Maybe just do `version = ...` at the top and run the case statement of a local variable?

pirj: We skip this for:
- pre 6.1 because it doesn't have Action Mailbox
- 7.0.1+ because it does the same (presumably) itself

The only `-stable` build left is `5-2-stable` that we'll drop.

In any case, this doesn't hurt on 3.1. And there's no other way to work around this for 6.1.

I don't really want to overcomplicate this.

Ideally, we could:
```diff
- version = ENV['RAILS_VERSION'] || (File.exist?(version_file) && File.read(version_file).chomp) || ''
+ version = ENV['RAILS_VERSION'] || (File.exist?(version_file) && File.read(version_file).chomp) || "~> 6.0.0"
```

and combine the last two branches.

This would allow to keep `&& version.split(' ').last < '7.0'`.
Is it worth it?

JonRowe: Or we could return from the case statement a skip variable? Seems like skipping it on rails 7 is a good idea to ensure we don't change its behaviour?

pirj:
I've just realized that it's not only for Rails 6.1.
On Rails 6.0 that is used in sub-builds, Action Mailer loads the `mail` gem, [and it](https://github.com/mikel/mail/blob/master/lib/mail.rb#L9):
```ruby
require 'net/smtp'
```

Adjusting accordingly.

JonRowe (approved):

PR #2570, opened 2022-01-28
Fix mailer argument deserialisation for 6.1 on Ruby 3.1

pirj:
I've cherry-picked #2566 to `main`.
Locally,on Ruby 3.1 it passed for 6.0 (to fix [failing sub-builds](https://github.com/rspec/rspec-core/runs/4965401014?check_suite_focus=true)) and for 7.0, too.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -9,7 +9,3 @@ Layout/LineLength:
 # Over time we'd like to get this down, but this is what we're at now.
 Metrics/MethodLength:
   Max: 43 # default: 10
-
-Metrics/ClassLength:
-  Exclude:
-    - lib/rspec/rails/matchers/have_enqueued_mail.rb
--- a/features/matchers/have_enqueued_mail_matcher.feature
+++ b/features/matchers/have_enqueued_mail_matcher.feature
@@ -68,6 +68,7 @@ Feature: have_enqueued_mail matcher
     When I run `rspec spec/mailers/my_mailer_spec.rb`
     Then the examples should all pass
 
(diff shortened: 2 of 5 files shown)


pirj:
Without fixes from #2552, 7.0 builds expectedly fail.

What fails on 6.1 and 6.0 are [those 2 scenarios](https://github.com/rspec/rspec-rails/pull/2546/files#diff-bbe9cb28b3444186be72a3d5556ef8b9fcede6ead12dc3f17b7f2f7e68a1eef7R71).
It seems it needs more specs for various combinations depending on `self.delivery_job = ActionMailer::MailDeliveryJob` setting like even the same Rails version can have a different global or even a mailer-local setting for a job that affects serialization, e.g. mentioned [here](https://github.com/rspec/rspec-rails/issues/2351#issuecomment-916229119), and in our own specs:
```ruby
    class UnifiedMailer < ActionMailer::Base
      self.delivery_job = ActionMailer::MailDeliveryJob
# vs
      self.delivery_job = ActionMailer::DeliveryJob
```

I can hack on it next week.

My impression is that we won't be able to make `enuque_mail` to match with both `with({'foo' => 'bar'}, 1, 2)` and `with(params: {'foo' => 'bar'}, args: [1, 2])`, so this will be a breaking change. I admit recklessly merging #2546 without noticing this change of behaviour.

JonRowe:
@pirj I've merged this because I'm reconcilling the `main` branch with the released `5.1.1`, my intent is to ressurrect the discussion around the changed behaviour in another branch on top of this

PR #2573, opened 2022-02-14
Fixup `have_enqueued_job` matcher on job retries

### LINKED ISSUE #2668: New ActiveJob is not counted when other job executed

Ruby version: ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
Rails version: Rails 7.0.4.3
RSpec version: RSpec 3.12
  - rspec-core 3.12.1
  - rspec-expectations 3.12.2
  - rspec-mocks 3.12.3
  - rspec-rails 6.0.1
  - rspec-support 3.12.0

## Observed behaviour

Consider this example:

```rb
    ReportJob.perform_later(*report_args)

    expect {
      perform_enqueued_jobs(only: ReportJob)
    }.to enqueue_job(ActionMailer::MailDeliveryJob)
```

The ReportJob sends an email and this test should pass but fails:

```
     Failure/Error:
       expect {
         perform_enqueued_jobs(only: ReportJob)
       }.to enqueue_job(ActionMailer::MailDeliveryJob)
     
       expected to enqueue exactly 1 jobs, but enqueued 0
```

I found the reason in the matcher simply taking the count of pre-existing jobs and dropping them off the queue when comparing:
https://github.com/rspec/rspec-rails/blob/01704c50c146f720db914724c25681781ecefb23/lib/rspec/rails/matchers/active_job.rb#L233-L235

## Expected behaviour

I expected the matcher to only ignore pre-existing jobs and not newly generated jobs. I would also have understood if the matcher would only look for MailDeliveryJobs but the types of jobs executed and enqueued are different here.

I saw in the original pull request that it was counting only the requested job type but I'm wondering if we could actually remember the previous jobs (with id) and drop only those off the queue. I'm not familiar with the adapters but I was wondering if this would work:

```rb
            original_enqueued_jobs = queue_adapter.enqueued_jobs
            proc.call
            in_block_jobs = queue_adapter.enqueued_jobs - original_enqueued_jobs
```

It may have performance implications.

I hope that the above examples are easy enough but let me know if I need to create an app for this. Maybe more useful, here's an (untested) spec:

```diff
diff --git a/spec/rspec/rails/matchers/active_job_spec.rb b/spec/rspec/rails/matchers/active_job_spec.rb
index 38aee9d3..0c24276d 100644
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -94,6 +94,14 @@ RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_activ
       }.to have_enqueued_job.exactly(1)
     end
 
+    it "counts jobs even when others are removed" do
+      heavy_lifting_job.perform_later
+      expect {
+        perform_enqueued_jobs
+        hello_job.perform_later
+      }.to have_enqueued_job(hello_job)
+    end
+
     it "passes when negated" do
       expect { }.not_to have_enqueued_job
     end
```
https://github.com/mkllnk/rspec-rails/commit/7ecda545c72b776d0632ba78e79a25b403e8a32c

ojab:
Previously we were checking only job counts, so if one job was
performed and one job was added - matcher failed. Check by unique
id (`#hash`) instead.
We cannot use `job['job_id']` here, because job retains `job_id` on retry.

This works only with `rails >= 6.1`, because `retry_on` was introduced in 6.0 and apparently `test` queue adapter is broken for retries in 6.0: it adds executed job back with added `exception_executions` along with actual retried job. So dunno how to proceed here, please advise /shrug

Test CI run: https://github.com/ojab/rspec-rails/runs/5183906574?check_suite_focus=true 

Also:
fixes #2668

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -230,26 +230,11 @@ module RSpec
           def matches?(proc)
             raise ArgumentError, "have_enqueued_job and enqueue_job only support block expectations" unless Proc === proc
 
-            original_enqueued_jobs_hashes = queue_adapter.enqueued_jobs.map(&:hash)
-
+            original_enqueued_jobs = Set.new(queue_adapter.enqueued_jobs)
             proc.call
+            enqueued_jobs = Set.new(queue_adapter.enqueued_jobs)
 
-            in_block_jobs = queue_adapter.enqueued_jobs.each_with_object({}) do |job, jobs|
-              jobs[job.hash] ||= { job: job, count: 0 }
-              jobs[job.hash][:count] += 1
-            end
-
-            original_enqueued_jobs_hashes.each do |job_hash|
-              in_block_jobs[job_hash][:count] -= 1 if in_block_jobs.key?(job_hash)
-            end
-
-            in_block_jobs = in_block_jobs.each_value.flat_map do |job_and_count|
-              count, job = job_and_count.values_at(:count, :job)
-
-              Array.new(count, job) if count.positive?
-            end
-
-            check(in_block_jobs.compact)
+            check(enqueued_jobs - original_enqueued_jobs)
           end
 
           def does_not_match?(proc)
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -230,11 +230,26 @@ module RSpec
           def matches?(proc)
             raise ArgumentError, "have_enqueued_job and enqueue_job only support block expectations" unless Proc === proc
 
-            original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
-            proc.call
-            in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
+            original_enqueued_jobs_hashes = queue_adapter.enqueued_jobs.map(&:hash)
 
-            check(in_block_jobs)
+            proc.call
+
+            in_block_jobs = queue_adapter.enqueued_jobs.each_with_object({}) do |job, jobs|
+              jobs[job.hash] ||= { job: job, count: 0 }
+              jobs[job.hash][:count] += 1
+            end
+
+            original_enqueued_jobs_hashes.each do |job_hash|
+              in_block_jobs[job_hash][:count] -= 1 if in_block_jobs.key?(job_hash)
+            end
+
+            in_block_jobs = in_block_jobs.each_value.flat_map do |job_and_count|
+              count, job = job_and_count.values_at(:count, :job)
+
+              Array.new(count, job) if count.positive?
+            end
+
+            check(in_block_jobs.compact)
           end
 
           def does_not_match?(proc)
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -98,6 +98,31 @@ RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_activ
       expect { }.not_to have_enqueued_job
     end
 
+    context "when job is retried" do
+      include ActiveJob::TestHelper
+
+      let(:retried_job) do
+        Class.new(ActiveJob::Base) do
+          retry_on StandardError, wait: 5, queue: :retry
+
+          def self.name; "RetriedJob"; end
+          def perform; raise StandardError; end
+        end
+      end
+
+      before do
+        stub_const("RetriedJob", retried_job)
+        queue_adapter.perform_enqueued_jobs = true
+      end
+
+      it "passes with reenqueued job" do
+        time = Time.current.change(usec: 0)
+        travel_to time do
+          expect  { retried_job.perform_later }.to have_enqueued_job(retried_job).on_queue(:retry).at(time + 5)
+        end
+      end
+    end
+
     it "fails when job is not enqueued" do
       expect {
         expect { }.to have_enqueued_job
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -98,6 +98,20 @@ RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_activ
       expect { }.not_to have_enqueued_job
     end
 
+    context "when previously enqueued jobs were performed" do
+      include ActiveJob::TestHelper
+
+      before { stub_const("HeavyLiftingJob", heavy_lifting_job) }
+
+      it "counts newly enqueued jobs" do
+        heavy_lifting_job.perform_later
+        expect {
+          perform_enqueued_jobs
+          hello_job.perform_later
+        }.to have_enqueued_job(hello_job)
+      end
+    end
+
     context "when job is retried" do
       include ActiveJob::TestHelper
 
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -101,24 +101,22 @@ RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_activ
     context "when job is retried" do
       include ActiveJob::TestHelper
 
-      let(:retried_job) do
+      let(:unreliable_job) do
         Class.new(ActiveJob::Base) do
           retry_on StandardError, wait: 5, queue: :retry
 
-          def self.name; "RetriedJob"; end
+          def self.name; "UnreliableJob"; end
           def perform; raise StandardError; end
         end
       end
 
-      before do
-        stub_const("RetriedJob", retried_job)
-        queue_adapter.perform_enqueued_jobs = true
-      end
+      before { stub_const("UnreliableJob", unreliable_job) }
 
       it "passes with reenqueued job" do
         time = Time.current.change(usec: 0)
         travel_to time do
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -5,6 +5,7 @@ on:
       - 'main'
       - '*-maintenance'
       - '*-dev'
+      - '*'
```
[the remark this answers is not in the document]
JonRowe: You will have to remove this when you've finished work on it, as it causes double builds on our pipeline, your PR is not running because of Github rules surrounding first time contributors, we will have to hit the button each time you force push.

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -98,6 +98,31 @@ def self.name; "LoggingJob"; end
...
+
+      before { stub_const("RetriedJob", retried_job) }
+
+      it "passes with reenqueued job" do
+        retried_job.perform_later
+
+        time = Time.current.change(usec: 0)
+        travel_to time do
+          expect { perform_enqueued_jobs(queue: :default) }
+            .to have_enqueued_job(retried_job).on_queue(:retry).at(time + 3)
```
pirj, line 121: I don't get it. Does this mean 3 seconds later? Where does this 3 come from?

ojab: default, I'll change it to be explicit `retry_on …,wait: 3` along with dropping unrelated stuff.

ojab:
hmm.. Actually looking @ CI, 6.0 is green, so probably failure is rails-6.0 <-> ruby-3.1 compat locally. And 5.2 fails with `unknown keyword queue`. so `retry_on` is there. 
I'll check tomorrow and I guess would get green CI without any workarounds :sweat_smile:

ojab:
It became more complicated due to rails-6.0 pushing jobs with the same `#hash` to the queue, but CI is green https://github.com/ojab/rspec-rails/runs/5198727860?check_suite_focus=true
Cleaned up, still two additional commits: first fixing rspec dependencies version in gemspec, second restricting rails-5.2 version for ruby-2.2 compat (filled https://github.com/rails/rails/pull/44435 about it).

JonRowe:
We've merged build fixes for Rails and updated the minimum versions so this should be ok to be rebased without your ci changes

ojab:
:bow: done

pirj (approved):
Thanks!

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -230,11 +230,26 @@ def initialize(job)
...
             raise ArgumentError, "have_enqueued_job and enqueue_job only support block expectations" unless Proc === proc
 
-            original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
+            original_enqueued_jobs_hashes = queue_adapter.enqueued_jobs.map(&:hash)
+
             proc.call
-            in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
 
-            check(in_block_jobs)
+            in_block_jobs = queue_adapter.enqueued_jobs.each_with_object({}) do |job, memo|
```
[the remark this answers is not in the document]
ojab: Done.
I tried to name it more clearly and failed. IMHO `jobs` is not ideal because everything is about jobs in this method and it's not clear what `jobs` we're talking about, but I guess it's better than `memo`.

ojab:
:crying_cat_face:

pirj:
Green. @JonRowe any objections on merging this change?

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -230,11 +230,26 @@ def initialize(job)
...
             raise ArgumentError, "have_enqueued_job and enqueue_job only support block expectations" unless Proc === proc
 
-            original_enqueued_jobs_count = queue_adapter.enqueued_jobs.count
+            original_enqueued_jobs_hashes = queue_adapter.enqueued_jobs.map(&:hash)
+
             proc.call
-            in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
 
-            check(in_block_jobs)
+            in_block_jobs = queue_adapter.enqueued_jobs.each_with_object({}) do |job, jobs|
```
pirj: It appears to me that storing by the `hash` key is not necessary.
I've added a couple more jobs of the same kind, and they have a different `job_id` assigned, resulting in a different `hash`.

As a result, `count` is always a 1.

Below, a `-= 1` and a further `.positive?` with `Array.new` and a `compact` is basically subtraction of previously enqueued jobs from all enqueued jobs.

I dared pushing a commit that compares with a `Set#-`.

Can you write a failing spec to prove this approach is insufficient?

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -98,6 +98,31 @@ def self.name; "LoggingJob"; end
       expect { }.not_to have_enqueued_job
     end
 
+    context "when job is retried" do
```
pirj, line 101: What confuses me is that this spec is passing with the original implementation, too.
Can you write one that would fail with the original, and would pass with yours/mine?

pirj (changes requested):
Needs a test that would fail without changes, so that changes would make it green.

pirj:
The previous implementation was:
1. Save the number of currently enqueued jobs
2. Run the block passed to `expect { ... }` that can:
  - perform previously scheduled jobs and remove them from enqueued
  - enqueue more jobs
3. Remove the number of previously enqueued jobs from the list of currently enqueued jobs

Of course, since the matcher is scoped to one specific job class, it's quite an edge case.

This (previous, before the PR's) implementation was not considering jobs that were at the bottom of the list, and moved to the top of the list, e.g.:
```diff
- job with args [1]
- job with args [2]
+ job with args [1]
+ job with args [2]
```
And `expect { ... }.to have_enqueued_jobs` would fail because it would think that no jobs were (re-)enqueued.

Equally, if the order of jobs would change, or if some jobs succeed, or if more jobs succeed than fail and a re-scheduled, the expectation would fail.

I believe we can leave some of those cases for follow-up improvements, but one I'd love to see is:
```ruby
# some setup
expect { ... }.to have_enqueued_job(MyJob)
```
that would fail if we just compare the counts of matching jobs.

It is unfair to do so with counts because yes, one job was performed, and one enqueued.
If you scope it to the `:retry` queue, the expectation would pass. Because there were no jobs on that queue, and now there's one. But this is unfair, too, as it works fine with the previous count-only implementation, too.

How about an example with a job that was on `:retry` queue already and got re-enqueued again?

JonRowe:
This looks fine to me, but I defer to @pirj as we do want a failing spec this fixes :)

pirj:
> Previously we were checking only job counts, so if one job was performed and one job was added - matcher failed.

This is an incorrect assumption.

The "before" for the matcher is everything prior to what is inside the `expect { ... }` block.
I am not aware of an option to perform jobs with the test adapter that were enqueued previously.
`queue_adapter.perform_enqueued_jobs = true` and `perform_enqueued_jobs { ... }` do not perform previously enqueued jobs.

So if a job is enqueued, performed and schedules itself all inside `expect { ... }`, all previously enqueued jobs are out of equation with
```ruby
in_block_jobs = queue_adapter.enqueued_jobs.drop(original_enqueued_jobs_count)
```
and all the jobs that are left enqueued (e.g. when `perform_enqueued_jobs = true` but `perform_enqueued_at_jobs = false`) like in the case with job retries, are still detected. This is what your newly added example shows, as it is passing just fine with the current implementation.

Even though I'd love to merge my shiny refactoring involving `Set` that considers if the **same** job gets re-enqueued, I have no good idea how it can affect a real-world spec.

I failed to write an example that would break the existing implementation, and with this I'm more inclined to close this PR.

Please do not hesitate to reopen if you can write a failing spec.

pirj:
Actually, a spec might be possible to write with a direct non-block usage of `perform_enqueued_jobs` [as here](https://github.com/rspec/rspec-rails/issues/2668#issuecomment-1513880992).

pirj (approved):

pirj:
@ojab Does this look correct to you?
Please accept my apologies for nitpicking and ignorance of the block-less form of `perform_enqueued_jobs` that was needed to reproduce the issue.

ojab:
@pirj oh, right, sorry for bogus spec. I'll update the PR tomorrow to use non-block form and will check that it fails properly in master.

ojab:
Checked the commit where monkeypatch was introduced into local codebase, it's 
```ruby
retried_job.perform_later

expect  { perform_enqueued_jobs }.to 
```
so yeah, spec is indeed doesn't mirror what was broken :facepalm: 
Changed/verified that it breaks with rails-6.1 on main and works with changes in this PR/pushed.

pirj:
Which monkey patch?

How does this work?
We set the adapter to perform jobs, then schedule a job.
The job should immediately execute, fail and be re-scheduled for later, right?
What does the second `perform_enqueued_jobs` do, performs it once again, and then it is re-scheduled once again?

```ruby
        queue_adapter.perform_enqueued_jobs = true
      end

      it "passes with reenqueued job" do
        time = Time.current.change(usec: 0)

        retried_job.perform_later

        travel_to time do
          expect  { perform_enqueued_jobs }.to
```

I strongly prefer the solution with `Set`, as it's significantly simpler.
Does it solve the same issue?

pirj:
Thank you, @ojab !
I went with my implementation approach for the fix. If you feel that yours is superior, please don't hesitate to open a PR with a failing spec.

JonRowe:
Released in 6.0.2

PR #2574, opened 2022-02-14
Update RSpec gem versions

JonRowe:


pirj:
Does it make sense to remove Rails 5 from the matrix? Will we still support it in `rspec-rails` 6?
```
[rspec-rails 5.x](https://github.com/rspec/rspec-rails/tree/5-0-maintenance) for Rails 6.x.
[rspec-rails 4.x](https://github.com/rspec/rspec-rails/tree/4-1-maintenance) for Rails from 5.x or 6.x.
```

pirj:
The error seems to come from `rack`

```
rack-2.1.4/lib/rack/media_type.rb:18: warning: `&' interpreted as argument prefix
```

Wondering if [this controversial commit](https://github.com/rack/rack/pull/1586/commits/565350a207cd822117db4cfd8f5a8e53845a3f87) that is both "Cleanups since master is 2.3+" and "Cleanups since master is 2.2+". The gemspec says `>= 2.3.0`.

JonRowe:
@pirj the Rack one is just a warning where as the change that broke the build has a PR open to fix https://github.com/rails/rails/pull/44435

Gemfile-rails-dependencies
```diff
@@ -23,17 +23,31 @@ when /stable$/
   gem 'activerecord-jdbcsqlite3-adapter', git: 'https://github.com/jruby/activerecord-jdbc-adapter', platforms: [:jruby]
 
   gem_list.each do |rails_gem|
-    gem rails_gem, :git => "https://github.com/rails/rails.git", :branch => version
+    if version == '5-2-stable' && RUBY_VERSION < '2.3'
```
pirj, line 26: `version == '5-2-stable'` would never trigger here, as if falls under `when /stable$/` above.

Gemfile-rails-dependencies
```diff
@@ -23,17 +23,31 @@ when /stable$/
...
+      # Until 5-2-stable is fixed on Ruby 2.2, pin to last working version
```
pirj, line 27: I thought it was broken on `5.2.x`, and was fine on `5-2-stable`. As far as I recall, it was later fixed on `5.2.x`, too (https://github.com/rspec/rspec-rails/commit/2b8fffed490125c5b08de36177765cb0cb234037#). Did it break again, now on `5-2-stable`?

PR #2579, opened 2022-03-09
Prevent unrelated exceptions by requiring `rails_helper` at `.rspec`

aeroastro:
When there is a bug in the code on the load,
RSpec catches the exception, aborts the spec file,
proceeds to the following spec file, and retries `require`
with partially loaded Rails application.

In some cases, we see many identical errors,
such as `NameError`, which are noisy but help developers
find the bug. In other cases, we see many unrelated errors,
typically `FrozenError` on finalized objects,
which push out the original exception raised only from the first spec.

The following comment describes this issue in detail.
https://github.com/thoughtbot/factory_bot_rails/issues/303#issuecomment-434560625

This patch prevents unrelated exceptions by the following rspec-core feature.
The reason why I edited the `.rspec` is that this feature is only applicable to
`--require` option.
https://github.com/rspec/rspec-core/pull/2568

This issue is a relatively well-known issue, and the well-known
workaround is to use `--fail-fast` when we see many errors.
https://stackoverflow.com/questions/47320772/rails-runtimeerror-cant-modify-frozen-array-when-running-rspec-in-rails

By introducing this patch, we no longer need the workaround.
All we need to do is run `bundle exec rspec` as usual.

### DIFF
--- a/lib/generators/rspec/install/install_generator.rb
+++ b/lib/generators/rspec/install/install_generator.rb
@@ -41,6 +41,9 @@ DESC
 
         replace_generator_command(spec_helper_path)
         remove_warnings_configuration(spec_helper_path)
+
+        dot_rspec_path = File.join(tmpdir, '.rspec')
+        prepend_require_rails_helper_option(dot_rspec_path)
       end
 
       def replace_generator_command(spec_helper_path)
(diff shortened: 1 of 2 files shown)


JonRowe:
:wave: Thanks for taking the time to suggest this, the whole purpose of splitting the `spec_helper.rb` into `spec_helper.rb` for plain ruby and `rails_helper.rb` is to allow not loading Rails when not necessary. This change would completely remove the point of that split so its not something we would want to change.

aeroastro:
Thank you for your prompt reply and the detailed description.
I understood the detailed reason behind the separation of `spec_helper.rb` and `rails_helper.rb`.

PR #2580, opened 2022-03-10
Include ActiveSupport::Testing::TaggedLogging for Rails >= 7

### LINKED ISSUE #2545: undefined local variable or method `tagged_logger' during perform_enqueued_jobs

Ruby version: 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-darwin21]
Rails version: 7.0.0
RSpec 3.10
  - rspec-core 3.10.1
  - rspec-expectations 3.10.1
  - rspec-mocks 3.10.2
  - rspec-rails 5.0.2
  - rspec-support 3.10.3

## Observed behaviour

Inside an rspec support module, my app is calling `perform_enqueued_jobs`. The spec in question that is using this support module is expecting this call to raise an exception.

When I attempt to run this spec, the following is raised:
```
got #<NameError: undefined local variable or method `tagged_logger'
```

When I debug into `perform_enqueued_jobs`, I am taken into `_assert_nothing_raised_or_warn` inside `activesupport-7.0.0/lib/active_support/testing/assertions.rb`, When debugging the block below, it shows `e` as having the expected exception.

```ruby
        rescue Minitest::UnexpectedError => e
          if tagged_logger && tagged_logger.warn?
            warning = <<~MSG
              #{self.class} - #{name}: #{e.error.class} raised.
              If you expected this exception, use `assert_raises` as near to the code that raises as possible.
              Other block based assertions (e.g. `#{assertion}`) can be used, as long as `assert_raises` is inside their block.
            MSG
            tagged_logger.warn warning
          end

          raise
        end
```
Debugging further reveals that the `if tagged_logger` check above is causing the undefined error.

I reported this on Rails forum [here](https://discuss.rubyonrails.org/t/undefined-local-variable-or-method-tagged-logger-inside-rails-7-activesupport-assertions/79541) and a member of the core team suggested the following:

"This seems to be a bug in RSpec that doesn’t implement the tagged_logger. Please report the issue in the RSpec issue tracker."

## Expected behaviour

The spec should finish and allow `to raise_error` to check for the expected exception. This spec is working as expected in Rails 6.

shanecav84:
Fixes #2545. Not sure if anyone got to this yet. There was also another problem mentioned in that issue about a missing 'name' method, which this PR does not address.

### DIFF
--- a/snippets/include_activesupport_testing_tagged_logger.rb
+++ b/snippets/include_activesupport_testing_tagged_logger.rb
@@ -44,10 +44,6 @@ ActiveJob::Base.queue_adapter = :test
 # This connection will do for database-independent bug reports
 ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
 
-RSpec.configure do |config|
-  config.use_transactional_fixtures = true
-end
-
 class TestJob < ActiveJob::Base
   def perform; end
 end
--- a/snippets/include_activesupport_testing_tagged_logger.rb
+++ b/snippets/include_activesupport_testing_tagged_logger.rb
@@ -57,11 +57,12 @@ RSpec.describe 'Foo', type: :job do
     it 'raises the explicitly thrown error' do
       allow_any_instance_of(TestJob).to receive(:perform).and_raise(TestError)
 
+      test_error = TestError.new('foo')
       # Rails 6.1+ wraps unexpected errors in tests
       expected_error = if Rails::VERSION::STRING.to_f >= 6.1
-                         Minitest::UnexpectedError.new(TestError)
+                         Minitest::UnexpectedError.new(test_error)
                        else
-                         TestError
+                         test_error
                        end
 
       expect { perform_enqueued_jobs { TestJob.perform_later } }
--- a/snippets/include_activesupport_testing_tagged_logger.rb
+++ b/snippets/include_activesupport_testing_tagged_logger.rb
@@ -1,0 +1,75 @@
+if __FILE__ =~ /^snippets/
+  fail "Snippets are supposed to be run from their own directory to avoid side " \
+       "effects as e.g. the root `Gemfile`, or `spec/spec_helpers.rb` to be " \
+       "loaded by the root `.rspec`."
+end
+
+# We opt-out from using RubyGems, but `bundler/inline` requires it
+require 'rubygems'
+
+require "bundler/inline"
+
+# We pass `false` to `gemfile` to skip the installation of gems,
+# because it may install versions that would conflict with versions
+# from the main `Gemfile.lock`.
+gemfile(false) do
+  source "https://rubygems.org"
+
+  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
+
+  # Those Gemfiles carefully pick the right versions depending on
+  # settings in the ENV, `.rails-version` and `maintenance-branch`.
+  Dir.chdir('..') do
+    eval_gemfile 'Gemfile-sqlite-dependencies'
+    # This Gemfile expects `maintenance-branch` file to be present
+    # in the current directory.
+    eval_gemfile 'Gemfile-rspec-dependencies'
+    # This Gemfile expects `.rails-version` file
+    eval_gemfile 'Gemfile-rails-dependencies'
+  end
+
+  gem "rspec-rails", path: "../"
+end
+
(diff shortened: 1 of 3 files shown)


pirj (approved):
Thank you.
Do you think it's possible to add a snippet that would fail without this fix? [An example snippet](https://github.com/rspec/rspec-rails/blob/b4f58f44170526edd1f3d25bbb918e2fcdaae7a8/snippets/avoid_fixture_name_collision.rb#L48).

shanecav84:
I've added the snippet. Not quite sure if I've done it correctly. The snippet will fail without the fix.

snippets/include_activesupport_testing_tagged_logger.rb
```diff
@@ -0,0 +1,75 @@
...
+require "rspec/rails"
+
+ActiveJob::Base.queue_adapter = :test
+
+# This connection will do for database-independent bug reports
+ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
+
+RSpec.configure do |config|
+  config.use_transactional_fixtures = true
+end
```
pirj, line 49: This seems redundant.

pirj:
Perfect, thank you!

pirj:
I've re-triggered the build as Rails seems to have finally released 5.2.x with a fix for Ruby 2.2 support.

But they also released 6.1.5 and broke something for us.

Do you think [this](https://github.com/rspec/rspec-rails/runs/5508066551?check_suite_focus=true) is related?
```
[ActiveJob] Failed enqueuing TestJob to Test(default): TestError (TestError)
F

Failures:

  1) Foo error raised in perform_enqueued_jobs with block raises the explicitly thrown error
     Failure/Error:
       expect { perform_enqueued_jobs { TestJob.perform_later } }
         .to raise_error(expected_error)

       expected TestError, got #<Minitest::UnexpectedError: Unexpected exception> with backtrace:
         # include_activesupport_testing_tagged_logger.rb:71:in `block (5 levels) in <main>'
         # include_activesupport_testing_tagged_logger.rb:71:in `block (4 levels) in <main>'
         # include_activesupport_testing_tagged_logger.rb:71:in `block (3 levels) in <main>'
     # include_activesupport_testing_tagged_logger.rb:71:in `block (3 levels) in <main>'
```

shanecav84:
Not sure what's up with that failing jRuby test.

pirj:
Quite an interesting case.
@headius Would you like to have a look?

JonRowe:
I tried to install JRuby this morning to take a look at this failure, but it seems broken on M1 macs via asdf at the moment, do we need this to be a snippet? Its currently a regression test as well, what actually does the tagged logger do? Could we write an integration test instead?

headius:
@JonRowe Latest JRuby 9.3 should be working well on M1. Perhaps asdf is not updated to the latest release? (9.3.3).

The failure error is very peculiar. I don't know why `TestError` would not have the `backtrace` method under any circumstance. I do notice it is running 9.3.1.0, which is a couple releases old now.

I'll try to reproduce locally.

headius:
I do see now that it appears to be trying to call `backtrace` against the `TestError` **class**, not against an instance. Unsure why that is happening so far.

headius:
I have not been able to reproduce so far, but I got some other weird errors that seem to be an issue in my environment.

I would suggest trying to upgrade to 9.3 in this PR since it should be upgraded anyway and might fix the issue.

shanecav84:
Upgrading JRuby did not fix the issue. See also https://github.com/rspec/rspec-rails/pull/2580/commits/5d9d3b0d7edbd627510ebb0d0de9b035fd024238 where I instantiated the expected error to get around the JRuby error but that caused other errors.

headius:
I have cleaned up my local env issue and can now reproduce the same error!

headius:
I will try to look into this for a fix for JRuby 9.3.4 but no promises... we are hoping to release next week. If anyone can narrow this down to some specific test that fails, it would be helpful. It almost looks like it is happening *after* the suite completes, but I am unfamiliar with your suites' output.

shanecav84:
@headius 

1. Checkout the branch for this PR
2. Run bundle
3. `cd snippets; ruby include_activesupport_testing_tagged_logger.rb` will run the test.

pirj:
Superseded by #2587 
Thank you for the hard work!

PR #2587, opened 2022-04-03
Include ActiveSupport::Testing::TaggedLogging for Rails >= 7 (rebased)

JonRowe:
This is #2580 rebased

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -181,5 +181,15 @@ module RSpec
     #
     # @private
     TestUnitAssertionAdapter = MinitestAssertionAdapter
+
+    # @private
+    module TaggedLoggingAdapter
+      require 'active_support/testing/tagged_logging'
+      include ActiveSupport::Testing::TaggedLogging
+
+      # Just a stub as TaggedLogging is calling `name`
+      def name
+      end
+    end
   end
 end
--- a/lib/rspec/rails/example/rails_example_group.rb
+++ b/lib/rspec/rails/example/rails_example_group.rb
@@ -12,11 +12,7 @@ module RSpec
       include RSpec::Rails::MinitestLifecycleAdapter
       include RSpec::Rails::MinitestAssertionAdapter
       include RSpec::Rails::FixtureSupport
-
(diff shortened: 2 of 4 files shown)


pirj (approved):

JonRowe:
> I have mixed feeling from defining the name method for all Rails example groups. A spec with let(:name) works as expected.
Wondering what other side effects might be.

A `let(:name) { .. }` would override the definition here, and thus if it was not string safe it likely would break... but... this is the case regardless...

pirj:
I see what you mean. With the var naming, though, `name`, chances of it not having a `to_s` are lower.
Still, it would be a weird side effect to show it in tagged logging output.
As far as I understand it, `name` should evaluate to the example title.
Any idea why it is not yet available in this context?
I tried with `RSpec.current_example.metadata[:name]`, it was empty too.
In any case, `@example` resolves to `nil`. And it's not because it's [explicitly nullified in the constructor](https://github.com/rspec/rspec-rails/blob/a34dea796e0c470d5483d732a8a3c596dceb6abe/lib/rspec/rails/adapters.rb#L131), it is `nil` if you set a breakpoint above, too.

As an alternative approach with no side effects, we would have to wrap `tagger_logger`, delegate to it, and provide it some `name`. In the best case, our example name. Or just `nil` if it turns out to be impossible.
I can take a stab at it if you think it's a viable approach.

pirj:
Unfortunately enough, `name` is used not only in `before_setup`, but in assertions, too, e.g. [this](https://github.com/rails/rails/blob/de53ba56cab69fb9707785a397a59ac4aaee9d6f/activesupport/lib/active_support/testing/assertions.rb#L254):
```ruby
        def _assert_nothing_raised_or_warn(assertion, &block)
          assert_nothing_raised(&block)
        rescue Minitest::UnexpectedError => e
          if tagged_logger && tagged_logger.warn?  warning = <<~MSG
              #{self.class} - #{name}: #{e.error.class} raised.
```

I'll push the wrapped that I've written, but it fails.

pirj:
I'd go with a risk of `name` raising when its definition is incapable of `.to_s`, and it logging nonsense when `name` is defined.
@JonRowe Any objections on merging as is?

PR #2589, opened 2022-04-08
Increase Aruba exit_timeout value for TruffleRuby

bjfish:
To allow enough time for subprocesses to complete when using TruffleRuby.


pirj:
Thanks!

Issue #2595, opened 2022-04-22
reset_sessions! during system specs (ie. after_teardown) should occur as late as possible


timdiggins:
**Is your feature request related to a problem? Please describe.**

It's really difficult to integrate the capybara-screenshot gem with rails system specs. (see https://github.com/mattheworiordan/capybara-screenshot/issues/225), but I think there is a generic issue for other libraries and integrations that you want the resetting of capybara sessions to happen as late as possible (and critically, after all RSpec Notifiers have formatted their output).

The minitest `after_teardown` callback occurs during the `after` hooks in the lifecycle of system specs, but occurs after all the `after` hooks (after example.run during the  `around`) in all other specs.  

https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/lib/rspec/rails/adapters.rb#L69-L78

https://github.com/rspec/rspec-rails/blob/9fb6c9c482da29b8803dd6f17b34d37a6367b8f2/lib/rspec/rails/example/system_example_group.rb#L103-L116

The `after_teardown` is where the `Capybara.reset_sessions!` happens and (as of Rails 6.0) the take_screenshot occurs  occurs during before_teardown/
 
**Describe the solution you'd like**

I'd like the after_teardown to happen during an around callback.
This wouldn't be possible if rspec-rails was still supporting Rails v5.2

This may change behaviour for downstream users, but should be able to be included in 6 because a major•revision with an appropriate warning in the changelog?

I'll do a possible implementation/illustration in a PR.

**Describe alternatives you've considered**

I've considered whether this should be an adaptation of rails (I don't think there's any argument for changing rails).

I've considered if this should be an adaptation of downstream libraries such as capybara-screenshot (see https://github.com/mattheworiordan/capybara-screenshot/issues/225, this seems tricky to do -- though this may be down to architectural decisions in capybara-screenshot).

I've also attempted to monkey-patch either the rails `ActionDispatch::SystemTesting::TestHelpers::SetupAndTeardown` class in an ad hoc way in my own rails projects but this is quite difficult to do because it is already heavily monkey-patched by `RSpec::Rails::SystemExampleGroup`.

**Additional context**

I will raise this in rspec@googlegroups.com, but I think it's probably easier to see as set out in this issue.

JonRowe:
Closed in #2596

Issue #2526, opened 2021-10-13
`after` hook for system tests has no access to the page


arielj:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 2.6.5
Rails version: 6.1.4.1
RSpec-core: 3.10.1
RSpec-rails: 5.0.2

## Observed behaviour

When trying to get the JavaScript code coverage data from the window object on an after hook (shown here https://jtway.co/collecting-javascript-code-coverage-with-capybara-in-ruby-on-rails-application-d0cb83a86a90), when the hook is executed, the page is blank.

The same code using MiniTest works fine (the before_teardown runs before the test is done so the page is at the last state) and it seems like at some point this worked fine on RSpec too since that blogpost describes the process and I've seen the same in other places.

It seems like rspec's `after` is equivalent to the `after_teardown` instead of the `before_teardown` hook from minitest?

## Expected behaviour

I would expect the hook to have access to the page as it was at the end of the execution of the example to be able to get the code coverage information from the `window` object.

## Can you provide an example app?

https://github.com/fastruby/js-coverage-sample-app this sample app has both MiniTest and RSpec configured with some code related to getting the JS test coverage.

The only way I'm able to get the JS coverage data is calling the `dump_js_coverage` method at the end of the test (in `spec/system/index_spec.rb`). If I try to use an after hook in `spec/rails_helper.rb#65`, running `page.evaluate_script("window.__coverage__")` returns nil instead of the coverage data.

You can run `COVERAGE=true rails spec` to try this.

Same function is being used as a `before_teardown` hook for MiniTest at `test/application_system_test_case.rb`.

You can run `COVERAGE=true rails test:all` to try this.

arielj:
After some debugging, I found a workaround. I added this to my rails_spec.rb file so it sets the before_teardown function of the system specs so when it executes the `original_before_teardown` here https://github.com/rspec/rspec-rails/blob/fe95eacb376f7fa558d3ffeeaa9afa6d3110a540/lib/rspec/rails/example/system_example_group.rb#L112, my code runs in the correct order 

```
module RSpec::Rails::SystemExampleGroup
  def before_teardown
    dump_js_coverage
  end
end
```

pirj:
It is indeed confusing that `config.after` and example group level `after` run on the different sides of `teardown`.
[...]

PR #2596, opened 2022-04-22
run after_teardown in system specs after around (like in other specs)

timdiggins:
This will only work as expected (print out screenshot location after failed tests)
in >= rails 6.0 but 5.2 is no longer supported in main.

This may possibly change behaviour, but can be included in major
revision?

This should enable easier integration of other libraries such as
capybara-screenshot which require predictable (and late) running of
Capybara.reset_sessions!

This is an illustration (or possible candidate to fix) #2595

### DIFF
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -108,9 +108,7 @@ module RSpec
           orig_stdout = $stdout
           $stdout = StringIO.new
           begin
-            if ::Rails::VERSION::STRING >= '6.0'
-              original_before_teardown.bind(self).call
-            end
+            original_before_teardown.bind(self).call
             original_after_teardown.bind(self).call
           ensure
             myio = $stdout
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -109,13 +109,17 @@ module RSpec
           $stdout = StringIO.new
           begin
             original_before_teardown.bind(self).call
-            original_after_teardown.bind(self).call
           ensure
             myio = $stdout
             myio.rewind
             RSpec.current_example.metadata[:extra_failure_lines] = myio.readlines
             $stdout = orig_stdout
           end
+        end
+
+        around do |example|
+          example.run
+          original_after_teardown.bind(self).call
         end
       end
     end
--- a/spec/rspec/rails/example/system_example_group_spec.rb
+++ b/spec/rspec/rails/example/system_example_group_spec.rb
@@ -91,5 +91,38 @@ module RSpec::Rails
         expect(example.metadata[:extra_failure_lines]).to eq(["line 1\n", "line 2\n"])
       end
     end
+
+    describe "hook order" do
+      it 'calls Capybara.reset_sessions (TestUnit after_teardown) after any after hooks' do
+        calls_in_order = []
+        allow(Capybara).to receive(:reset_sessions!) { calls_in_order << :reset_sessions! }
+
+        group = RSpec::Core::ExampleGroup.describe do
+          include SystemExampleGroup
+
+          before do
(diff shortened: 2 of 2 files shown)


pirj:
Might also be related to https://github.com/rspec/rspec-rails/issues/2526, but to my understanding this change won't fix it.

JonRowe:
Can you write up a test for this that fails without the change?

timdiggins:
> Might also be related to #2526, but to my understanding this change won't fix it.

I think the reason (some) after hooks have no access to the page is because the reset_sessions! has already occurred. If this is the case then this would fix it. However if you are loading rspec-rails first (before requiring any other gems / adding otheer around hooks), I think that the reset_sessions should occur last (because `after` always load before, and the reset_sessions! after is created first when rspec-rails is first required). But if you do any `append_after` those would happen after reset_sessions!

timdiggins:
> Can you write up a test for this that fails without the change?

Before I write this, can I try writing some pseudo code and see if this makes sense:

**Given**: A custom (system spec only) notifier/formatter that relies on presence of capybara session 
**When**: you run this example system spec in the presence of this notifier
**Then**: the custom notifier can't access the capybara session (and then either fails to print, or could raise an exception)

If that's the kind of thing, then I could look at notifier/formatter specs for a template? @JonRowe

JonRowe:
I'd write a integration test that shows that the hook methods are called in the order you want to change this to, you could add a fake 3rd party module to show that is executed correctly in the sequence, but this change distills down to the order of which our hooks interact with the "test unit" hooks.

timdiggins:
@JonRowe Thanks for the suggestion for the approach for writing the test -- really helped.

I started by writing a test that wanted the `Capybara.reset_sessions!` call (in the `after_teardown` hook) to happen after all the `after` hooks AND after the notification for "example_failed" ran. But I'm not sure if this is possible or even actually desirable.

Here's my thinking: 

1) Capybara.reset_sessions! needs to occur before the rollback of the transaction (this is so that any lurking ajax calls for example are stopped before the transaction rolls back - otherwise you can get these happening at the beginning of the next spec surprising)
2) We do want this as late as possible - no after hook (or ideally any around hook) should get this
3) However a formatter handling an `example_failed` notification shouldn't necessarily expect capybara to still have valid sessions (ie. the reset_sessions! to have not yet happened). This is capybara-screenshot's current approach.

Therefore I still want to change how the ordering of the callbacks (for 2 above), but I don't think this will solve the issue for capybara-screenshot.

I think here it would be better if capybara-screenshot (for system specs) monkey-patched the rails take_screenshot (which it is in effect improving on) and either let rspec-rails capture output, or if it's still using a formatter.

So I'll update the PR to have implement 2 above and also if I can propose a change to capybara-screenshot. But when I can...  In the meantime (or afterwards) feedback welcome

timdiggins:
@JonRowe I've updated my PR as discussed with a test and a fix. It seems to fix some the integration with capybara-screenshot

timdiggins:
FYI I've also tested manually that:

* given a vanilla rails 6.1 app (with this branch of rspec-rails) when theres a failing system test, it takes a screenshot (via rails mechanism) and outputs the path
* given a vanilla rails 6.1 app (with this branch of rspec-rails, and the latest Capybara-screenshot), it takes a screenshot via  capybara-screenshot and outputs the html and png paths. 

NB  with capybara-screenshot, both the capybara-screenshot and the rails mechanisms run and output their paths. I think this can be resolved (or lived with) either in capybara-screenshot, or the specific rails-app repo.

One more note. It's not easily possible to backport this fix into supporting rails 5.2 while supporting the rails screenshot because this happens in the same rails callback.

timdiggins:
> So ..   has the screenshot capability really been broken since Rails 5.1 ?    

In system specs, capybara-screenshot has not been printing out screenshot locations on failure (though saving the files). Feature specs working fine.

PR #2597, opened 2022-04-23
Infer the spec type for generator specs

jasonkarns:
Generator specs are [generated to spec/generator](https://github.com/rspec/rspec-rails/blob/6f8e3ab154a1562919ff1c08b8d19ae5ff6152e9/lib/generators/rspec/generator/generator_generator.rb#L12)

Spec-type inferrence (based on file location) is not presently active for generator specs.

This adds `:generator ` spec type to the mapping for inference by file location.


JonRowe:
Thanks, I'm merging this for completeness and its simple, but at some point this functionality will go away as its much more preferred to tag specs manually as its less problematic.

PR #2605, opened 2022-06-06
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ on:
   pull_request:
     branches:
       - '*'
+permissions:
+  contents: read
+
 jobs:
   rubocop:
     name: Rubocop


pirj:
I guess the [key here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) is:

> If you specify the access for any of these scopes, all of those that are not specified are set to none.

pirj:
I can't find the default values for permissions. Are they all on by default?
Unrelated to this PR, just in general. This seems to be a major security hole. What is the impact? Is it possible to steal secret keys, e.g. for repository push access, if there's something like "publish doc changes in gh-pages branch" flow?
Is there a related GH ticket (https://github.com/github/securitylab or https://github.com/actions/runner)?

naveensrinivasan:
> I can't find the default values for permissions. Are they all on by default?

AFAIK they are on by default because GH Action wasn't designed with permissions.

> Unrelated to this PR, just in general. This seems to be a major security hole. What is the impact? Is it possible to steal secret keys, e.g. for repository push access, if there's something like "publish doc changes in gh-pages branch" flow?
> Is there a related GH ticket (https://github.com/github/securitylab or https://github.com/actions/runner)?

pirj:
To spare you of the burden of dealing with the rest of our repos, I'll replicate those changes myself if @JonRowe approves.

JonRowe:
Thanks!

PR #2619, opened 2022-09-15
use the tag itself instead of the variable

JuanVqz:
[while testing rails 7](https://github.com/JuanVqz/lalo/blob/acbc3a62d0d091e82e4c136746e2408b9ca496b7/spec/views/orders/index.html.erb_spec.rb#L14-L18) I did notice the following test has a variable called `cell_selector` then I asked my self why do we need to know what rails version are we working on. could we generate only the tag without telling everybody how to do it?
and here we are, trying to implement it. if you think this is not useful please let me know.

```rb
  # before
  it "renders a list of orders" do
    render
    cell_selector = Rails::VERSION::STRING >= "7" ? "div>p" : "tr>td"
    assert_select cell_selector, text: Regexp.new("MyText".to_s), count: 2
  end
  
  # after, >= rails 7
  it "renders a list of orders" do
      render

      assert_select "div>p", text: Regexp.new("MyText"), count: 2
  end

  # after <= rails 6 
  it "renders a list of orders" do
    render

    assert_select "tr>td", text: Regexp.new("MyText".to_s), count: 2
  end
```

### DIFF
--- a/lib/generators/rspec/scaffold/templates/index_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/index_spec.rb
@@ -18,9 +18,10 @@ RSpec.describe "<%= ns_table_name %>/index", <%= type_metatag(:view) %> do
 
   it "renders a list of <%= ns_table_name %>" do
     render
-    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
+
+<% cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td' -%>
 <% for attribute in output_attributes -%>
-    assert_select cell_selector, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
+    assert_select <%= cell_selector %>, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
 <% end -%>
   end
 end
--- a/lib/generators/rspec/scaffold/templates/index_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/index_spec.rb
@@ -19,9 +19,9 @@ RSpec.describe "<%= ns_table_name %>/index", <%= type_metatag(:view) %> do
   it "renders a list of <%= ns_table_name %>" do
     render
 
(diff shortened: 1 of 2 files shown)


lib/generators/rspec/scaffold/templates/index_spec.rb
```diff
@@ -18,9 +18,10 @@
 
   it "renders a list of <%= ns_table_name %>" do
     render
-    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
+
+<% cell_selector = Rails::VERSION::STRING.to_f >= 7.0 ? 'div>p' : 'tr>td' -%>
```
JuanVqz, line 22: I was wondering if we need the parent Rails version
 
```suggestion
<% cell_selector = ::Rails::VERSION::STRING.to_f >= 7.0 ? 'div>p' : 'tr>td' -%>
```

pirj:
> why do we need to know what rails version are we working on

There's a lengthy discussion [here](https://github.com/rspec/rspec-rails/pull/2552#discussion_r781619422).

An alternative approach like you suggest is to bake this in into the generator.
There's a case when you've generated specs on Rails 6, and then switched to Rails 7, and a some other scenarios.
Which approach is more appealing to you, and why?

JuanVqz:
> > why do we need to know what rails version are we working on
> 
> 
> 
> There's a lengthy discussion [here](https://github.com/rspec/rspec-rails/pull/2552#discussion_r781619422).
> 
> 
> 
> An alternative approach like you suggest is to bake this in into the generator.
> 
> There's a case when you've generated specs on Rails 6, and then switched to Rails 7, and a some other scenarios.
> 
> Which approach is more appealing to you, and why?

To be honest I don't think somebody will keep the scaffold as it is, you will ended up removing it.
I'll close the PR, thank you for sharing the info. 👍

JonRowe:
I think this is reasonable tbh, or maybe this information should some how be baked into a method which the test uses, I didn't realise we were generating `if rails_version` in specs, that information is definietly intended to be invisible to developers.

PR #2621, opened 2022-10-03
Fix job spec file name in job_spec.feature

mogulla3:
Hi! 

I have corrected the spec file name in job_spec.feature as I believe the Job spec file name is generally the product code file name with `_spec` at the end. 
I noticed this when reading [Relish](https://relishapp.com/rspec/rspec-rails/v/5-1/docs/job-specs/job-spec).

If I have missed something, you can close this PullRequest.

Thank you.


pirj:
Thank you!

Issue #2624, opened 2022-10-13
v6.0.0 breaks if a helper spec subject is named `name` or if a helper module defines a `#name` method with arguments


magni-:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 3.0.3
Rails version: 7.0.3.1
RSpec version: 6.0.0

## Observed behaviour

We had a helper spec start to fail when attempting to upgrade `rspec-rails` to `v6.0.0`. While debugging to figure out why it was just the one helper spec that failed (while many others continued passing), I figured out the issue had to do with the helper in question defining a method called `#name`. We usually name our subjects after the method they're testing. In this case, switching to an unnamed subject caused a different error: it turned out there are two separate issues.

### When using `subject(:name)`
``NoMethodError: undefined method `view_context' for nil:NilClass``

### When defining a `#name` helper with at least one argument
`ArgumentError: wrong number of arguments (given 0, expected 1)`
This breaks all specs in the file, not just the specs for `#name`

## Expected behaviour

The specs in the example app should all pass.

## Can you provide an example app?

https://github.com/magni-/rspec-rails-dummy

pirj:
Oh [I was wrong](https://github.com/rspec/rspec-rails/pull/2587#issuecomment-1097060817)

JonRowe:
It turns out the module has a setup hook which is triggered which uses `#name`, so in #2625 I've replaced it with the method by hand, that branch should fix the issue @magni-

magni-:
@JonRowe Thanks for the speedy fix! I've updated https://github.com/magni-/rspec-rails-dummy, this fixes the errors seen on 6.0.0 👍🏼

If you try running the specs yourself, you'll see there's still a couple of errors, but those were already present on 5.1.2 and seem to be due to `helper` including _all_ helpers rather than just the one being described, which breaks when multiple helpers define a method with the same name. I don't know if that's worth opening another Issue for or if it's `As Designed` ? (I couldn't find where in the RSpec or Rails internals this behavior was coming from. I'm guessing it's [here](https://github.com/rails/rails/blob/v7.0.4/actionview/lib/action_view/test_case.rb#L102-L105), but I couldn't see where `_helpers` was getting set).

JonRowe:
The last two specs seem to be Rails behaviour due to the overlapping method names, its not something we control we just provide access to an action view testcase.

magni-:
Dependabot PR for 6.0.1 came in, our build passed ☺️

PR #2625, opened 2022-10-14
Implement tagged_logger by hand

JonRowe:
Similar to #2461, rather than use Rails modules verbatim, vendor the import part of what they do (in this case define the private `tagged_logger` method) so we have less of an overlap. In particular the actual module does this:

```
def before_setup
  if tagged_logger && tagged_logger.info?
    heading = "#{self.class}: #{name}"
    divider = "-" * heading.size
    tagged_logger.info divider
    tagged_logger.info heading
    tagged_logger.info divider
  end
  super
end
```

Which is what is causing issues for #2624

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -184,11 +184,12 @@ module RSpec
 
     # @private
     module TaggedLoggingAdapter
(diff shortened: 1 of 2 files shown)


pirj (approved):
So much better, thank you!

PR #2628, opened 2022-10-25
Reduce rubocop_todo things

SHinGo-Koba:
Hi!

RSpec always helps me to get my products powerful so I'm thinking that I would like to make codes better a little bit.

What I did is to reduce some of things in .rubocop_todo.yml which seem like correctable.

I would be happy if you like it.

Thank you :)


pirj:
Thank you, @SHinGo-Koba !
Next time we won't have to approve running CI jobs.

PR #2630, opened 2022-10-27
Update RSpec gem versions to 3.12/3.13.pre

pirj:
RSpec 3.12 was recently released.

PR builds fail, since in `rspec-*` repos we have 3.13.pre now, see e.g [this](https://github.com/rspec/rspec-rails/actions/runs/3336212627/jobs/5523226123)

### DIFF
--- a/rspec-rails.gemspec
+++ b/rspec-rails.gemspec
@@ -46,12 +46,12 @@ Gem::Specification.new do |s|
   # get released.
   %w[core expectations mocks support].each do |name|
     if ENV['RSPEC_CI']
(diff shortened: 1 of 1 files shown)


JonRowe:
Main should use 13.pre @pirj

pirj:
@JonRowe Sorry, I don't understand what you mean. Is this change sufficient?
`main` is `6.1.0.pre`, and the gemspec:
```ruby
    elsif RSpec::Rails::Version::STRING =~ /pre/ # prerelease builds
      expected_rspec_version = "3.13.0.pre"
```
So, it appears to me that `main` is using 3.13.pre as you say.

JonRowe:
My bad I misread this last night, good job 👍

PR #2638, opened 2022-11-17
Fix action mailer check

tycooon:
In some cases, `ActionMailer` migth be defined, but `ActionMailer::TestCase` not.

### DIFF
--- a/lib/rspec/rails/example/mailer_example_group.rb
+++ b/lib/rspec/rails/example/mailer_example_group.rb
@@ -10,7 +10,7 @@ module RSpec
   end
 end
 
-if defined?(ActionMailer)
+if defined?(ActionMailer::TestCase)
   module RSpec
     module Rails
       # Container module for mailer spec functionality.


JonRowe:
Can you provide a reproduction for this?

tycooon:
For example, some app or gem might do `require "action_mailer/log_subscriber"`. In this case, they will get the following error: 

```
Failure/Error: require "rspec/rails"

NameError:
  uninitialized constant ActionMailer::TestCase

          include ActionMailer::TestCase::Behavior
```

I can create a sample app if that is needed.

JonRowe:
🤔 In that scenario they should check to see if ActionMailer is turned on with a similar defined? check to us, or require the whole gem. Howver we could add a `require "action_mailer/test_case"` as well

tycooon:
I tried requiring `action_mailer/test_case`, but after that i got other error:

```
NameError:
  uninitialized constant ActionMailer::TestCase::Behavior::TestHelper

        include TestHelper
```

tycooon:
Also by the way I might have my own `ActionMailer` module that is totally different from the Rails one. So the idea is that if we need `ActionMailer::TestCase`, we better check for its existence 🤔

JonRowe:
> Also by the way I might have my own ActionMailer module that is totally different from the Rails one.

We subscribe to the principle that we follow Ruby semantics and don't protect from people violating those semantics, e.g that you should not shadow someone elses module names, (we also couldn't feasibly check for that).

tycooon:
It's the correct module, so I don't think this will help :(

tycooon:
Actually I have real `ActionMailer`. The problem is, I only require `action_mailer/log_subscriber` and not `action_mailer`. So I have a top-level constant but I don't have `ActionMailer::TestCase`. Are you saying that `defined?`check will not work with autoloading whereas `include ActionMailer::TestCase::Behavior` will work? That's unfortunate.

pirj:
That's correct, `defined?` won't trigger autoloading, and `include` will.

If you don't require `action_mailer`, which defines autoloading of nested constants, autoloading won't work, which you already observe.

I think we've had [a similar problem with Active Record](https://github.com/rspec/rspec-rails/issues/1876), but I'm opposed to introducing `RSpec.configuration.use_action_mailer = false` just for this case, if there might be another way around.

tycooon:
I see, I guess I will just close the PR then. Thank you!

PR #2642, opened 2022-12-03
Use slash-separated path in Rails.root.join

### LINKED ISSUE #2641: Wouldn't it be preferable to use slashes in Rails.root.join?
In Pull Request #2640, it was commented to use multiple strings instead of slashes for `Rails.root.join`, but rubocop-rails defaults to using slashes.
- https://github.com/rspec/rspec-rails/pull/2640#discussion_r1029075106
- https://docs.rubocop.org/rubocop-rails/cops_rails.html#railsfilepath

This may be in consideration of platforms that do not use `/`, such as Windows, but since `Rails.root.join` is joined with `File::SEPARATOR (/)`, both writing styles return the same result separated by slashes.

The Windows path separator can also be interpreted, so it will work as intended but will not be significant. Wouldn't it be preferable to adopt the rubocop-rails default and write `Rails.root.join('foo/bar/baz')` for the same behavior? (Otherwise rubocop will still warn)

### Environment
- rails (7.0.4)
- rspec-core (3.12.0)
- rspec-rails (6.0.1)
- rubocop-rspec (2.15.0)
- rubocop-rails (2.17.3)

### Behavior of Rails.root.join under Windows
```
C:\Users\test\sample>systeminfo
...
OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.19044 N/A Build 19044
...

C:\Users\test\sample>ruby -v
ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x64-mingw-ucrt]

C:\Users\test\sample>rails c
Loading development environment (Rails 7.0.4)

irb(main):001:0> p Rails.root
#<Pathname:C:/Users/test/sample>
=> #<Pathname:C:/Users/test/sample>
irb(main):002:0> p Rails.root.join('foo/bar')
#<Pathname:C:/Users/test/sample/foo/bar>
=> #<Pathname:C:/Users/test/sample/foo/bar>
irb(main):003:0> p Rails.root.join('foo', 'bar')
#<Pathname:C:/Users/test/sample/foo/bar>
=> #<Pathname:C:/Users/test/sample/foo/bar>
```

### Windows path separators can be mixed (and still work fine)
```
irb(main):004:0> p Pathname.new(ENV['TMP']).join('foo', 'bar')
#<Pathname:C:\Users\test\AppData\Local\Temp/foo/bar>
=> #<Pathname:C:\Users\test\AppData\Local\Temp/foo/bar>
```

### Result for rubocop
```
$ bundle exec rubocop spec/rails_helper.rb
Inspecting 1 file
C

Offenses:

spec/rails_helper.rb:36:25: C: Rails/FilePath: Prefer Rails.root.join('path/to').
  config.fixture_path = Rails.root.join('spec', 'fixtures')
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected
```

castor4bit:
Fixes #2641

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -34,7 +34,7 @@ end
 RSpec.configure do |config|
 <% if RSpec::Rails::FeatureCheck.has_active_record? -%>
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
-  config.fixture_path = Rails.root.join('spec', 'fixtures')
+  config.fixture_path = Rails.root.join('spec/fixtures')
 
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
@@ -52,7 +52,7 @@ RSpec.configure do |config|
   # note if you'd prefer not to run each example within a transaction, you
   # should set use_transactional_fixtures to false.
   #
-  # config.fixture_path = Rails.root.join('spec', 'fixtures')
+  # config.fixture_path = Rails.root.join('spec/fixtures')
   # config.use_transactional_fixtures = true
 
 <% end -%>


pirj:
Thank you!

PR #2663, opened 2023-04-03
Fix RuboCop offenses in rails_helper.rb template

kengallego:
Rubocop:

```Rails/FilePath: Prefer Rails.root.join('path/to').```
~```Lint/RedundantDirGlobSort: Remove redundant sort.```~

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -20,7 +20,7 @@ require 'rspec/rails'
 # directory. Alternatively, in the individual `*_spec.rb` files, manually
 # require only the support files necessary.
 #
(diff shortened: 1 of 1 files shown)


pirj:
For the reference of `join`: https://github.com/rubocop/rubocop-rails/pull/128

Regarding redundant sort, are you confident about [this](https://github.com/rubocop/rubocop/blob/4b4c2a409b14908c4344995720e04a0c65dc55e4/lib/rubocop/cop/lint/redundant_dir_glob_sort.rb#L10)?
>       #   This cop is unsafe, in case of having a file and a directory with
>       #   identical names, since directory will be loaded before the file, which
>       #   will break `exe/files.rb` that rely on `exe.rb` file.

kengallego:
@pirj Good point about the `sort`. And thank you for the reference.

PR #2664, opened 2023-04-06
Fix deprecated TestFixtures#fixture_path call

### LINKED ISSUE #2661: DEPRECATION WARNING: TestFixtures#fixture_path is deprecated  and will be removed in Rails 7.2
FYI: Getting 
```
DEPRECATION WARNING: TestFixtures#fixture_path is deprecated 
and will be removed in Rails 7.2. Use #fixture_paths instead. (called from block in 
<module:FixtureSupport> at /usr/local/rvm/gems/ruby-3.2.1@ruby3.2-railsedge/
gems/rspec-rails-6.0.1/lib/rspec/rails/fixture_support.rb:24)
```
while running "rake" command.

nsimmons:
Fixes this error message from showing up when running Rails edge:

```
DEPRECATION WARNING: TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2. Use #fixture_paths instead. (called from block in <module:FixtureSupport> at /Users/nsimmons/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/rspec-rails-6.0.1/lib/rspec/rails/fixture_support.rb:24)
```

Closes #2661

### DIFF
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -164,8 +164,14 @@ RSpec.describe "Configuration" do
 
       group = RSpec.describe("Arbitrary Description", :use_fixtures)
 
-      expect(group).to respond_to(:fixture_path)
-      expect(group.fixture_path).to eq("custom/path")
+      if ::Rails::VERSION::MAJOR < 7
+        expect(group).to respond_to(:fixture_path)
+        expect(group.fixture_path).to eq("custom/path")
+      else
+        expect(group).to respond_to(:fixture_paths)
+        expect(group.fixture_paths).to include("custom/path")
+      end
+
       expect(group.new.respond_to?(:foo, true)).to be(true)
     end
   end
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -164,7 +164,7 @@ RSpec.describe "Configuration" do
 
       group = RSpec.describe("Arbitrary Description", :use_fixtures)
 
-      if ::Rails::VERSION::MAJOR < 7
+      if ::Rails::VERSION::STRING < '7.1.0'
         expect(group).to respond_to(:fixture_path)
         expect(group.fixture_path).to eq("custom/path")
       else
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -21,7 +21,12 @@ module RSpec
           if RSpec.configuration.use_active_record?
             include Fixtures
 
-            self.fixture_path = RSpec.configuration.fixture_path
+            # TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
+            if respond_to?(:fixture_paths=)
+              fixture_paths << RSpec.configuration.fixture_path
+            else
+              self.fixture_path = RSpec.configuration.fixture_path
+            end
             self.use_transactional_tests = RSpec.configuration.use_transactional_fixtures
             self.use_instantiated_fixtures = RSpec.configuration.use_instantiated_fixtures
 


Velisarius (approved):
I like it!

pirj:
`Rails: main` is deceivingly green, but actually has this:
```
DEPRECATION WARNING: TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2. Use #fixture_paths instead.
If multiple fixture paths have been configured with #fixture_paths, then #fixture_path will just return
the first path.
 (called from block (3 levels) in <top (required)> at /home/runner/work/rspec-rails/rspec-rails/spec/rspec/rails/configuration_spec.rb:168)
```
Where is it coming from?

I wonder if we should also add a setting that would allow multiple fixture paths.

jguecaimburu:
@pirj the `group` variable defined in the example inherits from `ActiveRecord::TestFixtures`, so in this line https://github.com/rspec/rspec-rails/blob/main/spec/rspec/rails/configuration_spec.rb#L168 the deprecated method gets called:
```rb
expect(group).to respond_to(:fixture_path)
expect(group.fixture_path).to eq("custom/path")
```

Maybe it could be tested like this:
```rb
if ActiveRecord::TestFixtures.respond_to?(:fixture_paths=)
  expect(group).to respond_to(:fixture_paths)
  expect(group.fixture_paths).to include("custom/path")
else
  expect(group).to respond_to(:fixture_path)
  expect(group.fixture_path).to eq("custom/path")
end
```
Not sure if there's a better way to get around it if we wanna keep the `respond_to` expectation.

> I wonder if we should also add a setting that would allow multiple fixture paths.
>
It makes sense to me. I also wonder if we should remove Rails default fixture folder from the array. With this change we'll have something like:
```rb
if self.respond_to?(:fixture_paths=)
  self.fixture_paths << RSpec.configuration.fixture_path
  # ["test/fixtures", "rspec/custom/path"]
else
  self.fixture_path = RSpec.configuration.fixture_path
  # "rspec/custom/path"
end
```
I would prefer to replicate the Rails idea of having a default `"spec/fixtures"` folder, then the developer could add extra paths in the config. But we could force `["spec/fixtures"]` into the config generator as well.

Here's the commit where `fixture_path` was deprecated: https://github.com/rails/rails/commit/6902cbce1b3ea7babab588171d413b9b78cb3f0f

mathieujobin (approved):

jguecaimburu:
> I suggest reducing the scope of this PR to just fixing the deprecation. We can implement fixture_paths= in our configuration at a later stage.

Yeah, makes sense. I'll open a PR after this one is merged.

JonRowe (changes requested):
This needs to add a spec / fix the existing spec that still emits a deprecation, I hope to merge https://github.com/rspec/rspec-support/pull/575 at some point which will handle this better as currently these deprecations don't trip our warning detection but just fail outright.

JonRowe:
This also triggers a rubocop rule because `self.fixture_paths <<` doesn't need a `self` like `self.fixture_path =`:

```
lib/rspec/rails/fixture_support.rb:25:16: C: [Correctable] Style/RedundantSelf: Redundant self detected.
            if self.respond_to?(:fixture_paths=)
               ^^^^
lib/rspec/rails/fixture_support.rb:26:15: C: [Correctable] Style/RedundantSelf: Redundant self detected.
              self.fixture_paths << RSpec.configuration.fixture_path
              ^^^^
```

nsimmons:
Corrected Rubocop issues and fixed the spec that was emitting the deprecation message.

pirj:
Re-running jobs failed with "marshal data too short" after merging the fix in https://github.com/rspec/rspec-support/pull/575

spec/rspec/rails/configuration_spec.rb
```diff
@@ -164,8 +164,14 @@ def in_inferring_type_from_location_environment
 
       group = RSpec.describe("Arbitrary Description", :use_fixtures)
 
-      expect(group).to respond_to(:fixture_path)
-      expect(group.fixture_path).to eq("custom/path")
+      if ::Rails::VERSION::MAJOR < 7
```
pirj, line 167: I think `fixture_paths` were only introduced on the edge, as they are [missing in 7.0.4.3](https://github.com/rails/rails/blob/v7.0.4.3/activerecord/lib/active_record/test_fixtures.rb#L20). 
```suggestion
      if ::Rails::VERSION::STRING < '7.1.0'
```
Should work fine for `7.0.4.3`, `7.1.0-alpha` and `7.1.0`.

pirj:
> This needs to add a spec / fix the existing spec that still emits a deprecation

@JonRowe This change fixed those [multiple `Rails: main` failures](https://github.com/rspec/rspec-rails/actions/runs/4740141301/jobs/8415648472?pr=2573):
```

  299) RSpec::Rails::FixtureFileUploadSupport with fixture path set in config resolves fixture file
       Failure/Error: example.run

       RuntimeError:
         Warnings were generated: DEPRECATION WARNING: TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2. Use #fixture_paths instead. (called from block in <module:FixtureSupport> at /home/runner/work/rspec-rails/rspec-rails/lib/rspec/rails/fixture_support.rb:24)
```

There's just one unrelated spec failure left on `Rails: main`:
```
Failures:

  1) be_valid matcher includes the error messages in the failure message
     Failure/Error:
       expect {
         expect(post).to be_valid
       }.to raise_exception(/Title can't be blank/)
```
that I've fixed in https://github.com/rspec/rspec-rails/pull/2669

Is this good to merge?

JonRowe:
Released in 6.0.2

jasnow:
@JonRowe - Thanks for the release.

PR #2665, opened 2023-04-14
fix deprecation failure with rails 7.1

mathieujobin:
DEPRECATION WARNING: TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2. Use #fixture_paths instead. (called from block in <module:FixtureSupport> at /home/mathieu/.asdf/installs/ruby/3.0.6/lib/ruby/gems/3.0.0/bundler/gems/rspec-rails-70984c1ab6a2/lib/rspec/rails/fixture_support.rb:24)


mathieujobin:
#2664 is better

PR #2666, opened 2023-04-14
Fix the order

pirj:
"default" is not an allowed value for order, and started failing after https://github.com/rspec/rspec-core/pull/3025


JonRowe:
Thanks I'd been meaning to get to that

PR #2670, opened 2023-04-19
Implement send_email matcher

ka8725:
Currently, it already has a `have_enqueued_mail` matcher, but it's too low-level, requiring attention to details such as job name, mailer name, queue, and so on. While all of that information can be useful, from a business perspective, it's rarely necessary. This suggested matcher aims to improve the situation so that when you use it, you can focus on what matters, such as 'an email with a specific subject and specific text within the body being sent'.

Examples:

```
@example Positive expectation
  expect { action }.to send_email

@example Negative expectations
  expect { action }.not_to send_email

@example More precise expectation with attributes to match
  expect { action }.to send_email(to: 'test@example.com', subject: 'Confirm email')
```

Originally that was a gist: https://gist.github.com/ka8725/6053d55be74f15b53cabfc3ac4dc99df

A related LinkedIn post - https://www.linkedin.com/feed/update/urn:li:activity:6990304469477429248?utm_source=share&utm_medium=member_desktop

### TODO

- [x] find a way to surface better the multiple email failure case especially as if you don't match say, subject, you could easily trigger it

### DIFF
--- a/features/matchers/send_email_matcher.feature
+++ b/features/matchers/send_email_matcher.feature
@@ -1,0 +1,68 @@
+Feature: `send_email` matcher
+
+  The `send_email` matcher is used to check if an email with the given parameters has been sent inside the expectation block.
+
+  NOTE: It implies that the spec example actually sends the email using the test adapter and does not schedule it for background.
+
+  To have an email sent in tests make sure:
+  - `ActionMailer` performs deliveries - `Rails.application.config.action_mailer.perform_deliveries = true`
+  - If the email is sent asynchronously (with `.deliver_later` call), ActiveJob uses the inline adapter - `Rails.application.config.active_job.queue_adapter = :inline`
+  - ActionMailer uses the test adapter - `Rails.application.config.action_mailer.delivery_method = :test`
+
+  If you want to check an email has been scheduled for background, use the `have_enqueued_email` matcher.
+
+  Scenario: Checking email sent with the given multiple parameters
+    Given a file named "spec/mailers/notifications_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "checks email sending by multiple params" do
+          expect {
+            NotificationsMailer.signup.deliver_now
+          }.to send_email(
+            from: 'from@example.com',
+            to: 'to@example.org',
+            subject: 'Signup'
+          )
+        end
+      end
+      """
+    When I run `rspec spec/mailers/notifications_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking email sent with matching parameters
+    Given a file named "spec/mailers/notifications_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "checks email sending by one param only" do
+          expect {
+            NotificationsMailer.signup.deliver_now
+          }.to send_email(
+            to: 'to@example.org'
+          )
+        end
+      end
+      """
+    When I run `rspec spec/mailers/notifications_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking email not sent with the given parameters
+    Given a file named "spec/mailers/notifications_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "checks email not sent" do
+          expect {
+            NotificationsMailer.signup.deliver_now
+          }.to_not send_email(
+            to: 'no@example.org'
+          )
+        end
+      end
+      """
+    When I run `rspec spec/mailers/notifications_mailer_spec.rb`
+    Then the examples should all pass
--- a/lib/rspec/rails/matchers/send_email.rb
+++ b/lib/rspec/rails/matchers/send_email.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+module RSpec
+  module Rails
+    module Matchers
+      # @api private
+      #
+      # Matcher class for `send_email`. Should not be instantiated directly.
+      #
+      # @see RSpec::Rails::Matchers#send_email
+      class SendEmail < RSpec::Rails::Matchers::BaseMatcher
+        # @api private
+        # Define the email attributes that should be included in the inspection output.
+        INSPECT_EMAIL_ATTRIBUTES = %i[subject from to cc bcc].freeze
+
+        def initialize(criteria)
+          @criteria = criteria
+        end
+
+        # @api private
+        def supports_value_expectations?
+          false
+        end
+
+        # @api private
+        def supports_block_expectations?
+          true
+        end
+
+        def matches?(block)
+          define_matched_emails(block)
+
+          @matched_emails.one?
+        end
+
+        # @api private
+        # @return [String]
+        def failure_message
+          result =
+            if multiple_match?
+              "More than 1 matching emails were sent."
+            else
+              "No matching emails were sent."
+            end
+          "#{result}#{sent_emails_message}"
+        end
+
+        # @api private
+        # @return [String]
+        def failure_message_when_negated
+          "Expected not to send an email but it was sent."
+        end
+
+        private
+
+        def diffable?
+          true
+        end
+
+        def deliveries
+          ActionMailer::Base.deliveries
+        end
+
+        def define_matched_emails(block)
+          before = deliveries.dup
+
+          block.call
+
+          after = deliveries
+
+          @diff = after - before
+          @matched_emails = @diff.select(&method(:matched_email?))
+        end
+
+        def matched_email?(email)
+          @criteria.all? do |attr, value|
+            expected =
+              case attr
+              when :to, :from, :cc, :bcc then Array(value)
+              else
+                value
+              end
+
+            values_match?(expected, email.public_send(attr))
+          end
+        end
+
+        def multiple_match?
+          @matched_emails.many?
+        end
+
+        def sent_emails_message
+          if @diff.empty?
+            "\n\nThere were no any emails sent inside the expectation block."
+          else
+            sent_emails =
+              @diff.map do |email|
+                inspected = INSPECT_EMAIL_ATTRIBUTES.map { |attr| "#{attr}: #{email.public_send(attr)}" }.join(", ")
+                "- #{inspected}"
+              end.join("\n")
+            "\n\nThe following emails were sent:\n#{sent_emails}"
+          end
+        end
+      end
+
+      # @api public
+      # Check email sending with specific parameters.
+      #
+      # @example Positive expectation
+      #   expect { action }.to send_email
+      #
+      # @example Negative expectations
+      #   expect { action }.not_to send_email
+      #
+      # @example More precise expectation with attributes to match
+      #   expect { action }.to send_email(to: 'test@example.com', subject: 'Confirm email')
+      def send_email(criteria = {})
+        SendEmail.new(criteria)
+      end
+    end
+  end
+end
--- a/spec/rspec/rails/matchers/send_email_spec.rb
+++ b/spec/rspec/rails/matchers/send_email_spec.rb
@@ -1,0 +1,168 @@
+RSpec.describe "send_email" do
+  let(:mailer) do
+    Class.new(ActionMailer::Base) do
+      self.delivery_method = :test
+
+      def test_email
+        mail(
+          from: "from@example.com",
+          cc: "cc@example.com",
+          bcc: "bcc@example.com",
+          to: "to@example.com",
+          subject: "Test email",
+          body: "Test email body"
+        )
+      end
+    end
+  end
+
+  it "checks email sending by all params together" do
+    expect {
+      mailer.test_email.deliver_now
+    }.to send_email(
+      from: "from@example.com",
+      to: "to@example.com",
+      cc: "cc@example.com",
+      bcc: "bcc@example.com",
+      subject: "Test email",
+      body: a_string_including("Test email body")
+    )
+  end
+
+  it "checks email sending by no params" do
+    expect {
+      mailer.test_email.deliver_now
+    }.to send_email
+  end
+
+  it "with to_not" do
+    expect {
+      mailer.test_email.deliver_now
+    }.to_not send_email(
+      from: "failed@example.com"
+    )
+  end
+
+  it "fails with a clear message" do
+    expect {
+      expect { mailer.test_email.deliver_now }.to send_email(from: 'failed@example.com')
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+      No matching emails were sent.
+
+      The following emails were sent:
+      - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+    MSG
+  end
+
+  it "fails with a clear message when no emails were sent" do
+    expect {
+      expect { }.to send_email
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+      No matching emails were sent.
+
+      There were no any emails sent inside the expectation block.
+    MSG
+  end
+
+  it "fails with a clear message for negated version" do
+    expect {
+      expect { mailer.test_email.deliver_now }.to_not send_email(from: "from@example.com")
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, "Expected not to send an email but it was sent.")
+  end
+
+  it "fails for multiple matches" do
+    expect {
+      expect { 2.times { mailer.test_email.deliver_now } }.to send_email(from: "from@example.com")
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+      More than 1 matching emails were sent.
+
+      The following emails were sent:
+      - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+      - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+    MSG
+  end
+
+  context "with compound matching" do
+    it "works when both matchings pass" do
+      expect {
+        expect {
+          mailer.test_email.deliver_now
+        }.to send_email(to: "to@example.com").and send_email(from: "from@example.com")
+      }.to_not raise_error
+    end
+
+    it "works when first matching fails" do
+      expect {
+        expect {
+          mailer.test_email.deliver_now
+        }.to send_email(to: "no@example.com").and send_email(to: "to@example.com")
+      }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+        No matching emails were sent.
+
+        The following emails were sent:
+        - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+      MSG
+    end
+
+    it "works when second matching fails" do
+      expect {
+        expect {
+          mailer.test_email.deliver_now
+        }.to send_email(to: "to@example.com").and send_email(to: "no@example.com")
+      }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+        No matching emails were sent.
+
+        The following emails were sent:
+        - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+      MSG
+    end
+  end
+
+  context "with a custom negated version defined" do
+    define_negated_matcher :not_send_email, :send_email
+
+    it "works with a negated version" do
+      expect {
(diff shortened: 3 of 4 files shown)


spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    expect {
+      expect { TestMailer.test_email.deliver_now }.to send_email(from: 'failed@example.com')
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, "No matching emails were sent.")
+  end
+
+  it "fails with a clear message for negated version" do
+    expect {
+      expect { TestMailer.test_email.deliver_now }.to_not send_email(from: 'from@example.com')
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, "Expected not to send an email but it was sent.")
+  end
```
pirj, line 31: Need an example to cover 'More than 1 matching emails were sent.'

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,109 @@
...
+              result[:to] = Array(init_criteria[:to]) if init_criteria.key?(:to)
+              result[:from] = Array(init_criteria[:from]) if init_criteria.key?(:from)
+              result
+            end
+        end
+
+        def init_criteria
+          @init_criteria ||= {}
+        end
+
```
pirj, line 87:
```suggestion
```

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,109 @@
...
+        def matched_email?(email)
+          criteria.all? do |attr, expected|
+            values_match?(expected, email.public_send(attr))
+          end
+        end
+
+        def criteria
+          @criteria ||=
+            begin
+              result = init_criteria
```
pirj, line 77:
```suggestion
              result = {}
```

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,109 @@
...
+        def failure_message_when_negated
+          "Expected not to send an email but it was sent."
+        end
+
+        def deliveries
+          ActionMailer::Base.deliveries.clone
+        end
+
+        def define_matched_emails(block)
+          before = deliveries
```
pirj, line 54:
```suggestion
          before = deliveries.clone
```
to clone just once

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,109 @@
...
+        def supports_block_expectations?
+          true
+        end
+
+        def matches?(block)
+          raise ArgumentError, 'send_email only work with block arguments' unless block.respond_to?(:call)
+
+          define_matched_emails(block)
+
+          @matched_emails.size == 1
```
pirj, line 23:
```suggestion
          @matched_emails.one?
```

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,109 @@
...
+          criteria.empty? ? emails : emails.select(&method(:matched_email?))
+        end
+
+        def matched_email?(email)
+          criteria.all? do |attr, expected|
+            values_match?(expected, email.public_send(attr))
+          end
+        end
+
+        def criteria
```
pirj, line 74: can this be done in `initialize`?

pirj (commented):
Nice!

pirj:
A feature spec would be helpful to make this more discoverable.

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,85 @@
...
+  If you want to check an email has been scheduled for background, use the `have_enqueued_email` matcher.
+
+  Scenario: Checking email sent with the given multiple parameters
+    Given a file named "spec/mailers/notifications_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "checks email sending by multiple params" do
+
```
JonRowe, line 19: Remove white space

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,85 @@
+Feature: `send_email` matcher
+
+  The `send_email` matcher is used to check if an email with the given parameters has been sent during a spec example run.
+
+  NOTE: It implies that the spec example actually sends the email using the test adapter and does not schedule it for background.
+  To have an email sent in tests make sure:
+  - `ActionMailer` performs deliveries - `Rails.application.config.action_mailer.perform_deliveries = true`
+  - If the email is sent asynchronously (with `.deliver_later` call), ActiveJob uses the inline adapter - `Rails.application.config.active_job.queue_adapter = :inline`
+  - ActionMailer uses the test adapter - `Rails.application.config.action_mailer.delivery_method = :test`
+  If you want to check an email has been scheduled for background, use the `have_enqueued_email` matcher.
```
JonRowe, line 10: These get rendered as part of our docs when published, so can you break this up for readability?

```suggestion
  NOTE: It implies that the spec example actually sends the email using the test adapter and does not schedule it for background.
  
  To have an email sent in tests make sure:
  - `ActionMailer` performs deliveries - `Rails.application.config.action_mailer.perform_deliveries = true`
  - If the email is sent asynchronously (with `.deliver_later` call), ActiveJob uses the inline adapter - `Rails.application.config.active_job.queue_adapter = :inline`
  - ActionMailer uses the test adapter - `Rails.application.config.action_mailer.delivery_method = :test`
  
  If you want to check an email has been scheduled for background, use the `have_enqueued_email` matcher.
```

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,85 @@
...
+            to: 'to@example.org',
+            subject: 'Signup'
+          )
+        end
+      end
+      """
+    When I run `rspec spec/mailers/notifications_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking email sent with one parameter
```
JonRowe, line 33:
```suggestion
  Scenario: Checking email sent with matching parameters
```

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,100 @@
...
+        def failure_message_when_negated
+          "Expected not to send an email but it was sent."
+        end
+
+        def deliveries
+          ActionMailer::Base.deliveries
+        end
+
+        def define_matched_emails(block)
+          before = deliveries.clone
```
JonRowe, line 59: Why `clone` and not `dup`?

spec/support/mailers.rb
```diff
@@ -0,0 +1,48 @@
+ActionMailer::Base.delivery_method = :test
+
+class GlobalIDArgument
```
JonRowe, line 3: If you're moving these out of the spec can you nest them in a module to take them off the top level, e.g. `Support::Mailers`, alternatively use shared examples to define them in both specs

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,100 @@
...
+    module Matchers
+      # @api private
+      #
+      # Matcher class for `send_email`. Should not be instantiated directly.
+      #
+      # @see RSpec::Rails::Matchers#send_email
+      class SendEmail < RSpec::Rails::Matchers::BaseMatcher
+        def initialize(criteria)
+          normalized = %i[to from].map { |attr| criteria.key?(attr) ? [attr, Array(criteria[attr])] : nil }.compact.to_h
+          @criteria = { **criteria, **normalized }
```
JonRowe, line 12: I think this would be more readable if you did this check at match time e.g. (and see: )
```suggestion
          @criteria = criteria
```

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,100 @@
...
+          @matched_emails = select_matching(@diff)
+        end
+
+        def select_matching(emails)
+          @criteria.empty? ? emails : emails.select(&method(:matched_email?))
+        end
+
+        def matched_email?(email)
+          @criteria.all? do |attr, expected|
+            values_match?(expected, email.public_send(attr))
```
JonRowe, line 75: Continued from above:
```suggestion
        def matched_email?(email)
          @criteria.all? do |attr, value|
            expected =
              case attr
              # :to / :from are arrays
              when :to,:from then Array(value)
              else
                value
              end

            values_match?(expected, email.public_send(attr))
```

JonRowe (changes requested):

JonRowe:
Thanks for this, it mostly looks great just some readability tweaks, I'm also wondering if theres a way to surface better the multiple email failure case especially as if you don't match say, subject, you could easily trigger it

spec/support/shared_contexts.rb
```diff
@@ -0,0 +1,9 @@
+RSpec.shared_context "test mailer" do
+  class TestMailer < ActionMailer::Base
```
ka8725, line 2: I don't like this idea actually, Rubocop is right - https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/ConstantDefinitionInBlock

Defining classes with let also complicates things.

Classes should be defined within modules. Moving it out to a shared module.

spec/support/mailers.rb
ka8725: I tried to use shared contexts, but Rubocop has complaints about that - https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/ConstantDefinitionInBlock

I agree that it should not define classes within blocks. I also don't like defining them within `let` blocks as it complicates things.

I would go with the current solution without surrounding these classes with modules. And this is why:
1. The previous implementation would define these classes globally anyways. Yes, that would be conditionally, but **globally** - once runtime comes to `spec/rspec/rails/matchers/have_enqueued_mail_spec.rb` it defines these classes and they are available globally by all other specs. So moving this code into the "support" folder as is doesn't change anything from the runtime point of view, but actually fixes potential issues with the surprising dynamic appearance of these classes in the code base.
2. Namespacing would need to either rewrite many specs to use the full path to classes, e.g. instead of `TestMailer` it needs `Support::Mailers::TestMailer`. An `include Support::Mailers` on the top of the spec file would solve this point, but that would define class names globally dynamically again. An alternative solution could be redefining the class name (`def self.name`) for these classes, but that would look weird in general and like an over-engineering.
3. The codebase already has an example of similar classes defined globally - see `support/ar_classes.rb`

So, looks like defining these classes in namespaces would just complicate things and that's unnecessary. WDYT? Any specific reason for defining them within a namespace?

ka8725:
@JonRowe to address this point:

> find a way to surface better the multiple email failure case especially as if you don't match say, subject, you could easily trigger it

I have an idea to inspect the sent emails. https://github.com/rspec/rspec-rails/pull/2670/commits/0e25b6605d8aa72f875aa4b8fe41f6b4e990bb00 That should give a clue for devs where the failure comes from. Ideally, that would be great to inspect the body as well, but the output can be very huge. Therefore, I would better not do that. Basically, having just the subject inspected would suffice, but the other important attributes like from/to not hurt.

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,123 @@
...
+            if multiple_match?
+              'More than 1 matching emails were sent.'
+            else
+              'No matching emails were sent.'
+            end
+          "#{result}#{sent_emails_message}"
+        end
+
+        def failure_message_when_negated
+          "Expected not to send an email but it was sent."
```
avsej, line 54: Freeze string literals. You are doing it already for array `INSPECT_EMAIL_ATTRIBUTES`. Makes sense to do it here too. Or declare modeline to freeze them automatically.

ka8725: I see `freeze` is used occasionally in other places, so removing it from the const array declaration `INSPECT_EMAIL_ATTRIBUTES`. Freezing it or not is not very important, as it's highly unlikely that someone would mutate it. Freezing strings is not common in other similar places within this project. Therefore, this question is related to the project as a whole rather than just this pull request. The project uses Rubocop, and if configured properly, it would help eliminate this question as well as the one mentioned above.

JonRowe: Adding the magic freeze string literals to the top of the file is certainly fine. We have a very custom rubocop check list in part due to legacy history, a lot of our code predates Rubocop's existence. (Rubocop's RSpec rules are also not endorsed by us in anyway).

spec/support/mailers.rb
JonRowe: The reason for my original comment is I don't want to accidentally pollute our root namespace making it seem like these are real classes, as you're moving them and creating churn, might as well fix that.

lib/rspec/rails/matchers/send_email.rb
ka8725: Ok, changing to freezing literals. Rubocop can be configured to validate new files (the rest files can be added to "rubocop todo"), but it's a separate story.

spec/support/mailers.rb
ka8725: That causes too many questions that seem not related to the PR. I found another solution: I will define the mailer inside the new test using `let` as it should have been done in the beginning. The issue of polluting root space remains though. It should be resolved separately.

spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
```
pirj, line 2:
```suggestion
```
I've once tested on a huge codebase, no noticeable difference with or without this directive **in specs**.

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,86 @@
+Feature: `send_email` matcher
+
+  The `send_email` matcher is used to check if an email with the given parameters has been sent during a spec example run.
```
pirj, line 3: "during a spec example run" or "inside the expectation block"?

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,86 @@
...
+    When I run `rspec spec/mailers/notifications_mailer_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking email sent with a wrong parameter
+    Given a file named "spec/mailers/notifications_mailer_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe NotificationsMailer do
+        it "checks email not sent" do
```
pirj, line 76: This docstring doesn't match

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,125 @@
...
+          @matched_emails.one?
+        end
+
+        private
+
+        def diffable?
+          true
+        end
+
+        def match_when_negated(block)
```
pirj, line 40: Would it be possible to do this with `does_not_match?`?
```suggestion
        def does_not_match?(block)
```

I think of uses with compound matching like:
```ruby
expect {}
 .to send_email(to: 'foo@...')
 .and send_email(to: 'bar@...')
```
and if in certain cases when the first or the second matcher fails there would be a difference in the output.

spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    MSG
+  end
+
+  it "fails with a clear message when no emails were sent" do
+    expect {
+      expect { }.to send_email
+    }.to raise_error(RSpec::Expectations::ExpectationNotMetError, <<~MSG.strip)
+      No matching emails were sent.
+
+      No emails were sent.
```
pirj, line 65: Those two statements sound similar.

lib/rspec/rails/matchers/send_email.rb
```diff
@@ -0,0 +1,125 @@
...
+          block.call
+
+          after = deliveries
+
+          @diff = after - before
+          @matched_emails = select_matching(@diff)
+        end
+
+        def select_matching(emails)
+          @criteria.empty? ? emails : emails.select(&method(:matched_email?))
```
pirj, line 75: `.empty?` check seems redundant
```ruby
[].all? { true } # => true
[].all? { false } # => true
```

```suggestion
          emails.select(&method(:matched_email?))
```

pirj (approved):
Looks great!
A few nitpicks below.

lib/rspec/rails/matchers/send_email.rb
ka8725: Good catch! Actually, the compound version didn't work at all, because the `failure_message` defined below should be public.

I didn't see the difference between the usage of `match_when_negated` and `does_not_match?`. At least my tests did not expose any difference. Moreover, I realized the default implementation of `does_not_match?` is good enough. So removing it.

Another fun thing is that `failure_message_when_negated` can be private while `failure_message` should be public. But I move them both to public space for consistency as the other matchers do.

features/matchers/send_email_matcher.feature
ka8725: Initially, I added that scenario for integration testing purposes to check a failed scenario. But no I see how weird it sounds. It definitely can't be used for documentation purposes and misleads. So removing it. The unit tests cover that case and that's enough.

spec/rspec/rails/matchers/send_email_spec.rb
ka8725: Improved the message a bit.

spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+    )
+  end
+
+  it "checks email sending by no params" do
+    expect {
+      mailer.test_email.deliver_now
+    }.to send_email
+  end
+
+  it "with with to_not" do
```
pirj, line 38:
```suggestion
  it "with to_not" do
```

spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+        No matching emails were sent.
+
+        The following emails were sent:
+        - subject: Test email, from: ["from@example.com"], to: ["to@example.com"], cc: ["cc@example.com"], bcc: ["bcc@example.com"]
+      MSG
+    end
+  end
+
+  context "with a custom negated version defined" do
+    before { RSpec::Matchers.define_negated_matcher :not_send_email, :send_email }
```
pirj, line 122:
```suggestion
    RSpec::Matchers.define_negated_matcher :not_send_email, :send_email
```

pirj (approved):
Thank you so much for the idea and for implementing it!

spec/rspec/rails/matchers/send_email_spec.rb
ka8725: It also pollutes the global scope. Changing it to use `around`

spec/rspec/rails/matchers/send_email_spec.rb
```diff
@@ -119,7 +119,12 @@ def test_email
   end
 
   context "with a custom negated version defined" do
-    before { RSpec::Matchers.define_negated_matcher :not_send_email, :send_email }
+    around do |example|
+      RSpec::Matchers.define_negated_matcher :not_send_email, :send_email
+      example.run
+    ensure
+      undef not_send_email
+    end
```
pirj, line 127:
```suggestion
    define_negated_matcher :not_send_email, :send_email
```

ka8725:
@pirj can the failed CI job be restarted? looks like there was a network issue - https://github.com/rspec/rspec-rails/actions/runs/4790425603/jobs/8522781666?pr=2670

pirj:
Green. It slipped my mind, do we squash commits in meaningful ones before merging?

ka8725:
@pirj I've squashed them.

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,68 @@
+Feature: `send_email` matcher
+
+  The `send_email` matcher is used to check if an email with the given parameters has been sent inside the expectation block.
+
+  NOTE: It implies that the spec example actually sends the email using the test adapter and does not schedule it for background.
```
JonRowe, line 5:
```suggestion
  NOTE: This matcher implies that the example(s) actually send email(s) using the test adapter, and *not* scheduled as a job to be sent in the background later.
```

features/matchers/send_email_matcher.feature
```diff
@@ -0,0 +1,68 @@
...
+
+  To have an email sent in tests make sure:
+  - `ActionMailer` performs deliveries - `Rails.application.config.action_mailer.perform_deliveries = true`
+  - If the email is sent asynchronously (with `.deliver_later` call), ActiveJob uses the inline adapter - `Rails.application.config.active_job.queue_adapter = :inline`
+  - ActionMailer uses the test adapter - `Rails.application.config.action_mailer.delivery_method = :test`
+
+  If you want to check an email has been scheduled for background, use the `have_enqueued_email` matcher.
```
JonRowe, line 12:
```suggestion
  If you want to check an email has been scheduled as a job, use the `have_enqueued_email` matcher.
```

JonRowe (approved):
Just a couple of wording tweaks to the feature, and this is good to go, sorry for the delayed review.

JonRowe:
Merged in  69c09bc5fd56c3c24ee849cdf61cb313f59ab51e thanks for the hard work on this.

JonRowe:
Released in 6.1.0

PR #2671, opened 2023-04-19
Fixes issue with ActiveStorage::FixtureSet.blob causing a TypeError

### LINKED ISSUE #2637: ActiveStorage::FixtureSet.blob incompatible with rspec

Ruby version: 3.1.1
Rails version: 7.0.4
RSpec version:
rspec-core (3.12.0)
rspec-rails (5.0.3)
rspec-support (3.12.0)

## Observed behaviour

When setting up file fixtures, and calling e.g. `fixtures/active_storage/blobs.yml`
```
my_blob: <%= ActiveStorage::FixtureSet.blob(
   filename: "test.jpg",
) %>
```

This will call `ActiveSupport::Testing::FileFixtures.file_fixture_path`, which is nil. This then fails.

The rspec config should be set up correctly.
```
RSpec.configure do |config||
  config.fixture_path = "#{::Rails.root}/spec/fixtures"
  config.file_fixture_path = "#{::Rails.root}/spec/fixtures/files"
end
```

## Expected behaviour

Expect rspec to find file_fixture_path

jaywhy:
This PR fixes an issue where calling `ActiveStorage::FixtureSet.blob` will error with `TypeError: no implicit conversion of nil into String` because the `blob` method expects `ActiveStorage::FixtureSet.file_fixture_path` to be set.

This PR adds a default value to `ActiveStorage::FixtureSet.file_fixture_path`, fixing the issue. It checks to see if it is defined because this is a new feature of Rails 7.

More information can be found in  #2637.
fixes #2637

### DIFF
--- a/features/file_fixture.feature
+++ b/features/file_fixture.feature
@@ -36,3 +36,22 @@ Feature: Using `file_fixture`
       """
     When I run `rspec spec/lib/file_spec.rb`
     Then the examples should all pass
+
+  @rails_post_7
+  Scenario: Creating a ActiveStorage::Blob from a file fixture
+    Given a file named "spec/fixtures/files/sample.txt" with:
+      """
+      Hello
+      """
+    And a file named "spec/lib/fixture_set_blob.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe "blob" do
+        it "creates a blob from a sample file" do
+          expect(ActiveStorage::FixtureSet.blob(filename: "sample.txt")).to include("sample.txt")
+        end
+      end
+      """
+    When I run `rspec spec/lib/fixture_set_blob.rb`
+    Then the examples should all pass
--- a/features/support/rails_versions.rb
+++ b/features/support/rails_versions.rb
@@ -1,0 +1,15 @@
+def rails_version
+  string_version = ENV.fetch("RAILS_VERSION", "~> 7.0.0")
+  if string_version == "main" || string_version.nil?
+    Float::INFINITY
+  else
+    string_version[/\d[.-]\d/].tr('-', '.')
+  end
+end
+
+Before "@rails_post_7" do |scenario|
+  if rails_version.to_f < 7.0
+    warn "Skipping scenario #{scenario.name} on Rails v#{rails_version}"
+    skip_this_scenario
+  end
+end
--- a/lib/rspec/rails/file_fixture_support.rb
+++ b/lib/rspec/rails/file_fixture_support.rb
@@ -9,6 +9,9 @@ module RSpec
 
       included do
         self.file_fixture_path = RSpec.configuration.file_fixture_path
+        if defined?(ActiveStorage::FixtureSet)
+          ActiveStorage::FixtureSet.file_fixture_path = RSpec.configuration.file_fixture_path
+        end
       end
     end
   end


pirj:
Nice! Would it be possible to add a scenario to `features/file_fixture.feature`?

jaywhy:
@pirj I've added a scenario covering the issue.

features/file_fixture.feature
```diff
@@ -36,3 +36,21 @@ Feature: Using `file_fixture`
...
+      require "rails_helper"
+
+      RSpec.describe "blob" do
+        it "creates a blob from a sample file" do
+          expect(ActiveStorage::FixtureSet.blob(filename: "sample.txt")).to include("sample.txt")
+        end
+      end
+      """
+    When I run `rspec spec/lib/fixture_set_blob.rb`
+    Then the examples should all pass
```
JonRowe, line 56:
```suggestion
    Then the examples should all pass
```

JonRowe: Line ending issue, the above should fix it

pirj:
Some feature failed. Can you please have a look?

jaywhy:
@pirj  thanks for the heads up.

The test fails because `ActiveStorage::FixtureSet` is not defined in Rails 6. I put an `if` check around the actual code but didn't consider the test.

I added a check around the test. This should fix the issue.

features/file_fixture.feature
```diff
@@ -36,3 +36,23 @@ Feature: Using `file_fixture`
...
+      """
+      Hello
+      """
+    And a file named "spec/lib/fixture_set_blob.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe "blob" do
+        it "creates a blob from a sample file" do
+          if defined?(ActiveStorage::FixtureSet)
```
pirj, line 51: Let's skip this scenario altogether for pre-7 Rails like we [used to do here with pre-6 features](https://github.com/rspec/rspec-rails/commit/229f5f7c#diff-f57037e7ce9e7f58b722ba688654211d528c6c30ed7cd89df2cae8db9896ad14). That keeps it easier to read, and there's a visual note that this is only for Rails 7.

pirj (approved):
Thank you!

pirj:
It seems that RubyGems times out, and causes red ci.

features/file_fixture.feature
jaywhy: I see. I added the change.

pirj:
I don't understand why it won't let me force push.

pirj:
```
From 106cca999d2aa1afff5c478631045c35cf0380df Mon Sep 17 00:00:00 2001
From: Jason Yates <jason@jasonyates.me>
Date: Wed, 19 Apr 2023 14:27:26 -0400
Subject: [PATCH] fixup!

---
 features/file_fixture.feature      |  7 +++----
 features/support/rails_versions.rb | 15 +++++++++++++++
 2 files changed, 18 insertions(+), 4 deletions(-)
 create mode 100644 features/support/rails_versions.rb

diff --git features/file_fixture.feature features/file_fixture.feature
index 845c280c..c6a765a5 100644
--- features/file_fixture.feature
+++ features/file_fixture.feature
@@ -37,6 +37,7 @@ Feature: Using `file_fixture`
     When I run `rspec spec/lib/file_spec.rb`
     Then the examples should all pass
 
+  @rails_post_7
   Scenario: Creating a ActiveStorage::Blob from a file fixture
     Given a file named "spec/fixtures/files/sample.txt" with:
       """
@@ -48,11 +49,9 @@ Feature: Using `file_fixture`
 
       RSpec.describe "blob" do
         it "creates a blob from a sample file" do
-          if ::Rails::VERSION::STRING >= '7'
-            expect(ActiveStorage::FixtureSet.blob(filename: "sample.txt")).to include("sample.txt")
-          end
+          expect(ActiveStorage::FixtureSet.blob(filename: "sample.txt")).to include("sample.txt")
         end
       end
       """
     When I run `rspec spec/lib/fixture_set_blob.rb`
-    Then the examples should all pass
\ No newline at end of file
diff --git features/support/rails_versions.rb features/support/rails_versions.rb
new file mode 100644
index 00000000..3b2e3ef7
--- /dev/null
+++ features/support/rails_versions.rb
@@ -0,0 +1,15 @@
+def rails_version
+  string_version = ENV.fetch("RAILS_VERSION", "~> 7.0.0")
+  if string_version == "main" || string_version.nil?
+    Float::INFINITY
+  else
+    string_version[/\d[\.-]\d/].tr('-', '.')
+  end
+end
+
+Before "@rails_post_7" do |scenario|
+  if rails_version.to_f < 7.0
+    warn "Skipping scenario #{scenario.name} on Rails v#{rails_version}"
+    skip_this_scenario
+  end
+end
-- 
2.39.1
```

jaywhy:
@pirj I was able to add your patch. You're awesome 💯 👍

jaywhy:
I fixed the Rubocop issue. Hopefully that is everything 🤞.

PR #2673, opened 2023-04-20
Add fixture_paths configuration setting

jguecaimburu:
`ActiveRecord::TestFixtures#fixture_path` was [deprecated](https://github.com/rails/rails/commit/6902cbce1b3ea7babab588171d413b9b78cb3f0f) by Rails in 7.1 and will be removed from version 7.2.

https://github.com/rspec/rspec-rails/pull/2664 fixed the deprecation message. Here a new setting is added to the config to allow multiple paths.

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -36,6 +36,10 @@ RSpec.configure do |config|
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
 <% if ::Rails::VERSION::STRING < "7.1.0" -%>
   config.fixture_path = Rails.root.join('spec/fixtures')
+  # Starting with Rails 7.1, use this instead:
+  # config.fixture_paths = [
+  #  Rails.root.join('spec/fixtures')
+  # ]
 <% else -%>
   config.fixture_paths = [
     Rails.root.join('spec/fixtures')
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -34,7 +34,13 @@ end
 RSpec.configure do |config|
 <% if RSpec::Rails::FeatureCheck.has_active_record? -%>
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
+<% if ::Rails::VERSION::STRING < "7.1.0" -%>
   config.fixture_path = Rails.root.join('spec/fixtures')
+<% else -%>
+  config.fixture_paths = [
+    Rails.root.join('spec/fixtures')
+  ]
+<% end -%>
 
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -57,7 +57,7 @@ module RSpec
     end
 
     # @private
-    def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength,Metrics/CyclomaticComplexity
+    def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength,Metrics/CyclomaticComplexity,Metrics/AbcSize,Metrics/PerceivedComplexity
       config.backtrace_exclusion_patterns << /vendor\//
       config.backtrace_exclusion_patterns << %r{lib/rspec/rails}
 
@@ -69,7 +69,13 @@ module RSpec
       config.add_setting :use_transactional_fixtures, alias_with: :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
-      config.add_setting :fixture_path
+
+      if ::Rails::VERSION::STRING < "7.1.0"
+        config.add_setting :fixture_path
+      else
+        config.add_setting :fixture_paths
+      end
+
       config.include RSpec::Rails::FixtureSupport, :use_fixtures
 
       # We'll need to create a deprecated module in order to properly report to
@@ -156,6 +162,18 @@ module RSpec
           filter_gems_from_backtrace "actionmailer", "actionpack", "actionview"
           filter_gems_from_backtrace "activemodel", "activerecord",
                                      "activesupport", "activejob"
+        end
+
+        # @deprecated TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
+        if ::Rails::VERSION::STRING >= "7.1.0"
+          def config.fixture_path=(path)
+            RSpec.deprecate(
+              "config.fixture_path =  #{path.inspect}",
+              replacement: "config.fixture_paths = [#{path.inspect}]",
+              message: "Rails 7.1 has deprecated the singular fixture_path in favour of an array"
+            )
+            self.fixture_paths = Array(path)
+          end
         end
       end
 
--- a/lib/rspec/rails/configuration.rb
+++ b/lib/rspec/rails/configuration.rb
@@ -166,7 +166,7 @@ module RSpec
 
         # @deprecated TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
         if ::Rails::VERSION::STRING >= "7.1.0"
-          def config.fixture_path=(path)
+          def fixture_path=(path)
             RSpec.deprecate(
               "config.fixture_path =  #{path.inspect}",
               replacement: "config.fixture_paths = [#{path.inspect}]",
--- a/lib/rspec/rails/fixture_file_upload_support.rb
+++ b/lib/rspec/rails/fixture_file_upload_support.rb
@@ -14,6 +14,8 @@ module RSpec
         resolved_fixture_path =
           if respond_to?(:file_fixture_path) && !file_fixture_path.nil?
             file_fixture_path.to_s
+          elsif respond_to?(:fixture_paths)
+            (RSpec.configuration.fixture_paths&.first || '').to_s
           else
             (RSpec.configuration.fixture_path || '').to_s
           end
@@ -26,7 +28,11 @@ module RSpec
         include ActiveSupport::Testing::FileFixtures
 
         class << self
-          attr_accessor :fixture_path
+          if ::Rails::VERSION::STRING < "7.1.0"
+            attr_accessor :fixture_path
+          else
+            attr_accessor :fixture_paths
+          end
 
           # Get instance of wrapper
           def instance
--- a/lib/rspec/rails/fixture_support.rb
+++ b/lib/rspec/rails/fixture_support.rb
@@ -23,10 +23,11 @@ module RSpec
 
             # TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
             if respond_to?(:fixture_paths=)
-              fixture_paths << RSpec.configuration.fixture_path
+              self.fixture_paths = RSpec.configuration.fixture_paths
             else
               self.fixture_path = RSpec.configuration.fixture_path
             end
+
             self.use_transactional_tests = RSpec.configuration.use_transactional_fixtures
             self.use_instantiated_fixtures = RSpec.configuration.use_instantiated_fixtures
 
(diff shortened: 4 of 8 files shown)


lib/rspec/rails/configuration.rb
```diff
@@ -69,7 +69,12 @@ def self.initialize_configuration(config) # rubocop:disable Metrics/MethodLength
       config.add_setting :use_transactional_fixtures, alias_with: :use_transactional_examples
       config.add_setting :use_instantiated_fixtures
       config.add_setting :global_fixtures
+
+      config.add_setting :fixture_paths
+
+      # @deprecated TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
       config.add_setting :fixture_path
+
```
JonRowe, line 77: We shouldn't add the new setting on Rails that doesn't support it, and the implementation of this original method should take into account the existence of the new one:
```suggestion
      if ::Rails::VERSION::STRING < '7.1.0'
        config.add_setting :fixture_paths
        # psuedo coding here, this should probably be in the instance_exec below
        def config.fixture_path=(path)
          RSpec.deprecate(
            "config.fixture_path =  #{path.inspect}",
            replacement: "config.fixture_paths = [#{path.inspect}],
            message: "Rails 7.1 has deprecated the singular fixture_path in favour of an array"
          )

          self.fixture_paths = [path]
        end
      else
        config.add_setting :fixture_path
      end
```
The message is optional for deprecation, and could be reworded

JonRowe (dismissed):
Thanks for this, I'd like to see the config conditionally defined, and the old option should use the new one and issue a deprecation via our mechanism (this is configurable as to errors and silencing) this would also then simplify the usage.

lib/rspec/rails/fixture_support.rb
```diff
@@ -22,11 +22,14 @@ def run_in_transaction?
...
-              fixture_paths << RSpec.configuration.fixture_path
-            else
-              self.fixture_path = RSpec.configuration.fixture_path
+            if respond_to?(:fixture_paths=) && RSpec.configuration.fixture_paths.any?
+              self.fixture_paths = RSpec.configuration.fixture_paths
+            elsif respond_to?(:fixture_paths=) && Rspec.configuration.fixture_path
+              self.fixture_paths = Array(RSpec.configuration.fixture_path)
+            elsif !respond_to?(:fixture_paths=) && respond_to?(:fixture_path=)
+              self.fixture_path = RSpec.configuration.fixture_path || RSpec.configuration.fixture_paths&.first
             end
```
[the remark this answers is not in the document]
jguecaimburu: Fixed

spec/generators/rspec/install/install_generator_spec.rb
```diff
@@ -84,7 +88,11 @@ def filter_rails_from_backtrace
 
     specify "with default fixture path" do
       run_generator
-      expect(rails_helper).to have_a_fixture_path
+      if ::Rails::VERSION::STRING < '7.1.0'
+        expect(rails_helper).to have_a_fixture_path
+      else
+        expect(rails_helper).to have_a_fixture_paths
```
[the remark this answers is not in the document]
jguecaimburu: :see_no_evil:

lib/rspec/rails/configuration.rb
jguecaimburu: Fixed

jguecaimburu:
Thanks for the suggestions @JonRowe! I wasn't aware of the deprecation mechanism.
Fixed all the issues and added tests for the new behaviour.

Some tests are failing in `main` as I never touched this file: https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/fixture_file_upload_support.rb
I'm working on it.

jguecaimburu:
Fixed the failing spec and added new checks here: https://github.com/rspec/rspec-rails/pull/2673/commits/b64a9a48a020bcc5f9e66ae77d46abbabb0d3e22#
Not a 100% sure which is the purpose of `RailsFixtureFileWrapper` but I wanted to keep existing behavior here.

I tried removing the `:fixture_path` accessor from it and not a single test failed. Except for some cases of the concern test file, the complete suite passed if I changed the concern to this:
```rb
module RSpec
  module Rails
    # @private
    module FixtureFileUploadSupport
      include ActionDispatch::TestProcess if defined?(ActionDispatch::TestProcess)
    end
  end
end
```

My understanding is that `FileFixtureSupport` sets `file_fixture_path` to the one defined in the RSpec config setting, that has a default of `'spec/fixtures/files'`. The wrapper handles the case where that value could be set manually to `nil` or am I missing something else?

For reference, Rails sets the default file fixture path here: https://github.com/rails/rails/blob/6f6a04f8781d03261b5cbc1d8b65ebb43311f2a4/railties/lib/rails/test_help.rb#L28

lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -34,7 +34,13 @@
 RSpec.configure do |config|
 <% if RSpec::Rails::FeatureCheck.has_active_record? -%>
   # Remove this line if you're not using ActiveRecord or ActiveRecord fixtures
+<% if ::Rails::VERSION::STRING < "7.1.0" -%>
   config.fixture_path = Rails.root.join('spec/fixtures')
```
pirj, line 38:
```suggestion
  config.fixture_path = Rails.root.join('spec/fixtures')
  # Starting with Rails 7.1, use this instead:
  # config.fixture_paths = [
  #  Rails.root.join('spec/fixtures')
  # ]
```

lib/rspec/rails/fixture_file_upload_support.rb
```diff
@@ -14,6 +14,8 @@ def rails_fixture_file_wrapper
         resolved_fixture_path =
           if respond_to?(:file_fixture_path) && !file_fixture_path.nil?
             file_fixture_path.to_s
+          elsif respond_to?(:fixture_paths)
+            (RSpec.configuration.fixture_paths&.first || '').to_s
```
pirj, line 18: Is it possible that `respond_to?(:file_fixture_path)` is `false` in Rails 7.1+?

jguecaimburu: I don't think so. This module is included in `TestCase`: https://github.com/rails/rails/blob/b72b7c35dd9506b08ed040594bbd4536139e6d3f/activesupport/lib/active_support/testing/file_fixtures.rb#L16

But it could be set to `nil` explicitly if somebody decides to do `Rspec.configure { |config| config.file_fixture_path = nil }` or `self.file_fixture_path = nil` in a test case. But in that case, I'm not sure why the wrapper should return a default value. Maybe I'm missing something here.

lib/generators/rspec/install/templates/spec/rails_helper.rb
jguecaimburu: Would you default to a deprecated method for a brand new installation? Is it to keep the template simple?
We'll need to change the file again once Rails 7.2 is released in that case.

pirj: My suggestion is to add a note and a clear suggestion on how to update for those starting with Rails 7.0 which is the current one.
When Rails 7.1 becomes current, the one from the below `else` clause with `fixture_paths = ` will be added.

jguecaimburu: Oh, so sorry! I misread your suggestion. Makes sense

lib/rspec/rails/fixture_file_upload_support.rb
JonRowe: Does Rails not support multiple paths here? Seems odd they would allow multiple paths for fixtures but not for files?

jguecaimburu: It is weird indeed, but I could not find any reference to a plural method in their codebase. [Here](https://github.com/rails/rails/pull/47675#discussion_r1136329301) the author of the change mentions that it is something that they might wanna do in the future.

lib/rspec/rails/configuration.rb
```diff
@@ -157,6 +163,18 @@ def filter_rails_from_backtrace!
           filter_gems_from_backtrace "activemodel", "activerecord",
                                      "activesupport", "activejob"
         end
+
+        # @deprecated TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
+        if ::Rails::VERSION::STRING >= "7.1.0"
+          def config.fixture_path=(path)
```
pirj, line 169:
```suggestion
          def fixture_path=(path)
```
We're inside `config.instance_exec`, should this be sufficient?

spec/rspec/rails/fixture_support_spec.rb
```diff
@@ -1,14 +1,19 @@
 module RSpec::Rails
   RSpec.describe FixtureSupport do
     context "with use_transactional_fixtures set to false" do
-      it "still supports fixture_path" do
+      it "still supports fixture_path and fixture_paths" do
```
pirj, line 4:
```suggestion
      it "still supports fixture_path/fixture_paths" do
```

pirj (approved):
Looks great, thanks!

lib/rspec/rails/configuration.rb
jguecaimburu: You're right

jguecaimburu:
@pirj @JonRowe made the last couple of fixes. I believe this is ready to merge.

Would you consider a PR to simplify `FileFixtureUploadSupport` as mentioned here https://github.com/rspec/rspec-rails/pull/2673#issuecomment-1517822088?

lib/rspec/rails/fixture_file_upload_support.rb
JonRowe: Does Rails solely configure this with file_fixture_path directly?

pirj:
> simplify FileFixtureUploadSupport

It looks sleek. WDYT of doing this in a separate PR?

lib/rspec/rails/fixture_file_upload_support.rb
jguecaimburu: I think so: https://github.com/search?q=repo%3Arails%2Frails%20file_fixture_path&type=code

It defaults to `"#{Rails.root}/test/fixtures/files"` most of the time and you can overwrite that config. Rails does not use any of the configured `fixture_paths` to build the `file_fixture_path`. Althought the first of the fixture paths is `"#{Rails.root}/test/fixtures"` by default, I think that relation is not enforced.

jguecaimburu:
> > simplify FileFixtureUploadSupport
> 
> It looks sleek. WDYT of doing this in a separate PR?

I believe that change is not completely related to the Rails deprecation addressed here so better to keep it separately and merge this one as it is. But I can add an extra commit here as well. I'll follow your advice on this.

jguecaimburu:
@JonRowe @pirj please let me know if there's anything missing here

benoittgt (approved):

JonRowe:
Thanks for the work on this and apologies for the delay, I've merged this manually 🎉

JonRowe:
Released in 6.1.0

lib/rspec/rails/configuration.rb
```diff
@@ -157,6 +163,18 @@ def filter_rails_from_backtrace!
...
                                      "activesupport", "activejob"
         end
+
+        # @deprecated TestFixtures#fixture_path is deprecated and will be removed in Rails 7.2
+        if ::Rails::VERSION::STRING >= "7.1.0"
+          def fixture_path=(path)
+            RSpec.deprecate(
+              "config.fixture_path =  #{path.inspect}",
+              replacement: "config.fixture_paths = [#{path.inspect}]",
+              message: "Rails 7.1 has deprecated the singular fixture_path in favour of an array"
```
benoittgt, line 173: And we should add.
```
Rails 7.1 has deprecated the singular fixture_path. You should migrate to plural fixture_paths and wrap value into an array.

config.fixture_paths = ["#{Rails.root}/spec/fixtures/"]

```

JonRowe: `config.fixture_paths = ["#{Rails.root}/spec/fixtures/"]` is implied by the replacement text, would support a message tweak to [or similar]:

`Rails 7.1 has deprecated the singular fixture_path. You should migrate to plural:`

JonRowe: The singular could also error if given an array

benoittgt: Do you want a PR @JonRowe ?

JonRowe: Always :p

benoittgt: https://github.com/rspec/rspec-rails/pull/2720

PR #2680, opened 2023-05-30
Allow to use symbols for broadcasting target

Vagab:
# Problem

Test below will not pass:
```ruby
# some_service.rb

class SomeService
  def call(record)
    record.broadcast_append_to(:some_records)
  end
end

# some_service_spec.rb

it "broadcasts to :some_records" do
  expect { SomeService.call(record) }.to broadcast_to(:some_records)
end
```

That is because internally rails does `String(broadcasting)`, so test below would actually pass:
```ruby
it "broadcasts to :some_records" do
  expect { SomeService.call(record) }.to broadcast_to("some_records")
end
```

Which is a bit counterintuitive.

# Proposed solution

Just `to_s` if the channel passed is a symbol. Can possibly break if someone internally checks for symbolized channel, but the matcher should work.

### DIFF
--- a/lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
+++ b/lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
@@ -96,8 +96,11 @@ module RSpec
         private
 
           def stream
-            @stream ||= if @target.is_a?(String)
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks!

PR #2686, opened 2023-06-26
Allow view specs to use implicit `render locals: {...}` calls.

JonRowe:
If only locals are provided when rendering a view spec, we can use _default_render_options to provide the view to render, this allows bare `render locals: {}` style calls.


benoittgt (approved):
LGTM

JonRowe:
Released in 6.1.0

PR #2687, opened 2023-07-05
Allow to use have_http_status on HTTP.rb responses too

stoivo:
I use http.rb and wanted to test some status codes. I have used have_status_code for a while and really enjoined it. 

I haved used this patch for a litle while and think I could ask if you want it upstream
```rb
module RSpec::Rails::Matchers::HaveHttpStatus
  # Example
  # Failure/Error: expect(HTTP.get("https://www.nrk.no")).to have_http_status(:created)
  #   expected the response to have status code :created (201) but it was :ok (200)

  alias as_test_response_before_httprb_support as_test_response
  def as_test_response(obj)
    if defined?(::HTTP::Response) && ::HTTP::Response === obj
      obj = ActionDispatch::Response.new.tap do |resp|
        resp.status  = obj.status.code
        resp.headers.clear
        resp.headers.merge!(obj.headers.to_h)
        resp.body    = obj.body.to_s
        resp.request = ActionDispatch::Request.new({})
      end
      ::ActionDispatch::TestResponse.from_response(obj)
    else
      as_test_response_before_httprb_support(obj)
    end
  end
end
```
If you don't I am 100% in on that.

### DIFF
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -33,7 +33,16 @@ module RSpec
         # @param obj [Object] object to convert to a response
         # @return [ActionDispatch::TestResponse]
         def as_test_response(obj)
-          if ::ActionDispatch::Response === obj
+          if defined?(::HTTP::Response) && ::HTTP::Response === obj
+            obj = ActionDispatch::Response.new.tap do |resp|
+              resp.status  = obj.status.code
+              resp.headers.clear
+              resp.headers.merge!(obj.headers.to_h)
+              resp.body    = obj.body.to_s
+              resp.request = ActionDispatch::Request.new({})
+            end
+            ::ActionDispatch::TestResponse.from_response(obj)
+          elsif ::ActionDispatch::Response === obj
             ::ActionDispatch::TestResponse.from_response(obj)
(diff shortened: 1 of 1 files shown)


JonRowe:
I'm on the fence about this, `rspec-rails` is for compatibility with Rails test helpers, and they wouldn't match these requests would they? If you want to implement matchers for `http.rb` you could always create `rspec-http.rb`

stoivo:
Yeah, I could make a small gem with just this but I don't know if I think cool to have a soo small gem. An alternative is to check if it would be appreciated in http.rb. I did a bit of testing and it seams like if you define two matchers with the same name the last one will overwrite the first one. Do you have some suggestion how to work around that?  If I want to include in into a separate gem or in http.rb I will not require rspec-rails. How do I overcome the name? the easiest is just to create a separate matcher like "httprb_have_status"

JonRowe:
Well you could require `rspec/rails` before you define your matcher, then call super as a fallback to allow it to "override" our matcher but still support it, or give it a seperate name, I don't think theres a problem with small gems for focused purposes, or ask http.rb to include the matcher themselves?

Anyway closing this for `rspec/rails`

stoivo:
okay, thanks

PR #2692, opened 2023-10-06
Add Rails 7.1 builds

JonRowe:

PR #2698, opened 2023-10-13
Stop using webdrivers

jcoyne:
selenium-webdriver provides all that is needed now

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -31,7 +31,8 @@ in_root do
   # sqlite3 is an optional, unspecified, dependency and Rails 6.0 only supports `~> 1.4`
   gsub_file "Gemfile", /.*gem..sqlite3.*/, "gem 'sqlite3', '~> 1.4'"
 
-  gsub_file "Gemfile", /.*chromedriver-helper.*/, "gem 'webdrivers'"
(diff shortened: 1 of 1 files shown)


pirj:
Isn’t selenium-webdriver 4.10+ that is capable to use the selenium manager Ruby 3+ only?

We’re fixing build issues at the moment, and this is the major one. If you have a clear idea how to download drivers for Ruby 2.6, 2.7 and 3+, that will help a lot.

Please ignore Rails 7.1 and 7.0 builds, there’s at least some mailers failing spec.

pirj:
3.1 and 3.0 fail because of this https://github.com/rails/rails/blob/1540cb381b4c78f1bcd7e081717a9e0a6706a9d8/railties/lib/rails/generators/rails/app/templates/Gemfile.tt#L73

pirj:
https://github.com/rspec/rspec-rails/pull/2692

JonRowe:
Bare in mind this branch supports Rails 6 as well as 7, and this file is about making a new rails app work on CI, so this probably needs at least a version switch around the rails version, as we are modifying a default app's Gemfile...

JonRowe:
I didn't think this would work on 6.1 but it does, thus I'm going to merge it into #2692 thank you!

Issue #2690, opened 2023-09-14
Rails 7.1.beta.1: second test failure on system specs 


zealot128:
Ruby version: 3.2.2
Rails version: 7.1.0.beta.1
RSpec version: 3.12.2

## Observed behaviour

When a failure on a system spec in a Rails app occurs (e.g. 500 error in a callback or navigation), than a second "1.2 )" failure is generated that looks like this:

```
1.2) Failure/Error: 

 `metadata` is not available from within an example (e.g. an `it` block) or from constructs that run in the scope of an example (e.g. `before` `let`, etc). It is only available on an example group (e.g. a `describe` or `context` block).
[Screenshot Image]: xxxx
```

```
gems/rspec-core-3.12.2/lib/rspec/core/example_group.rb:760:in `method_missing'
gems/activerecord-7.1.0.beta1/lib/active_record/test_fixtures.rb:278:in `method_missing'
gems/actionpack-7.1.0.beta1/lib/action_dispatch/testing/assertions/routing.rb:196:in `method_missing'
gems/actionpack-7.1.0.beta1/lib/action_dispatch/testing/integration.rb:440:in `method_missing'
gems/actionpack-7.1.0.beta1/lib/action_dispatch/system_testing/test_helpers/screenshot_helper.rb:48:in `take_failed_screenshot'
gems/actionpack-7.1.0.beta1/lib/action_dispatch/system_testing/test_helpers/setup_and_teardown.rb:8:in `before_teardown'
gems/rspec-rails-6.0.3/lib/rspec/rails/example/system_example_group.rb:110:in `call'
gems/rspec-rails-6.0.3/lib/rspec/rails/example/system_example_group.rb:110:in `block (2 levels) in <module:SystemExampleGroup>'
gems/rspec-core-3.12.2/lib/rspec/core/example.rb:457:in `instance_exec'
gems/rspec-core-3.12.2/lib/rspec/core/example.rb:457:in `instance_exec'
gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:372:in `run'
gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:529:in `block in run_owned_hooks_for'
gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:528:in `each'
gems/rspec-core-3.12.2/lib/rspec/core/hooks.rb:528:in `run_owned_hooks_for'
[...]

PR #2704, opened 2023-10-21
Fix failure screenshots on Rails 7.1

JonRowe:
On Rails 7.1 they access metadata directly to store the failure screenshot path, we don't allow metadata access like this and it throws an error, this patch makes a dummy metadata available for this purpose, whilst still maintaining our friendly error.


gap777:
@JonRowe Our upgrade to Rails 7.1 is blocked pending this or another fix to https://github.com/rspec/rspec-rails/issues/2690 ... when can this move forward?

pirj:
@gap777 The best you can do is to attempt to fix this in a separate PR.

JonRowe:
Technically released in 6.0.4 but please use 6.1.0 for Rails 7.1 support.

davidalejandroaguilar:
This was VERY annoying, thank you for your work!

PR #2705, opened 2023-10-21
Fix deprecated use of `config.action_mailer.preview_path =` in favor of `...preview_paths <<` to support Rails 7.1 upgrade

### LINKED ISSUE #2703: Rails 7.1 upgrade: using deprecated preview_path=

Ruby version: 3.2.2
Rails version: 7.1.1
RSpec version: 3.12

## Observed behaviour
Inside of `rspec-rails-6.0.3/lib/rspec-rails.rb` it's using the deprecated  method `preview_path`:

```
 options.preview_path = "#{::Rails.root}/spec/mailers/previews"
```
which ultimately gets set as a config for `action_mailer`:
```
      initializer "rspec_rails.action_mailer",
                  before: "action_mailer.set_configs" do |app|
                    setup_preview_path(app)
                  end
```

and causes the following message to trigger:

```
        ActionMailer.deprecator.warn(<<-MSG.squish)
          Using preview_path= option is deprecated and will be removed in Rails 7.2.
          Please use preview_paths= instead.
        MSG
```

## Expected behaviour
No deprecation warning

Thanks!

alexevanczuk:
Fixes #2703

Also see https://github.com/rails/rails/pull/31595, particularly [this comment](https://github.com/rails/rails/pull/31595#issuecomment-1769265335)

### DIFF
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -42,7 +42,7 @@ require_file_stub 'config/environment' do
         config.action_mailer.raise_delivery_errors = false unless ENV['NO_ACTION_MAILER']
 
         if ENV['CUSTOM_PREVIEW_PATH']
-          config.action_mailer.preview_path = ENV['CUSTOM_PREVIEW_PATH']
+          config.action_mailer.preview_paths << ENV['CUSTOM_PREVIEW_PATH']
         end
         if ENV['SHOW_PREVIEWS']
           config.action_mailer.show_previews = (ENV['SHOW_PREVIEWS'] == 'true')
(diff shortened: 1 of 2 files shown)


JonRowe:
Thanks, I'm actually working on this myself at the moment as its not quite this simple, we need to support both methods and also fix that check

alexevanczuk:
Gotcha @JonRowe. Do you know where I can find what versions of Rails that rspec-rails currently supports? I wonder if we could just check rails version and use the new api if the version is greater than or equal to the version of rails that introduced the api.  What do ya think?

pirj:
> use [rspec-rails 6.x](https://github.com/rspec/rspec-rails/tree/6-0-maintenance) for Rails 6.1 or 7.x.

> check rails version
Or like this https://github.com/rspec/rspec-rails/blob/96812669bdfe54d83fbd2043e033b8d7db4147e4/lib/rspec/rails/fixture_support.rb#L25

JonRowe:
We support the same matrix as Rails, so 6.1, 7.0 and 7.1

JonRowe:
Closing in favour of #2706

PR #2706, opened 2023-10-23
Fix preview paths

### LINKED ISSUE #2703: Rails 7.1 upgrade: using deprecated preview_path=

JonRowe:
On Rails 7.1 they changed the preview path setup for mailers to be multiple rather than singular, this PR handles that change conditionally in the railtie and fixes our test setup around the mailers. Fixes #2703.

### DIFF
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -65,7 +65,7 @@ require_file_stub 'config/environment' do
   Rails.application.initialize!
 end
 
-exit if ENV['NO_ACTION_MAILER']
+exit(0) if ENV['NO_ACTION_MAILER']
 if ENV['DEFAULT_URL']
   puts ActionMailer::Base.default_url_options[:host]
 elsif defined?(::ActionMailer::Preview)
@@ -79,3 +79,4 @@ end
 # This will force the loading of ActionMailer settings to ensure we do not
 # accidentally set something we should not
 ActionMailer::Base.smtp_settings
+exit 0
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -37,6 +37,9 @@ require_file_stub 'config/environment' do
     module ExampleApp
       class Application < Rails::Application
         config.eager_load = false
+        if Rails::VERSION::STRING.start_with?('7')
+          config.active_support.cache_format_version = 7.0
+        end
 
         # Don't care if the mailer can't send.
         config.action_mailer.raise_delivery_errors = false unless ENV['NO_ACTION_MAILER']
--- a/example_app_generator/spec/support/default_preview_path
+++ b/example_app_generator/spec/support/default_preview_path
@@ -37,14 +37,14 @@ require_file_stub 'config/environment' do
     module ExampleApp
       class Application < Rails::Application
         config.eager_load = false
-        if Rails::VERSION::STRING.start_with?('7')
+        if Rails::VERSION::STRING.to_f >= 7.0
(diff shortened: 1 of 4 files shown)


example_app_generator/spec/support/default_preview_path
```diff
@@ -37,12 +37,18 @@ require_file_stub 'config/environment' do
...
+        if Rails::VERSION::STRING.start_with?('7')
+          config.active_support.cache_format_version = 7.0
+        end
 
         # Don't care if the mailer can't send.
         config.action_mailer.raise_delivery_errors = false unless ENV['NO_ACTION_MAILER']
-
         if ENV['CUSTOM_PREVIEW_PATH']
-          config.action_mailer.preview_path = ENV['CUSTOM_PREVIEW_PATH']
+          if Rails::VERSION::STRING.start_with?('7.1')
```
ryanb, line 47: @JonRowe Won't this fail with Rails 7.2+? Would it be better to do a `>=` comparison?

alexevanczuk:
Hey @JonRowe let me know if this is something that you're still working on or if I can help out!

TylerRick:
Can you please make a release that includes this? :pray: 

The current [release](https://rubygems.org/gems/rspec-rails) (6.0.3) does not include this, and is therefore not compatible with Rails 7.1 (when using `config.active_support.disallowed_deprecation = :raise`, as some of us do). :disappointed:

TylerRick:
A related question: While waiting for a release, is there an easy way to use the edge/unreleased version?

I tried `gem 'rspec-rails', github: 'rspec/rspec-rails'` but run into this error (unless I also explicitly list every rspec sub-project in my Gemfile):

```
Could not find gem 'rspec-mocks (= 3.13.0.pre)', which is required by gem 'rspec-rails', in rubygems repository https://rubygems.org/ or
installed locally.
```

dannyy83:
@TylerRick I'm in the same situation as you, and this works for me:

```
    gem lib, git: "https://github.com/rspec/#{lib}.git", branch: 'main'
  end
```

TylerRick:
> @TylerRick I'm in the same situation as you, and this works for me:
> 
> ```
>   %w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
>     gem lib, git: "https://github.com/rspec/#{lib}.git", branch: 'main'
>   end
> ```

Thanks, @dannyy83, and glad to hear I'm not alone. I landed on something similar. Too bad you have to override all of those dependencies, but I guess it makes sense, due to the version constraint on "= 3.13.0.pre)".

JonRowe:
Fix released in 6.1.0 for Rails 7.1 support.

Side note, if you depend on the `<major>-<minor>-maintenance` branch you shouldn't need to pin the rest of RSpec, but we only tend to ship bugfixes to those.

PR #2712, opened 2023-11-16
Wrap tests in executor when configured or on Rails 7 by default

JonRowe:


lib/rspec/rails/example/rails_example_group.rb
```diff
@@ -13,6 +13,23 @@ module RailsExampleGroup
...
       include RSpec::Rails::TaggedLoggingAdapter if ::Rails::VERSION::MAJOR >= 7
+
+      included do |_other|
+        around do |example|
+          use_exectuor =
+            case ::Rails.configuration.active_support.executor_around_test_case
+            when nil then ::Rails::VERSION::MAJOR >= 7
+            when true then true
+            when false then false
+            end
```
semaperepelitsa, line 24: There is no need to do this. "Default in Rails 7" means it is set as true when calling `config.load_defaults 7.0`. Otherwise, it should stay disabled.

pirj: Do you mean it’s unnecessary to default to true on Rails 7, and it’s sufficient to just?
 ```ruby
use_exectuor = ::Rails.configuration.active_support.executor_around_test_case
```

semaperepelitsa: Yes, exactly same as it is used in Rails

https://github.com/rails/rails/blob/8278626a2c0e0ee4dc762891a91e70708380b4d6/activesupport/lib/active_support/railtie.rb#L52-L63

PR #2718, opened 2023-12-03
Migrate to Rails 7.1's `file_fixture_upload` naming approach

agrobbin:
rails/rails#48857 renamed `fixture_file_upload` to `file_fixture_upload` to reduce "confusion and surprise". With this change, `file_fixture_upload` now works in RSpec, with `fixture_file_upload` still functional to maintain backwards compatibility.


agrobbin:
I'm surprised by the test failures here, and am curious if they are due to something recent in another rspec gem (tests haven't been run on `main` in ~2 weeks). If there is something I've screwed up that's causing these failures, please let me know and I'll look into it!

JonRowe:
This is actually due to the ameter gem adding a deprecation warning to a matcher we apparently used in our generator specs it looks like its an RSpec problem but its just we check our builds for warnings and so the error is hidden slightly, working on a fix in #2719 and then this can be rebased :)

agrobbin:
@JonRowe thanks for the quick fix there! I've rebased this and gotten tests passing. It took an extra pass to add the appropriate testing for Rails 7.1+ and < 7.0, but I think things are in good shape now. Let me know if there's anything else I should change.

lib/rspec/rails/file_fixture_upload_support.rb
```diff
@@ -1,16 +1,16 @@
 module RSpec
   module Rails
     # @private
-    module FixtureFileUploadSupport
-      delegate :fixture_file_upload, to: :rails_fixture_file_wrapper
+    module FileFixtureUploadSupport
+      delegate :file_fixture_upload, :fixture_file_upload, to: :rails_file_fixture_wrapper
```
agrobbin, line 5: One minor note... I have included the delegation for `file_fixture_upload` in *all* Rails versions, even though it's only available in Rails 7.1.+. It should raise a `NoMethodError` if called in Rails < 7.1.

JonRowe: I'm confused as to what provides the old behaviour with this change, I can see the build passing but I don't understand what now provides `fixture_file_upload`

agrobbin: @JonRowe it's hard to see, since `file_fixture_upload` and `fixture_file_upload` look so similar, but `fixture_file_upload` is on this same `delegate` line, after `file_fixture_upload`.

JonRowe: Ah I see, sorry!

agrobbin: No problem!

JonRowe:
Overall this looks fine, but I'm nervous of the support differences for 6.1 and 7.x, especially the change in module. Could I be a pain and ask you split this into two, with the old module for the old behaviour and the new module, and conditionally include the right module? If you don't want to I'm happy to finish this off

agrobbin:
Happy to, @JonRowe! I should be able to do that tonight.

agrobbin:
@JonRowe when I started looking at this, I actually noticed that `FixtureFileUploadSupport` already includes some conditional logic for Rails 7.1+ compared other versions:

https://github.com/rspec/rspec-rails/blob/cbc6fccc17837247ee0123e10cfbdf7ea06d7971/lib/rspec/rails/fixture_file_upload_support.rb#L31-L35

With that in mind, I made a change that is more explicit about which methods are supported, but I left it in the same module. I'm curious if you think that's reasonable, or if you'd still prefer to break things apart.

mjankowski:
Curious about state of this one ...

I was attempting to rename some of the (soft deprecated) `fixture_file_upload` calls to use `file_fixture_upload` instead. It went fine in places where the call was directly in the params of a request being made in an example, but I saw `NoMethodError` when calling it from a `let`, and then found this issue.

Given the drop of rails 7.1 support in latest rspec-rails versions - would it make sense to rebase/rework this and drop all the version conditional stuff, since (I think?) both 7.2 and 8.0 will be ok with this?

Aside from that, any blockers here?

PR #2719, opened 2023-12-04
Remove usage of ammeter exist matcher

JonRowe:
This is breaking downstream builds due to their deprecation warning triggering out "no warnings" code, took this as an opporunity to rewrite some of these.


pirj (approved):
`contain` implies it exists, and with `and` specs look even better to my personal taste.
Thank you!

PR #2721, opened 2023-12-10
Add spec for namespaced fixtures

JonRowe:


janko:
Any chance we could get a new release with this fix? 🙏🏻 We could really use the Rails 7.1 compatibility fixes, just this was preventing us from upgrading. I tried pulling from GitHub, but realized I would have to pull all `rspec-*` gems from GitHub as well.

JonRowe:
This was released in 6.1.1 fwiw

PR #2728, opened 2024-02-02
Fix tests to expect a non-match, not an error

naveg:
In https://github.com/rspec/rspec-expectations/pull/1449, the `Include`
matcher is fixed so that it doesn't raise an error when types don't
match. Instead, the actual and expected will be deemed to not match.

### DIFF
--- a/spec/rspec/rails/matchers/be_a_new_spec.rb
+++ b/spec/rspec/rails/matchers/be_a_new_spec.rb
@@ -79,12 +79,7 @@ RSpec.describe "be_a_new matcher" do
 
           context "matcher is wrong type" do
             it "fails" do
-              expect {
-                expect(record).to be_a_new(record.class).with(
-                  foo: a_hash_including({ no_foo: "foo" }))
-              }.to raise_error { |e|
-                expect(e.message).to eq("no implicit conversion of Hash into String").or eq("can't convert Hash into String")
-              }
+              expect(record).not_to be_a_new(record.class).with(foo: a_hash_including({ no_foo: "foo" }))
             end
           end
         end


spec/rspec/rails/matchers/be_a_new_spec.rb
```diff
@@ -79,12 +79,7 @@ def new_record?; true; end
 
           context "matcher is wrong type" do
             it "fails" do
-              expect {
-                expect(record).to be_a_new(record.class).with(
-                  foo: a_hash_including({ no_foo: "foo" }))
-              }.to raise_error { |e|
-                expect(e.message).to eq("no implicit conversion of Hash into String").or eq("can't convert Hash into String")
-              }
+              expect(record).not_to be_a_new(record.class).with(foo: a_hash_including({ no_foo: "foo" }))
```
pirj, line 82: This makes me wonder what the failure message is if we `expect(record).to`

spec/rspec/rails/matchers/be_a_new_spec.rb
```diff
@@ -79,12 +79,7 @@ def new_record?; true; end
 
           context "matcher is wrong type" do
             it "fails" do
-              expect {
-                expect(record).to be_a_new(record.class).with(
-                  foo: a_hash_including({ no_foo: "foo" }))
```
pirj, line 84: What if we just use a different "wrong type" of matcher here and keep the example as is? Like `be_within`/`cover`/`contain_exactly`.
At the same time, this example won't break after your rspec-expectations fix.

pirj:
Closing as the root PR was closed

PR #2730, opened 2024-02-07
Remove type metadata from templates

r7kamura:
rspec-rails will generate spec/rails_helper.rb with `config.infer_spec_type_from_file_location!` enabled. So by default it assumes a project that does not explicitly specify type metadata.

https://github.com/rspec/rspec-rails/blob/06d05aba209565f5bc940aa9e67232c8616e5197/lib/generators/rspec/install/templates/spec/ rails_helper.rb#L78

Therefore, shouldn't the generators also generate a test file skeleton that does not explicitly specify type metadata?

### DIFF
--- a/lib/generators/rspec/channel/templates/channel_spec.rb.erb
+++ b/lib/generators/rspec/channel/templates/channel_spec.rb.erb
@@ -1,7 +1,7 @@
 require 'rails_helper'
 
 <% module_namespacing do -%>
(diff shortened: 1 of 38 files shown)


pirj:
https://github.com/rspec/rspec-rails/issues/2507#issuecomment-1107794265 we are more incluned to keep explicit types, and remove inferred.

PR #2731, opened 2024-02-09
Setup rspec-rails to allow main to be used without requiring git rspec

JonRowe:
Increase the desired rspec version to recently released 3.13, but also drop the requirement for pre release versions of RSpec when using rspec-rails from this repo, we seperated rspec-rails from rspec's release cadence so theres no real reason for this to be here, I've kept the CI addition because it allows the build to test against the head but this should make it easier for other people to use pre release versions of rspec-rails


pirj:
Fantastic!

PR #2733, opened 2024-02-19
Fix failing fixture support test on Rails main

eugeneius:
This test is failing since `setup_fixtures` was made private in https://github.com/rails/rails/pull/51021.

It was added in https://github.com/rspec/rspec-rails/pull/2215 to exercise the call to `name` in Rails' `run_in_transaction?` method: https://github.com/rails/rails/blob/b4ab1f19d8bcc4639c7379b560f000f01753b5b0/activerecord/lib/active_record/test_fixtures.rb#L101

However since https://github.com/rspec/rspec-rails/pull/2461, RSpec Rails overrides `run_in_transaction?`, so `name` is no longer called. We can get even better integration coverage of an example group with `use_transactional_tests` set to false by actually running an example.

### DIFF
--- a/spec/rspec/rails/fixture_support_spec.rb
+++ b/spec/rspec/rails/fixture_support_spec.rb
@@ -39,14 +39,21 @@ module RSpec::Rails
       end
     end
 
-    it "will allow #setup_fixture to run successfully" do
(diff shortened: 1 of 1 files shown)


JonRowe:
Thanks!

PR #2735, opened 2024-02-22
Fix generator for mailer

psantos10:
In Rails, I can generate a mailer in both ways:

```sh
bin/rails generate mailer User
```

or 

```sh
bin/rails generate mailer UserMailer
```

Both ways will create a class `UserMailer` under the file `app/mailers/user_mailer.rb` and the test file `test/mailers/user_mailer_test.rb` (Minitest).

When using RSpec, if I call the command like: `bin/rails generate mailer User`, it will create the test file name like: `spec/mailers/user_spec.rb` instead of `spec/mailers/user_mailer_spec.rb`. The class though, is correct for the mailer test , but no for the preview file.

This PR try to solve this


pirj:
The build will fail. Once the fix is merged, we’ll kick off ci

JonRowe:
It will still need to be rebased

psantos10:
Hi @pirj  

Sorry, not sure if I got you correctly. "Once the fix s. merged" - This fix needs to be done by me? What is failing? 

Sorry, this is my first time here :)

pirj:
Please rebase

psantos10:
> Please rebase

Done. Thanks

JonRowe:
Merged, thans!

PR #2740, opened 2024-02-26
Brakeman2

sumonsm:


JonRowe:
Please stop opening random PRs that look like you've either misconfigured a codespace or are using some other automated tool. Also note we have no desire to add brakeman or any further coverage to our build without prior discussion, we already have coverage checks as part of our build.

Note I've temporarily blocked you from being able to interact with the repo but it will expire in a few days.

PR #2741, opened 2024-03-05
Support chages to `show_helper` in Rails 7

rickselby:
#2492 overrode `show_helper`, which in Rails 6 looks like:

```ruby
def show_helper
  "#{singular_route_name}_url(@#{singular_table_name})"
end
```

But Rails 7 changed it to

```ruby
def show_helper(arg = "@#{singular_table_name}", type: :url) # :doc:
  "#{singular_route_name}_#{type}(#{arg})"
end
```

This change supports the arguments in Rails 7 while maintaining support for Rails 6.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -13,7 +13,7 @@ Enhancements:
 Bug Fixes:
 
 * Check wether rspec-mocks has been loaded before enabling signature
-  verification for `have_enqueued_job` et al (Jon Rowe, #2823)
+  verification for `have_enqueued_job` et al (Jon Rowe, rspec/rspec-rails#2823)
 
 ### 7.1.0 / 2024-11-09
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v7.0.2...v7.1.0)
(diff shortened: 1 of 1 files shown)


pirj:
Do you think it’s possible to add a failing example?

rickselby:
Not unless there's an easy way of providing a different template to a spec test.

Ideally I'd add a test template that called `show_helper(type: :path)` and could test the output of the generator with that template. But I can't see a way of doing that.

Also, if anyone has any preferences on how to address the rubocop failure, I'll make a change.

JonRowe (commented):
:wave: thanks for looking at this, can you provide a spec for this and see how we usually prefer to define these kind of conditionals.

If the refactoring doesn't take care of the rubocop warning then add a disable to the class.

rickselby:
I've updated the change to remove the override for Rails 6 (and have reverted `show_helper` to what it was before).

I'm still struggling to see a good way of setting up a spec test; but as I'm now not changing the function, just removing it from Rails 7, then the existing Rails 7 tests should show that it still works? The default templates do call `show_helper`.

rickselby:
Whoops! Restarted in #2833

PR #2742, opened 2024-03-08
Only load example group modules when they're used

eugeneius:
Currently the example group modules for all spec types are always loaded, regardless of which type of spec is being run. Some of those modules force parts of the user's application to load; for example, `RSpec::Rails::HelperExampleGroup` includes `ActionView::TestCase::Behavior`, and loading `ActionView::TestCase` loads all of the application's helpers.

By autoloading these modules and including them lazily the first time the corresponding spec type is defined, we can avoid loading parts of the user's application unnecessarily.

Co-authored by @ilianah 🍐

### DIFF
--- a/spec/rspec/rails/configuration_spec.rb
+++ b/spec/rspec/rails/configuration_spec.rb
@@ -291,6 +291,7 @@ RSpec.describe "Configuration" do
   end
 
   it "metadata `type: :request` sets up request example groups" do
+    require "rspec/rails/example/request_example_group"
     a_rails_app = double("Rails application")
     the_rails_module = Module.new {
       def self.version; end
@@ -328,6 +329,7 @@ RSpec.describe "Configuration" do
   end
 
   it "metadata `type: :feature` sets up feature example groups" do
+    require "rspec/rails/example/feature_example_group"
     a_rails_app = double("Rails application")
     the_rails_module = Module.new {
       def self.version; end


lib/rspec/rails/configuration.rb
```diff
@@ -45,15 +45,40 @@ class Configuration
...
+      config.define_derived_metadata(type: :model) do
+        config.include RSpec::Rails::ModelExampleGroup, type: :model
+      end
+
+      config.define_derived_metadata(type: :request) do
+        config.include RSpec::Rails::RequestExampleGroup, type: :request
+      end
+
+      config.define_derived_metadata(type: :routing) do
+        require "action_controller/test_case"
```
eugeneius, line 65: This is necessary because `RSpec::Rails::RoutingExampleGroup` depends on `ActionDispatch::Assertions::RoutingAssertions`, which references `ActionController::TestRequest` without requiring the file where it's defined:
https://github.com/rails/rails/blob/v7.1.3.2/actionpack/lib/action_dispatch/testing/assertions/routing.rb#L242

I've fixed that upstream in https://github.com/rails/rails/commit/53079761a55a7460c75db178bd48bf6b26588408, but for compatibility with all existing Rails versions we need to keep eager loading `ActionController::TestCase` here.

JonRowe:
Anything we can do to load things only when needed gets a plus from me, however I'm slightly concerned the build is failing because our snippets are exposing the change in whats loaded by default (in the case of 3.2 its just minitest being missing) which makes me wonder if this is a breaking change from a users perspective but I'm willing to be convinced its not if its easy to fix the build.

The suggested improvement to rspec-core is also fine, I think the helpers you are using were actually invented for rspec-rails ironically (for the define derived types options)

eugeneius:
Previously `ActiveSupport::TestCase` was always loaded, so minitest would be required here: https://github.com/rails/rails/blob/v7.1.3.2/activesupport/lib/active_support/test_case.rb#L3

The most straightforward fix for that is to continue to load minitest by requiring it explicitly; I've pushed a commit to do that.

spec/rspec/rails/configuration_spec.rb
```diff
@@ -291,6 +291,7 @@ def in_inferring_type_from_location_environment
   end
 
   it "metadata `type: :request` sets up request example groups" do
+    require "rspec/rails/example/request_example_group"
```
JonRowe, line 294: I was going to comment "This could cause order dependent failures, we typically do such requires in a sub process to avoid posioning the env" but its made me think prehaps we should be doing that all over now if we lazy load so that we don't accidentally depend on a module load order that we just happen to have in our specs 😧

PR #2744, opened 2024-03-14
Clean up generator for generator specs

budu:
Don't force pluralization of the generator name in the generated spec and use the full generator name in the spec description.

Also clean up an extra whitespace and empty line in the template.

Note that the generator class constant cannot be directly used in the spec description as it isn't loaded by default.


pirj:
I’m happy to merge it on green CI. I guess it’s a matter of removing two chars?

budu:
Done, totally missed the cucumber specs, sorry!

pirj:
Thank you!

PR #2745, opened 2024-03-15
Verify ActiveJob job signature matches arguments

odlp:
Hello,

I'd like to suggest an improvement to the `have_enqueued_job` and `have_been_enqueued` matchers – ensuring the arguments passed by the subject/caller actually match the job's method signature.

## Motivation

At the moment it's possible to write an implementation and spec which calls & expects the wrong number of arguments, or mix up the positional/keyword arguments. This results in a passing test, but the implementation will fail at runtime.

This improvement would guard against these false-positives. This approach was inspired by RSpec's awesome verifying doubles.

## Example

```ruby
class KeywordArgsJob < ActiveJob::Base
  def perform(a:, b:) # expecting keyword arguments
  end
end

RSpec.describe "ensuring job arguments match the job signature" do
  it "throws an exception on mismatch" do
    expect { KeywordArgsJob.perform_later(1, 2) } # incorrectly passing positional args
      .to have_enqueued_job(KeywordArgsJob).with(1, 2) # incorrectly expecting positional args
  end
end
```

- Current outcome: Spec passes
- Desired outcome: `ArgumentError: Missing required keyword arguments: a, b`

## Notes / questions

I wasn't sure if this behaviour should initially be opt-in, then become the default in the next major version. For the initial PR I've not made the verifying arguments behaviour configurable because it seems like any resulting failures will always be underlying bugs which should be fixed.

### DIFF
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -41,6 +41,8 @@ module RSpec
         end
 
         def failure_message
+          return @failure_message if defined?(@failure_message)
+
           "expected to enqueue #{base_message}".tap do |msg|
             msg << "\n#{unmatching_mail_jobs_message}" if unmatching_mail_jobs.any?
           end
@@ -89,19 +91,20 @@ module RSpec
           super(job)
         end
 
-        def verify_arguments_match_signature!(job, args)
+        def detect_args_signature_mismatch(jobs)
           return if @method_name.nil?
 
-          mailer_method = mailer_class_name.constantize.public_instance_method(@method_name)
-          mailer_args = args - base_mailer_args
+          mailer_class = mailer_class_name.constantize
 
-          if parameterized_mail?(job)
-            mailer_args = mailer_args[1..-1] # ignore parameterized params
-          elsif mailer_args.last.is_a?(Hash) && mailer_args.last.key?(:args)
-            mailer_args = args.last[:args]
+          jobs.each do |job|
+            mailer_args = extract_args_without_parameterized_params(job)
+
+            if (signature_mismatch = check_args_signature_mismatch(mailer_class, @method_name, mailer_args))
+              return signature_mismatch
+            end
           end
 
-          verify_signature!(mailer_method, mailer_args)
+          nil
         end
 
         def base_mailer_args
@@ -170,6 +173,19 @@ module RSpec
           else
             args + [hash]
           end
+        end
+
+        def extract_args_without_parameterized_params(job)
+          args = deserialize_arguments(job)
+          mailer_args = args - base_mailer_args
+
+          if parameterized_mail?(job)
+            mailer_args = mailer_args[1..-1] # ignore parameterized params
+          elsif mailer_args.last.is_a?(Hash) && mailer_args.last.key?(:args)
+            mailer_args = args.last[:args]
+          end
+
+          mailer_args
         end
 
         def legacy_mail?(job)
--- a/lib/rspec/rails/matchers/have_enqueued_mail.rb
+++ b/lib/rspec/rails/matchers/have_enqueued_mail.rb
@@ -89,6 +89,10 @@ module RSpec
           super(job)
         end
 
+        def verify_arguments_match_signature!(_job, _args)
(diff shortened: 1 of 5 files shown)


lib/rspec/rails/matchers/have_enqueued_mail.rb
```diff
@@ -89,6 +89,10 @@ def arguments_match?(job)
           super(job)
         end
 
+        def verify_arguments_match_signature!(_job, _args)
+          # Not implemented for mailers
+        end
```
odlp, line 94: This no-op in is a temporary placeholder because `HaveEnqueuedMail` inherits from `ActiveJob::HaveEnqueuedJob`.

If this PR is welcome I also have a branch which adds a similar verification of arguments to mailers:
https://github.com/odlp/rspec-rails/compare/pr-verify-job-perform-signature...odlp:rspec-rails:pr-verify-mailer-signature

Happy to raise a PR for this too

JonRowe: I'm happy for you to include that in this PR 👍

JonRowe (approved):
Awesome work, I'd love to see the mailer implementation included in this as well.

spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -457,6 +472,14 @@ def perform; raise StandardError; end
       }.to raise_error(/expected to enqueue exactly 1 jobs, but enqueued 0/)
     end
 
+    it "fails if the arguments do not match the job's signature" do
+      two_args_job.perform_later(1)
+
+      expect {
+        expect(two_args_job).to have_been_enqueued.with(1)
+      }.to raise_error(ArgumentError, /Wrong number of arguments/)
+    end
```
JonRowe, line 481: It would be great to have a keyword example vs positional arguments as well given the original context.

lib/rspec/rails/matchers/have_enqueued_mail.rb
odlp: Great, updated 👍

spec/rspec/rails/matchers/active_job_spec.rb
odlp: Good point, added additional examples for this

odlp:
Hmm... I think the CI failures are unrelated flakes:

```
expected that command "bin/rspec spec/system/widget_system_spec.rb" has exit status of "0", but has "9". (RSpec::Expectations::ExpectationNotMetError)
features/system_specs/system_specs.feature:109:in `the exit status should be 0'
```

Matrix is passing on my fork:
https://github.com/odlp/rspec-rails/actions/runs/8342458003

JonRowe:
Thanks, I have one last thing I'd like to change which is I'm not sure that it should raise an `ArgumentError`, this is a matcher and I think it should fail the matcher instead

odlp:
@JonRowe switched to the matcher failing in [232ea6b](https://github.com/rspec/rspec-rails/pull/2745/commits/232ea6b04c0bd9a653e0e2bbc69f5e1bbf6dc831). Turns out this is much neater when it fails:

```
Failures:

  1) ActiveJob matchers throws an exception on mismatch
     Failure/Error:
       expect { KeywordArgsJob.perform_later(1, 2) } # incorrectly passing positional args
         .to have_enqueued_job(KeywordArgsJob).with(1, 2) # incorrectly expecting positional args

       Incorrect arguments passed to KeywordArgsJob: Missing required keyword arguments: a, b
     # ./spec/rspec/rails/matchers/active_job_spec.rb:51:in `block (2 levels) in <top (required)>'
     # ./spec/rspec/rails/matchers/active_job_spec.rb:46:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:74:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:68:in `block (2 levels) in <top (required)>'
```

Thanks for the suggestion

JonRowe:
Thanks so much for this

odlp:
It was a pleasure - glad to make a small contribution back to a tool I use almost every day. Thanks for you reviews & guidance @JonRowe.

JonRowe:
Released in 7.0.0

PR #2746, opened 2024-03-16
Use headless driver for next Rails release

stevepolitodesign:
In the next release of Rails, the default driver was switched from `:chrome` to `:headless_chrome` as see in: https://github.com/rails/rails/pull/50512 This is to ensure the new [ci template][] will "work out of the box".

However, this will not work with applications using `rspec-rails`, since it still defaults to `:selenium`. Instead, GitHub actions will fail with the following error:

```
Selenium::WebDriver::Error::SessionNotCreatedError:
            session not created: Chrome failed to start: exited normally.
              (session not created: DevToolsActivePort file doesn't exist)
              (The process started from chrome location /usr/bin/google-chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

[ci template]: https://github.com/rails/rails/blob/main/railties/lib/rails/generators/rails/app/templates/github/ci.yml.tt

---

I've created a [demo application](https://github.com/stevepolitodesign/rspec-rails-ci-demo) to highlight the problem, and test the proposed solution. The [commit history](https://github.com/stevepolitodesign/rspec-rails-ci-demo/commits/main/) aims to break everything out change by change, but I will highlight the important parts below.

1. This [commit](https://github.com/stevepolitodesign/rspec-rails-ci-demo/commit/d82c3cbd80505a078796c22028c26ff53e5d02af) replicates the bug. Here's the [corresponding failure](https://github.com/stevepolitodesign/rspec-rails-ci-demo/actions/runs/8302149308/job/22723821219).
2. Without changing the test, I simply [update the Gemfile](https://github.com/stevepolitodesign/rspec-rails-ci-demo/commit/0aab647ad538212ced4f7143e2b55c1517022ffe) to use this branch, resulting in a passing test.

### DIFF
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -137,7 +137,11 @@ module RSpec
           self.class.before do
             # A user may have already set the driver, so only default if driver
             # is not set
-            driven_by(:selenium) unless @driver
+            if ::Rails::VERSION::STRING.to_f >= 7.2
+              driven_by(:selenium_chrome_headless) unless @driver
+            else
+              driven_by(:selenium) unless @driver
+            end
           end
         end
 
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -137,7 +137,7 @@ module RSpec
           self.class.before do
             # A user may have already set the driver, so only default if driver
             # is not set
-            if ::Rails::VERSION::STRING.to_f >= 7.2
+            if ::Rails::VERSION::STRING == "main"
               driven_by(:selenium_chrome_headless) unless @driver
             else
               driven_by(:selenium) unless @driver
--- a/lib/rspec/rails/example/system_example_group.rb
+++ b/lib/rspec/rails/example/system_example_group.rb
@@ -137,7 +137,7 @@ module RSpec
           self.class.before do
             # A user may have already set the driver, so only default if driver
             # is not set
-            if ::Rails::VERSION::STRING == "main"
+            if ::Rails::VERSION::STRING.to_f >= 7.2
               driven_by(:selenium_chrome_headless) unless @driver
             else
               driven_by(:selenium) unless @driver
--- a/spec/rspec/rails/example/system_example_group_spec.rb
+++ b/spec/rspec/rails/example/system_example_group_spec.rb
@@ -29,15 +29,15 @@ module RSpec::Rails
       end
     end
 
-    describe '#driver' do
-      it 'uses :selenium driver by default' do
+    describe '#driver', if: ::Rails::VERSION::STRING.to_f >= 7.2 do
+      it 'uses :selenium_chrome_headless driver by default' do
         group = RSpec::Core::ExampleGroup.describe do
           include SystemExampleGroup
         end
         example = group.new
         group.hooks.run(:before, :example, example)
 
-        expect(Capybara.current_driver).to eq :selenium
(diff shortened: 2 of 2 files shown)


stevepolitodesign:
Just pointing out that https://github.com/rails/rails/pull/51289 will remove the `test` job altogether. However, I think it's still important we keep parity with Rails.

JonRowe:
:wave: thanks!, can you update the spec to match?

stevepolitodesign:
@JonRowe I made a change in f122744 in an attempt to fix CI, but Rails `main` is [currently][] set to `7.2.0.alpha`. However, `ci.yml` uses `'main'` for the `RAILS_VERSION`.

[currently]: https://github.com/rails/rails/blob/5ab13c5a7798c2a9f96bebb1c68285dfb842d4f9/RAILS_VERSION

stevepolitodesign:
I think my imlpementation is correct, since CI is passing with my [demo application](https://github.com/stevepolitodesign/rspec-rails-ci-demo/actions/runs/8303634076). Does `ci.yml` need to change instead? Something like:

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index f54e46d7..92453045 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,10 +39,10 @@ jobs:
          # Edge Rails (?) builds >= 2.7
          - ruby: 3.2
            env:
-             RAILS_VERSION: 'main'
+             RAILS_VERSION: '~> 7.2.0'
          - ruby: 3.1
            env:
-             RAILS_VERSION: 'main'
+             RAILS_VERSION: '~> 7.2.0'
 
          # Rails 7.1 builds >= 2.7
          - ruby: 3.2
```

JonRowe:
The version specified in our `ci.yml` triggers fetching Rails from git, it doesn't affect the version string of Rails exposed to rspec, your original check was correct but it changed the default and theres a spec that checks it: spec/rspec/rails/example/system_example_group_spec.rb:33 what I meant was to conditionally define that and add the second spec like you have.

stevepolitodesign:
@JonRowe thank you for the additional context. I was surprised by the [original failure](https://github.com/rspec/rspec-rails/actions/runs/8303716484/job/22733796923#step:7:45) since I figured the default would not change because of the conditional. Do I need to do something like this?

Is there a way for me to test this locally?

```diff
diff --git a/spec/rspec/rails/example/system_example_group_spec.rb b/spec/rspec/rails/example/system_example_group_spec.rb
index a9b9ff83..1a19d932 100644
--- a/spec/rspec/rails/example/system_example_group_spec.rb
+++ b/spec/rspec/rails/example/system_example_group_spec.rb
@@ -29,7 +29,7 @@ module RSpec::Rails
       end
     end
 
-    describe '#driver' do
+    describe '#driver', if: ::Rails::VERSION::STRING.to_f < 7.2 do
       it 'uses :selenium driver by default' do
         group = RSpec::Core::ExampleGroup.describe do
           include SystemExampleGroup
```

stevepolitodesign:
@JonRowe let me know if my latest change is what you expected. My confusion lies in the fact that I would have expected there to be two specs:

- The existing `#driver` spec the ensures itss set to `:selenium`
- A new `#driver` spec ensuring it's set to `:selenium_chrome_headless` if `::Rails::VERSION::STRING.to_f >= 7.2`

However, using `#take_screenshot` as an example, it looks like it's only tested once, and with the conditional.

JonRowe:
Thanks!

JonRowe:
Released in 7.0.0

PR #2749, opened 2024-03-27
Prepend controller path to lookup_context prefixes

### LINKED ISSUE #2729: View example group inserts controller path into lookup_context in the wrong order

Ruby version: 3.2.3
Rails version: 7.1.3
RSpec version: 3.12.2

## Observed behaviour

Testing a controller where the 'prefixes' logic in ActionView::Renderer::AbstractRenderer#merge_prefix_into_object_path applies fails to find view partials when called from inside a nested controller.

This happens because the logic uses `@context_prefix = lookup_context.prefixes.first`, which in view specs is `""`, instead of the controller path.

## Expected behaviour

`@context_prefix = lookup_context.prefixes.first` should return the controller's path

Hard to reproduce as it relates to https://github.com/rails/rails/issues/50916 – the logic in `merge_prefix_into_object_path` isn't very useful. I've got a monkey patch for that logic that address the issue, which is how I came across this issue.

I believe the logic in `view_example_group` is faulty because it uses append instead of prepend when setting up the lookup context prefixes.

https://github.com/rspec/rspec-rails/blob/06d05aba209565f5bc940aa9e67232c8616e5197/lib/rspec/rails/example/view_example_group.rb#L189

sfnelson:
Rails rendering assumes that the first prefix encountered is 'special', specifically, it's the controller's prefix. This can be seen in `ActionView::AbstractRenderer::ObjectRendering#initialize`. This change injects the controller path at the start of the prefix list instead of the end to make sure this assumption is satisfied.

Resolves #2729.

### DIFF
--- a/lib/rspec/rails/example/view_example_group.rb
+++ b/lib/rspec/rails/example/view_example_group.rb
@@ -186,7 +186,7 @@ module RSpec
 
         before do
           _include_controller_helpers
-          view.lookup_context.prefixes << _controller_path
+          view.lookup_context.prefixes.prepend(_controller_path)
 
           controller.controller_path = _controller_path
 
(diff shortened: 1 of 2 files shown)


JonRowe:
:wave: Can you rebase this?

Issue #2713, opened 2023-11-16
Wrap examples with Rails executor


bensheldon:
**Describe the solution you'd like**

In #2644, Execution Contexts were cleared, but did not go to fully wrapping specs with a Rails Executor. In the current version of Rails (7.1) TestCases look like [this](https://github.com/bensheldon/rails/blob/42523a01dbff0b140f7554a519cec37f1b256f8b/activesupport/lib/active_support/railtie.rb#L52-L63):

```ruby
ActiveSupport.on_load(:active_support_test_case) do
  if app.config.active_support.executor_around_test_case
    require "active_support/executor/test_helper"
    include ActiveSupport::Executor::TestHelper
  else
    require "active_support/current_attributes/test_helper"
    include ActiveSupport::CurrentAttributes::TestHelper

    require "active_support/execution_context/test_helper"
    include ActiveSupport::ExecutionContext::TestHelper
  end
end
```

...and it would be nice if rspec-rails offered that too of using `ActiveSupport::Executor::TestHelper`

JonRowe:
See also #2712 for build failures attempting this caused.

javierjulio:
I found my way here for the CurrentAttributes reset portion but seems several of these things are tied together. Worth noting that even with a rails spec type (e.g. `type: :request`) CurrentAttributes are not reset in rspec-rails. I've created #2752 and #2753 to help address this. The former is a temporary fix considering the current state of the RailsExampleGroup to match what Rails does will reset CurrentAttributes for each example. The ideal fix is in the latter which should use either the executor or the individual test helpers like noted in the description here.

Issue #2503, opened 2021-05-29
RSpec does not reset ActiveSupport::CurrentAttributes around each example


mattbrictson:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-darwin20]
Rails version: 6.1.3.2
RSpec version: RSpec 3.10
  - rspec-core 3.10.1
  - rspec-expectations 3.10.1
  - rspec-mocks 3.10.2
  - rspec-rails 5.0.1
  - rspec-support 3.10.2

## Observed behaviour

rspec-rails does not reset `CurrentAttributes`. That means if one test sets a value using `CurrentAttributes`, it can very easily pollute other tests.

## Expected behaviour

Rails `ActiveSupport::TestCase` automatically resets `CurrentAttributes` between each test (see [active_support/current_attributes/test_helper.rb](https://github.com/rails/rails/blob/4f53cb3b8752f0c39c6766a07fb98fe030ebe467/activesupport/lib/active_support/current_attributes/test_helper.rb)). I would expect rspec-rails to do the same.

## Can you provide an example app?

Here is a trivial example app using `CurrentAttributes` and an `ActiveSupport::TestCase` that passes:

```ruby
require "rails"
require "rails/test_help"

class MyApp < Rails::Application
  config.eager_load = false
end

MyApp.initialize!

class Current < ActiveSupport::CurrentAttributes
  attribute :user_id
end

class CurrentTest < ActiveSupport::TestCase
  test "allows setting user_id" do
    Current.user_id = 123
    assert_equal(123, Current.user_id)
  end

  test "has a nil user_id by default" do
    assert_nil(Current.user_id)
  end
end
```

Test output:

```
$ bundle exec ruby testcase.rb 
Run options: --seed 52779

# Running:

..

Finished in 0.001012s, 1976.2846 runs/s, 1976.2846 assertions/s.
2 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

However the same app, when tested via rspec-rails, fails:

```ruby
require "rails"
require "action_controller/railtie"
require "action_view/railtie"
require "rspec/rails"

class MyApp < Rails::Application
  config.eager_load = false
end

MyApp.initialize!

class Current < ActiveSupport::CurrentAttributes
  attribute :user_id
end

describe Current do
  it "allows setting user_id" do
    Current.user_id = 123
    expect(Current.user_id).to eq(123)
[...]

PR #2752, opened 2024-04-06
Reset CurrentAttributes on each rails example

javierjulio:
We need to include the `ActiveSupport::CurrentAttributes::TestHelper` in the rails example group, otherwise rspec does not reset `ActiveSupport::CurrentAttributes` around each example unless being reset manually. We need to [mimic what Rails v7 does for test setup](https://github.com/rails/rails/blob/v7.0.0/activesupport/lib/active_support/railtie.rb#L50-L61) for the "else" case until #2712 can be revisited. Related to #2713. This should address #2503 properly.

In the meantime, I've created #2753 which recreates #2712 with some new changes since the GHA logs for the original are no longer available, we can't address without a new attempt. Ideally, #2753 would replace this with the suggestion from https://github.com/rspec/rspec-rails/pull/2753#issuecomment-2041110829 if it would work.

### DIFF
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -34,6 +34,8 @@ module RSpec::Rails
     end
 
     describe 'CurrentAttributes', order: :defined do
+      include RSpec::Rails::RailsExampleGroup
+
       class Current < ActiveSupport::CurrentAttributes
         attribute :request_id
       end
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -1,13 +1,7 @@
 module RSpec::Rails
   RSpec.describe RailsExampleGroup do
-    if ::Rails::VERSION::MAJOR >= 7
-      class CurrentSample < ActiveSupport::CurrentAttributes
-        attribute :request_id
-      end
-
-      it 'supports tagged_logger' do
-        expect(described_class.private_instance_methods).to include(:tagged_logger)
-      end
+    it 'supports tagged_logger', if: ::Rails::VERSION::MAJOR >= 7 do
+      expect(described_class.private_instance_methods).to include(:tagged_logger)
     end
 
     it 'does not leak context between example groups', if: ::Rails::VERSION::MAJOR >= 7 do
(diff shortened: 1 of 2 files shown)


javierjulio:
@pirj this is ready for review. Considering the current state of rspec-rails, I figure its best to start here since its basically missing this CurrentAttributes test helper inclusion to ensure parity with Rails. I've added tests here that I confirmed failed first and now pass with the updated module.

spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -1,6 +1,10 @@
 module RSpec::Rails
   RSpec.describe RailsExampleGroup do
     if ::Rails::VERSION::MAJOR >= 7
+      class CurrentSample < ActiveSupport::CurrentAttributes
```
pirj, line 4: What fails exactly without this?
At a glance it may looks that it’s the following example that check for tagged logger, but I guess it’s not, is it?

spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -32,5 +36,18 @@ module RSpec::Rails
 
       expect(results).to all be true
     end
+
+    describe 'CurrentAttributes', order: :defined, if: ::Rails::VERSION::MAJOR >= 7 do
```
pirj, line 40: The concept is hard to absorb at first, but here there are two types of groups, the first that is part of our spec suite, that is first evaluated when you run `rspec` over rspec-rails specs, but before running any specs. The second is defined during specs runtime, like `RSpec.describe` you can see above. I wouldn’t mix in RailsExampleGroup into the former, but it can be freely done with the latter. See example above for the reference.

spec/rspec/rails/example/rails_example_group_spec.rb
JonRowe: Its just setup...

JonRowe:
Thanks for tackling this, I took the liberty of refactoring the spec to what we'd usually do I hope you don't mind.

javierjulio:
@JonRowe I don't mind, I was just trying to figure it out and contribute as best I could. Thank you again!

spec/rspec/rails/example/rails_example_group_spec.rb
javierjulio: @pirj I didn't get to this in time. Seeing Jon's update, I'm just not familiar with how specs are written here and struggled a bit. I'll know how to better approach it in the future. Thanks!

JonRowe:
Released in 6.1.3

pond:
This was a great idea and well intentioned but regrettably it has introduced a serious bug.

Pleas see https://github.com/rspec/rspec-rails/issues/2773.

PR #2753, opened 2024-04-06
Wrap tests in Rails executor when configured or on Rails 7 by default

javierjulio:
This recreates #2712 since the GHA logs are no longer available which we need to address this issue. We should be wrapping the Rails examples with the executor to mimic what Rails v7 does (as noted in #2713) to properly reset state. This has been tested in a test suite for an app but we need to address what issues there is within rspec-rails. This change request originated from #2503 and #2752 which the latter is meant as a temporary fix since this is the expected way to reset state.


javierjulio:
To properly handle the Rails config switch I figure what really needs to happen here is the following but wasn't sure if this would work (the config conditional) at that level. If it does, then that should account for resetting state whether using the Rails executor or not.

```ruby
if ::Rails::VERSION::MAJOR >= 7
  if ::Rails.configuration.active_support.executor_around_test_case
    included do |_other|
      around do |example|
        ::Rails.application.executor.perform { example.call }
      end
    end
  else
    require 'active_support/current_attributes/test_helper'
    include ActiveSupport::CurrentAttributes::TestHelper

    require 'active_support/execution_context/test_helper'
    include ActiveSupport::ExecutionContext::TestHelper
  end
end
```

javierjulio:
Thank you for approving CI. I've checked many of the runs and they are failing for the same error.

```
     NameError:
       uninitialized constant RSpec::Rails::OpenStruct
     # ./spec/rspec/rails/example/view_example_group_spec.rb:190:in `controller'
     # ./spec/rspec/rails/example/view_example_group_spec.rb:196:in `block (3 levels) in <module:Rails>'
```

I figure this has to do with a new `json` gem version as I [encountered an OpenStruct error with turbo_tests recently](https://github.com/serpapi/turbo_tests/pull/49). I just don't know how to resolve it here since I don't see any `Gemfile.lock` files in this repo to confirm. In the meantime, I'll push an update to `require ostruct` in the reported file.

javierjulio:
@pirj thank you. This is ready for another test run.

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -1,3 +1,4 @@
+require 'ostruct'
```
pirj, line 1: I’ve seen this started failing in other prs, too.
I haven’t had time to have a cloaer look.
If  we include it in our specs, would’t this cause failurea in user projects’ specs?

Ok to fix this in a separate pr?

pirj (commented):
Looks good, do you kind adding a spec that would prevent regressions?
I guess setting something Current and two examples in the same context with oder: :defined should do it.

spec/rspec/rails/example/view_example_group_spec.rb
javierjulio: @pirj thank you. Yes, I've created #2754 to hopefully fix the error. It has been failing elsewhere. More details in that PR.

javierjulio:
@pirj yes, I will reuse the tests from #2752 here. I need help with handling the `executor_around_test_case` config since there are multiple cases here that I think we need to handle, unless you feel otherwise.

Cases to consider:
- Rails 7+ and executor_around_test_case is true (or load_defaults is 7.0)
- Rails 7+ and executor_around_test_case is false
- Rails 6.1 and below

This could handle the first and third case, but the second one is tricky because how would we include/use the two test helpers inside the around block?

lib/rspec/rails/example/rails_example_group.rb
```diff
@@ -2,11 +2,6 @@
 # suite and ammeter.
 require 'rspec/rails/matchers'
 
-if ::Rails::VERSION::MAJOR >= 7
-  require 'active_support/current_attributes/test_helper'
-  require 'active_support/execution_context/test_helper'
-end
```
JonRowe, line 8: Why are we doing this? Is there harm in having these always required?

javierjulio: I will update these soon, thanks. There shouldn't be any harm. I just went with one approach here as I was just starting out without having an understanding from rspec's perspective at the time.

javierjulio: This was updated so the require statements are back in the original location with no diff and just the include statements later in the file. Thanks!

ngan:
This worked for us:
```ruby
RSpec::Core::Example.prepend(
  Module.new do
    def run(...)
      Rails.application.executor.perform { super }
    end
  end
)
```

I think we'd have to do the same here since context-level before hooks run _before_ around hooks. Essentially, modify `Example#run`.

ngan:
One thing to note is wrapping tests _drastically_ degraded test performance for us. I'll have numbers soon.

PR #2754, opened 2024-04-08
Fix uninitialized constant RSpec::Rails::OpenStruct error

javierjulio:
This started to occur on recent test suite runs e.g. #2752, #2753. It may be related to a `json` gem update ([example](https://github.com/serpapi/turbo_tests/pull/49)) that removed the `ostruct` dependency as I encountered that with `turbo_tests` which I use to run my rspec test suite.

### DIFF
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -1,3 +1,4 @@
+require 'ostruct'
 require 'support/group_failure_formatter'
 
 module RSpec::Rails
@@ -187,7 +188,7 @@ module RSpec::Rails
         Class.new do
           include ViewExampleGroup::ExampleMethods
           def controller
-            @controller ||= OpenStruct.new(params: nil)
+            @controller ||= ::OpenStruct.new(params: nil)
           end
         end.new
       end


pirj:
Thank you!

javierjulio:
@pirj thank you for the quick turnaround! I'll go back to update the other PR and see what errors remain.

JonRowe:
@pirj this needs to be reverted or the require moved to a sub process... it could cause dependency based issues for us / our users

javierjulio:
@JonRowe I can try to help. Why would it need to be reverted if the require is just in a spec file that only affects the test suite? I can submit a revert if it's really necessary. Do you have an example of what I can follow to do that in a safe manner?

pirj:
We could just use a Class.new to create an interface that that view helper needs.

javierjulio:
That would be ideal over using OpenStruct since it's been a long time now that Ruby has advised to not use it. I'm not familiar with the requirements for this project so likely missing something. Can I help with a PR to do that instead?

spec/rspec/rails/example/view_example_group_spec.rb
```diff
@@ -187,7 +188,7 @@ def _default_file_to_render; end # Stub method
         Class.new do
           include ViewExampleGroup::ExampleMethods
           def controller
-            @controller ||= OpenStruct.new(params: nil)
+            @controller ||= ::OpenStruct.new(params: nil)
```
pirj, line 191: I think this can become just `Struct.new(:params).new(nil)`, and we’re openstruct-free with no side effects.

javierjulio: Ok I will submit a PR with that change to start the replacement.

javierjulio: @pirj I made this change in #2755

JonRowe:
> the require 'ostruct' is just in a spec file that only affects the test suite?

The reason for this policy is that it means we could inadvertandly then use openstruct and cause breakages.

javierjulio:
Yes, I think it's better too to avoid OpenStruct if possible. Thanks!

PR #2755, opened 2024-04-09
Replace OpenStruct dependency with Struct

javierjulio:
Instead of reverting #2754 since it will still fail, this is a new attempt at replacing the OpenStruct dependency with a Struct from this suggestion: https://github.com/rspec/rspec-rails/pull/2754#discussion_r1556524959

### DIFF
--- a/spec/rspec/rails/example/view_example_group_spec.rb
+++ b/spec/rspec/rails/example/view_example_group_spec.rb
@@ -1,4 +1,3 @@
-require 'ostruct'
 require 'support/group_failure_formatter'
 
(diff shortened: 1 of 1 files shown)


javierjulio:
@pirj this is ready and looking good. Can you please re-run the failed CI jobs? Those should pass. Reviewing each, it's the same failure and I've seen on other builds so either a flaky and/or unrelated. Thank you for the suggestion. It's better than depending on OpenStruct.

pirj:
Thank you. This is indeed much better.

PR #2759, opened 2024-04-18
`with_priority` matcher for ActiveJob

mbajur:
Fixes https://github.com/rspec/rspec-rails/discussions/2757

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -31,7 +31,7 @@ module RSpec
             self
           end
 
-          def with_priority(priority)
+          def at_priority(priority)
             @priority = priority.to_i
             self
           end
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -152,8 +152,8 @@ module RSpec
             msg_parts << "on queue #{job[:queue]}" if job[:queue]
             msg_parts << "at #{Time.at(job[:at])}" if job[:at]
             msg_parts <<
-              if job[:priority]
-                "with priority #{job[:priority]}"
+              if fetch_priority(job)
+                "with priority #{fetch_priority(job)}"
               else
                 "with no priority specified"
               end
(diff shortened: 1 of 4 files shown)


spec/rspec/rails/matchers/active_job_spec.rb
```diff
@@ -262,6 +262,18 @@ def perform; raise StandardError; end
...
+        hello_job.set(priority: 2).perform_later
+      }.to have_enqueued_job.with_priority(2)
+    end
+
+    it "passes with provided priority number as string" do
+      expect {
+        hello_job.set(priority: 2).perform_later
+      }.to have_enqueued_job.with_priority("2")
+    end
+
```
pirj, line 276: Can you please add a fee more cases? I can think of a failure case when we expect priority and it wasn’t set. Also when it was set, but doesn’t match the expectation. In both examples with checks if our failure message is reasonable.
I had another scenario on the tip of my tongue, but it keeps slipping. Can you think of other cases we should cover? You can use other similar qualifiers as a source of inspiration .

By the way, does this apply to mailers?

pirj:
It also makes sense to add/extend scenario here https://github.com/rspec/rspec-rails/blob/4ab08fa954fe4e0ae6f2302fa7201dcad3f11521/features/job_specs/job_spec.feature#L63 
Those changes will appear here https://rspec.info/features/6-1/rspec-rails/job-specs/job-spec/, and some prefer those docs to the rdoc ones.

lib/rspec/rails/matchers/active_job.rb
```diff
@@ -30,6 +31,11 @@ def on_queue(queue)
             self
           end
 
+          def with_priority(priority)
```
JonRowe, line 34: Personally I would call this `priority` or `at_priority` to disambigute from `with`

cmrd-senya:
ping @mbajur

pirj:
Thank you @mbajur !

JonRowe:
Released in 7.0.0

PR #2764, opened 2024-06-14
fix: link at mailer preview generator template

chiaraani:
When I execute:
```
rails g mailer users/confirmations
```
The link to the preview at the comment is this:
```
http://localhost:3000/rails/mailers/users/confirmations
```

This route raises this error:
```
Unknown action
Mailer preview 'users/confirmations' not found
```
So I fixed it.


pirj:
Do you think it’s possible to write a failing spec for this?

JonRowe:
Thanks, merging despite build failure as its just a template typo

chiaraani:
Is it necessary to add a test for a generated code comment?

JonRowe:
Personally I don't think so, it would be hard to do in a way that wasn't brittle.

JonRowe:
Released in 6.1.3

PR #2765, opened 2024-06-17
HaveHttpStatus uses Rack's public API

### LINKED ISSUE #2763: Rack status code name changes breaks HttpStatusMatcher

Ruby version: 3.3.3
Rails version: 7.1.3.4
RSpec version: 3.13.0
Rack version: 3.1.3

## Observed behaviour

Rack has recently changed the name of the 422 status code from `unprocessable_entity` to `unprocessable_content`.  However, both names will still resolve *if* Rack's public API is being used.  However, the `HttpStatusMatcher` is looking at the Rack Constant directly, instead of using the provided API.

My suggestion is for the code in `have_http_status.rb` that looks like (around line 218)...

```ruby
def set_expected_code!
  @expected ||=
    Rack::Utils::SYMBOL_TO_STATUS_CODE.fetch(expected_status) do
      raise ArgumentError,
            "Invalid HTTP status: #{expected_status.inspect}"
    end
end
```

be changed to....

```ruby
def set_expected_code!
  @expected ||= Rack::Utils.status_code(expected_status)
end
```

Here's the behavior of the `Rack::Utils.status_code` method...

```
irb> Rack::Utils.status_code(:unprocessable_content)
  => 422
irb> Rack::Utils.status_code(:unprocessable_entity)
  => 422
irb> Rack::Utils.status_code(:invalidcode)
(irb):4:in `<main>': Unrecognized status code :invalidcode (ArgumentError)
```

darrenboyd:
Rack has recently changed the name of the 422 status code from `unprocessable_entity` to `unprocessable_content`. Since `HaveHttpStatus` was looking up the code directly through the constant in Rack, this forced developers to rename this value in their codebase in the same change upgrade Rack.  Using Rack's public API method (`Rack::Utils.status_code`) allows users to transition from the old name to the new name on their own schedule.

This change fixes #2763

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@ Bug Fixes:
 
 * Reset `ActiveSupport::CurrentAttributes` between examples. (Javier Julio, #2752)
 * Fix a broken link in generated mailer previews. (Chiara Núñez, #2764)
+* Have HttpStatusMatcher use Rack's public API for http status code. (Darren Boyd, #2763)
 
 ### 6.1.2 / 2024-03-19
 [Full Changelog](https://github.com/rspec/rspec-rails/compare/v6.1.1...v6.1.2)
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -216,11 +216,7 @@ module RSpec
           # @see Rack::Utils::SYMBOL_TO_STATUS_CODE
           # @raise [ArgumentError] if an associated code could not be found
           def set_expected_code!
-            @expected ||=
-              Rack::Utils::SYMBOL_TO_STATUS_CODE.fetch(expected_status) do
-                raise ArgumentError,
-                      "Invalid HTTP status: #{expected_status.inspect}"
-              end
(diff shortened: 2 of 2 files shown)


pirj:
Unrelated to your PR, but the failures are caused by the lack of [this](https://github.com/rails/rails/blob/9098f535b2accef27c6c284dede119c936c44595/actionpack/lib/action_dispatch/system_testing/browser.rb#L76) check [in 7.1.3.4](https://github.com/rails/rails/blob/19eebf6d33dd15a0172e3ed2481bec57a89a2404/actionpack/lib/action_dispatch/system_testing/browser.rb#L73).

I suppose we’d have to pin selenium-webdriver to ~> 4.19.0 [here](https://github.com/rspec/rspec-rails/blob/e65348e2675695c37c1cd8c57e7648cfed1c0535/Gemfile-rails-dependencies#L38) to fix this.

pirj:
May I kindly ask you to rebase to include the fix and also add a changelog entry?

pirj:
Thank you!

pirj:
2.7/7.1 looks flaky

JonRowe:
Released in 6.1.3

PR #2766, opened 2024-06-17
Pin selenium-webdriver to fix a warning

pirj:
Rails 7.1 is unaware of the new API and this crashes our CI, eg https://github.com/rspec/rspec-rails/actions/runs/9552072568/job/26328604445?pr=2765

### DIFF
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -44,7 +44,7 @@ else
   gem 'activerecord-jdbcsqlite3-adapter', platforms: [:jruby]
 
   if version_number == '7.1.0'
-    gem 'selenium-webdriver', '~> 4.19.0', require: false
+    gem 'selenium-webdriver', '< 4.20.0', require: false
   else
     gem 'selenium-webdriver', require: false
   end
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -37,9 +37,15 @@ when nil, false, ""
   gem 'activerecord-jdbcsqlite3-adapter', platforms: [:jruby]
   gem 'selenium-webdriver', require: false
 else
-  add_net_gems_dependency if version.split(' ').last < '7.0'
+  version_number = version.split(' ').last
+  add_net_gems_dependency if version_number < '7.0'
 
   gem "rails", version
   gem 'activerecord-jdbcsqlite3-adapter', platforms: [:jruby]
-  gem 'selenium-webdriver', require: false
+  
+  if version_number == '7.1'
+    gem 'selenium-webdriver', '~> 4.19.0', require: false
+  else
+    gem 'selenium-webdriver', require: false
+  end
 end
--- a/Gemfile-rails-dependencies
+++ b/Gemfile-rails-dependencies
@@ -43,7 +43,7 @@ else
   gem "rails", version
   gem 'activerecord-jdbcsqlite3-adapter', platforms: [:jruby]
 
-  if version_number == '7.1'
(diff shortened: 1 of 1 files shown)


pirj:
2.7 is broken with a `uninitialized constant Gem` apparently causing Puma not to get loaded.

pirj:
Which is confusing, as it seems to load rack/handler/puma, but the Rack::Handler is not loaded. In Puma 5, that file was [loading rack/handler and defined Rack::Handler::Puma](https://github.com/puma/puma/blob/f306d2e712ff5f37fcfe76e4666abb2ff5d046ee/lib/rack/handler/puma.rb#L3), but in 6.x this happens only if [Rackup is not loaded](https://github.com/puma/puma/blob/5fc43d73b6ff193325e657a24ed76dec79133e93/lib/rack/handler/puma.rb#L129).
But the [comment says](https://github.com/puma/puma/blob/5fc43d73b6ff193325e657a24ed76dec79133e93/lib/rack/handler/puma.rb#L117) that rackup was removed from rack 3, and [evidently we install rack 3](https://github.com/rspec/rspec-rails/actions/runs/9552708407/job/26329890643?pr=2766#step:6:34).

pirj:
I’ve merged this as it fixes Ruby 3.x builds with Rails 7.1.
Ruby 2.7 with Rails 7.1 runs into a different issue, which I hope to look into next.

pirj:
In both cases, for 2.7 and for 3.x with Rails 7.1, we bundle puma 6.4.2 and also rackup 2.1.0. This means, Rack::Handler::Puma should not be defined. Unless there’s some race condition/load order dependency and we: 1) load rack/handler 2) load puma’s rack/handler/puma 3) (optionally) load rackup.
Where does the rackup dependency even come from?

pirj:
The `` can be caused [by](https://github.com/rubygems/rubygems/blob/ec2089640dbbb0de136c77190cc20d78aadada15/lib/rubygems/deprecate.rb#L72) `module Gem::Deprecate` that was later [changed](https://github.com/rubygems/rubygems/commit/26b9f3691bad29b289a583db041caf47d871496f#diff-8460330c1d34360b650e57e625c5154ba2d0d2d217b5bb1fbe6cbda77dcbb351R23) into a nested declaration. But this change is 11 years old, why it’s not in rubygems [3.5.9](https://github.com/rubygems/rubygems/blob/4304697e0c5598f5f018484f88a81717e25f1b24/lib/rubygems/deprecate.rb#L72) or 3.4.22 that are recent? 🤔 
It appeared in 3.5.10 in [this PR](https://github.com/rubygems/rubygems/pull/7612/files).
Apparently, the namespace definition was collapsed [here](https://github.com/rubygems/rubygems/commit/9bb539a756d1ae05ae23d0da557de18d62ace763#diff-8460330c1d34360b650e57e625c5154ba2d0d2d217b5bb1fbe6cbda77dcbb351R23) more than a decade ago.
I hope this is just a warning with no consequences.

pirj:
What’s interesting is that in [another file](https://github.com/puma/puma/blob/4172f187b8fadcc9db641dfd1c109b9630243960/lib/puma/rack_default.rb#L14) there's similar code, but it suggests that rack > 3 must be used with rackup gem.
This whole problem seems due to rack::handler not being loaded when capybara registers its puma server, and puma registers itself as a rack handler.
Wondering, what makes a difference between 2.7 and 3.x?

JonRowe:
The problem is capybara fixed this in `3.40.0` but also pinned the base ruby version to 3.0

PR #2770, opened 2024-06-21
docs: Use `perform_enqueued_jobs` in an example code

toshimaru:
I think `perform_jobs` should be `perform_enqueued_jobs`,
so I fixed the documentation.

```rb
  test "account is charged" do
    perform_enqueued_jobs do
      BillingJob.perform_later(account, product)
    end
    assert account.reload.charged_for?(product)
  end
```

ref. [Testing Rails Applications — Ruby on Rails Guides](https://guides.rubyonrails.org/testing.html#testing-jobs-in-isolation)

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -413,33 +413,33 @@ module RSpec
       #
       # @example
       #     expect {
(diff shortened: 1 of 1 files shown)


pirj:
Thanks!

Issue #2502, opened 2021-05-23
get transforms to post if params given


exocode:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: `ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin20]`
Rails version: `Rails 6.1.3.1`
RSpec version:

```
RSpec 3.10
  - rspec-core 3.10.1
  - rspec-expectations 3.10.1
  - rspec-mocks 3.10.2
  - rspec-rails 5.0.1
  - rspec-support 3.10.2
 ```

## Observed behaviour
routes.rb

`  get '/users/:id/billing_addresses', to: "billing_addresses#index"`

my spec.rb
```
 get "/users/#{user.id}/billing_addresses",
          params: { some_param: some_param }, # <---- transforms get request to a post request
          headers: valid_headers,
          as: :json
```
`Started POST "/users/2/billing_addresses" `

If I remove `params: { some_param: some_param }` like so:

```
 get "/users/#{user.id}/billing_addresses",
          headers: valid_headers,
          as: :json
```

it keeps a get request
``Started GET "/users/2/billing_addresses" ``

## Expected behaviour

```
 get "/users/#{user.id}/billing_addresses",
          headers: valid_headers,
          as: :json
```
should trigger 

`Started GET "/users/2/billing_addresses" `

pirj:
`get` is a Rails helper method. I suggest you to check with them.
GET requests semantically do not include body, and `as:` implies you're setting body's content type.
Do you mean to say "request JSON results"?

benoittgt:
Hello

Did you tried with the parameter as a hash? 

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "rails", "6.1.0"
  gem "rspec-rails"
end

require "rspec-rails"
require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  config.session_store :cookie_store, key: "cookie_store_key"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  routes.draw do
    get "/" => "test#index"
  end
end

class TestController < ActionController::Base
  include Rails.application.routes.url_helpers

  def index
[...]

PR #2771, opened 2024-06-25
Support `have_http_status` with `Rack::MockResponse`

cbliard:
When using `rack-test`, doing something like `expect(last_response).to have_http_status(:ok)` would fail with the error "expected a response object, but an instance of Rack::MockResponse was received".

This PR creates a `ActionDispatch::TestResponse` from the `Rack::MockResponse` to make `have_http_status` usable in specs relying on `rack-test` to perform requests.

It should fix rubocop/rubocop-rspec_rails#20.

Reproduction script:
```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  # Activate the gem you are reporting the issue against.
  gem "rails", "6.1.0"
  gem "rspec-rails"
end

require "rspec-rails"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  config.session_store :cookie_store, key: "cookie_store_key"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  routes.draw do
    get "/" => "test#index"
  end
end

class TestController < ActionController::Base
  include Rails.application.routes.url_helpers

  def index
    render plain: "Hello, World!"
  end
end

require "rspec/autorun"
require "rspec/rails/matchers"

RSpec.describe "additions", type: :request do
  include Rack::Test::Methods
  include RSpec::Rails::Matchers

  it do
    get "/"
    expect(last_response).to have_http_status(:ok)
  end

  private

  def app
    Rails.application
  end
end
```

### DIFF
--- a/lib/rspec/rails/matchers/have_http_status.rb
+++ b/lib/rspec/rails/matchers/have_http_status.rb
@@ -33,7 +33,7 @@ module RSpec
         # @param obj [Object] object to convert to a response
         # @return [ActionDispatch::TestResponse]
         def as_test_response(obj)
-          if ::ActionDispatch::Response === obj
+          if ::ActionDispatch::Response === obj || ::Rack::MockResponse === obj
             ::ActionDispatch::TestResponse.from_response(obj)
           elsif ::ActionDispatch::TestResponse === obj
             obj
--- a/spec/rspec/rails/matchers/have_http_status_spec.rb
+++ b/spec/rspec/rails/matchers/have_http_status_spec.rb
@@ -16,6 +16,14 @@ RSpec.describe "have_http_status" do
       end
     end
 
+    context "given a Rack::MockResponse" do
+      it "returns true for a response with the same code" do
+        response = ::Rack::MockResponse.new(code, {}, "")
+
+        expect(matcher.matches?(response)).to be(true)
+      end
+    end
+
     context "given an ActionDispatch::TestResponse" do
       it "returns true for a response with the same code" do
         response = ::ActionDispatch::TestResponse.new(code).tap { |x|


lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -49,6 +49,12 @@ def as_test_response(obj)
               resp.request = ActionDispatch::Request.new({})
             end
             ::ActionDispatch::TestResponse.from_response(obj)
+          elsif obj.respond_to?(:status) && obj.respond_to?(:headers) && obj.respond_to?(:body)
```
pirj, line 52: rack-test is a [runtime dependency](https://rubygems.org/gems/actionpack) of actionpack. How comes it won't be loaded?

pirj: PS which is weird, because rack-test will also load in the production env.

pirj (commented):
Can you please provide a reproduction script a-la https://github.com/rspec/rspec-rails/issues/2502#issuecomment-846604540 ?

lib/rspec/rails/matchers/have_http_status.rb
```diff
@@ -49,6 +49,12 @@ def as_test_response(obj)
...
+            # Hack to support `Rack::MockResponse` without having to load
+            # Rack or catch `NameError`s for the undefined constants
+            resp = ::ActionDispatch::Response.new(obj.status, obj.headers, obj.body)
+            resp.request = ::ActionDispatch::Request.new({})
```
JonRowe, line 56: What was your intent with this? Its discarded immediately and isn't necessary to pass the spec you've written.

lib/rspec/rails/matchers/have_http_status.rb
JonRowe: I think the comment is copy-pasta from the other implementation

cbliard:
> Can you please provide a reproduction script a-la [#2502 (comment)](https://github.com/rspec/rspec-rails/issues/2502#issuecomment-846604540) ?

I added one in the description. Thanks for pointing to a nice example.

lib/rspec/rails/matchers/have_http_status.rb
cbliard: I assumed that `Rack::MockResponse` was not available, but I did not check.

Now that I have checked, it is provided by [rack](https://github.com/rack/rack/blob/main/lib/rack/mock_response.rb) itself, so I can simplify the condition.

JonRowe (changes requested):
👋 Thanks for drawing this to our attention, I think the simpler solution would be to both construct a real `Rack::MockResponse` in the test, and just match on that earlier.

lib/rspec/rails/matchers/have_http_status.rb
cbliard: You're right.

I initially wrote it a lot like the above condition to support Capybara Session. It looked like this:
```ruby
obj = ActionDispatch::Response.new.tap do |resp|
  resp.status = obj.status
  resp.headers.clear
  resp.headers.merge!(obj.headers)
  resp.body = obj.body
  resp.request = ActionDispatch::Request.new({})
end
::ActionDispatch::TestResponse.from_response(obj)
```

Then I refactored to use the initializer:

```ruby
obj = ActionDispatch::Response.new(obj.status, obj.headers, obj.body).tap do |resp|
  resp.request = ActionDispatch::Request.new({})
end
::ActionDispatch::TestResponse.from_response(obj)
```

Then I got rid of the `tap` which does not add much and renamed the variable:
```ruby
resp = ActionDispatch::Response.new(obj.status, obj.headers, obj.body)
resp.request = ActionDispatch::Request.new({})
::ActionDispatch::TestResponse.from_response(obj)
```

And I forgot to rename `obj` to `resp` in the last line.

As it works that way, I'll simplify the code.

cbliard:
Thanks @pirj and @JonRowe for your reviews and suggestions. I just pushed a fix.

JonRowe:
Thanks!

JonRowe:
Released in 6.1.4

PR #2774, opened 2024-07-04
Rework reset behaviour of ActiveSupport::CurrentAttributes

### LINKED ISSUE #2773: #2752 regrettably introduces a serious bug

Ruby version: 3.3.3
Rails version: 7.1
RSpec version: 6.1.3 (vs working, 6.1.2)

## Observed behaviour

The reset behaviour occurs _in between_ the suite's `config.around :each` (e.g. in `spec_helper.rb`) and any `context`/`describe` block's `before :each ...` (or more generally, before examples then run). This breaks any tests which had current attributes set up in the config, which tests then expect later. This bug is the reason for https://github.com/ErwinM/acts_as_tenant/issues/338.

## Expected behaviour

We strongly suggest that the reset should happen either _after_ all other "each"-style example hooks have concluded, or _before_ any "each"-style example hooks start. It should never happen invisibly _in between_ "each"-style hooks.

Yes. A tiny stripped down almost-Rails application with `README.md` containing additional information and a replication test case is included.

[rspecbug.tar.gz](https://github.com/user-attachments/files/16061697/rspecbug.tar.gz)

pond:
Compromise solution to [#2773](https://github.com/rspec/rspec-rails/issues/2773).
Fixes #2773

* TL;DR - only resets `ActiveSupport::CurrentAttributes` _after_ examples run, rather than both before and after.

Consider a `spec_helper.rb` that uses something like [ActsAsTenant](http://rubygems.org/gems/acts_as_tenant) to wrap all examples, or anything similar where a block is used in order for whatever it is you're calling to handle cleanup itself automatically when the block finishes executing.

```ruby
config.around :each do | example |
  ActsAsTenant.without_tenant do
    example.run()
  end
end
```

In RSpec Rails 6.1.3, inclusion of `ActiveSupport::CurrentAttributes::TestHelper` means that attributes get reset both (in Rails terms) before setup *and* after teardown. What this means for RSpec hooks is that the resets happen *between* the suite-wide configuration "around" hooks and any examples or contexts. So, if we subsequently have an innocent test like this:

```ruby
it 'does something that requires ActsAsTenant.without_tenant to be asserted' do
  some_operation()
end
```

...then the test will work with RSpec Rails 6.1.2 or earlier, but fail with 6.1.3, because ActsAsTenant uses `ActiveSupport::CurrentAttributes` under the hood and its no-tenancy assertion will have been reset between the "around" hook and the test run.

The most robust solution would to be to try and insert an RSpec around-hook that is at the very end of the chain, and resets *after* running the example. This is difficult to do cleanly within RSpec Rails and [there is a desire to not drift too far from what Rails does itself](https://github.com/rspec/rspec-rails/issues/2773#issuecomment-2205345723). Therefore, the compromise suggested here uses the "adapters" module pattern in RSpec Rails to add an adapter for attribute reset that does the after-teardown reset only. The compromise here is that yes, this still happens after any examples are finished, but going back to that code above:

```ruby
config.around :each do | example |
  ActsAsTenant.without_tenant do
    example.run()
  end
end
```

...once `example.run()` returned, code that cared to check would find attributes were now already reset. This is considered pretty unlikely, and a better compromise than the known and hard to fix use cases of things like asserting tenancy, or a logged in user suite-wide in an around-each hook. I'm not aware of cases where a suite-wide configured around-each hook would care *after* an example about attribute data that was set within it (except perhaps to verify that it was reset - and it will be!). Bear in mind that this behaviour is present in RSpec Rails 6.1.3 anyway since it also has the teardown callback; this patch just avoids the on-setup reset.

Test coverage is included and it's a bit gnarly because we do have to simulate an RSpec-configured "around each" hook without accidentally polluting subsequent tests with that configuration change (whether this would be harmless or not, it's bad behaviour and would very fractionally slow down the execution speed of RSpec Rails's test suite).

### DIFF
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -137,6 +137,45 @@ module RSpec
     end
 
     # @private
+    #
+    # ActiveSupport::CurrentAttributes::TestHelper uses both before-setup and
+    # after-teardown hooks to clear attributes. These are run in an awkward
+    # order relative to RSpec hooks. Client suite-configured "around each"
+    # hooks run first, the attribute reset hooks next, and contexts or examples
+    # (along with whatever their defined hooks might be) run last. Consequently
+    # a suite which sets up around-each-example attributes could fail, because
+    # the attributes would have been reset by time tests actually run.
+    #
+    # Reworking with "before" rather than "around" hooks sidesteps the problem.
+    # Unfortunately, "executes via a block to clean up after itself" cases such
+    # as with "ActsAsTenant.without_tenant ..." make such a rework either hard
+    # or impossible. Besides, even realising that you may need to do this can
+    # take some significant debugging effort and insight.
+    #
+    # The compromise is to use "half" of what ActiveSupport does - just add the
+    # after-hook. Attributes are now only reset after examples run, but still
+    # before a suite-wide "around" hook's call to "example.run()" returns. If a
+    # test suite depended upon current attribute data still being set *after*
+    # examples within its suite-wide "around" hook, it would fail. This is
+    # unlikely; it's overwhelmingly common to just *set up* data before a test,
+    # and usually the only things you're doing afterwards are cleanup.
+    #
+    # Include this module to reset under the above compromise, instead of
+    # including ActiveSupport::CurrentAttributes::TestHelper.
+    #
+    module ActiveSupportCurrentAttributesAdapter
+      extend ActiveSupport::Concern
+
+      included do
+        def after_teardown
+          puts "AFTER TEARDOWN"
+          super
+          ActiveSupport::CurrentAttributes.reset_all
+        end
+      end
+    end
+
+    # @private
     module MinitestAssertionAdapter
       extend ActiveSupport::Concern
 
--- a/lib/rspec/rails/adapters.rb
+++ b/lib/rspec/rails/adapters.rb
@@ -138,30 +138,10 @@ module RSpec
 
     # @private
     #
-    # ActiveSupport::CurrentAttributes::TestHelper uses both before-setup and
-    # after-teardown hooks to clear attributes. These are run in an awkward
-    # order relative to RSpec hooks. Client suite-configured "around each"
-    # hooks run first, the attribute reset hooks next, and contexts or examples
-    # (along with whatever their defined hooks might be) run last. Consequently
-    # a suite which sets up around-each-example attributes could fail, because
-    # the attributes would have been reset by time tests actually run.
+    # Reset ActiveSupport::CurrentAttributes only *after* tests run, rather
+    # than before *and* after.
     #
-    # Reworking with "before" rather than "around" hooks sidesteps the problem.
-    # Unfortunately, "executes via a block to clean up after itself" cases such
-    # as with "ActsAsTenant.without_tenant ..." make such a rework either hard
-    # or impossible. Besides, even realising that you may need to do this can
-    # take some significant debugging effort and insight.
-    #
-    # The compromise is to use "half" of what ActiveSupport does - just add the
-    # after-hook. Attributes are now only reset after examples run, but still
-    # before a suite-wide "around" hook's call to "example.run()" returns. If a
-    # test suite depended upon current attribute data still being set *after*
-    # examples within its suite-wide "around" hook, it would fail. This is
-    # unlikely; it's overwhelmingly common to just *set up* data before a test,
-    # and usually the only things you're doing afterwards are cleanup.
-    #
-    # Include this module to reset under the above compromise, instead of
-    # including ActiveSupport::CurrentAttributes::TestHelper.
+    # See https://github.com/rspec/rspec-rails/pull/2774 for more details.
     #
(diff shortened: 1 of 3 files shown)


spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -56,5 +56,85 @@ class CurrentSample < ActiveSupport::CurrentAttributes
...
+
+        # Client code might legitimately want to wrap examples to ensure
+        # all-conditions tidy-up, e.g. "ActsAsTenant.without_tenant do...",
+        # wherein an "around" hook is the only available solution, often used
+        # in the overall suite via "config.around". Tests would not expect
+        # anything set in CurrentAttributes here to suddenly be reset by the
+        # time their actual tests, or their test hooks ran.
+        #
+        RSpec.configure do | config |
+          config.around(:each, uniquely_identifiable_metadata()) do | example |
```
pirj, line 85: You can use [in_sub_process](https://github.com/rspec/rspec-support/blob/23b3756f27ba0a506aa014dfd3e310c2bcd5f0a3/lib/rspec/support/spec/in_sub_process.rb#L12) to avoid config leaks between examples.

To differentiate between our examples and examples under test, we usually `let(:example_group) { RSpec.describe("test example group") }` in specs.

Otherwise, 👍

pond: Had no idea about `in_sub_process`. About to head home for the day but I'll look into that soon as it's likely to be very much cleaner. Thanks!

pond: ...Actually, that was easy. Change pushed already.

lib/rspec/rails/adapters.rb
```diff
@@ -136,6 +136,24 @@ def method_name
       end
     end
 
+    # @private
+    #
+    # Reset ActiveSupport::CurrentAttributes only *after* tests run, rather
```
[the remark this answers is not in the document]
pond: True. That is a shortcoming of the approach used in 6.1.3 already, of course, since it only includes the ActiveSupport helper inside RailsExampleGroup.

Do you have a "feel" for where else in the hierarchy we might _include this module_ to achieve a wider area of effect? It can be put just about anywhere, especially with some `defined?` and/or `respond_to?` guards, and should be pretty bomb-proof. Once we've figured that out, we can decide if the adapter module's chosen namespace, as a Rails adapter, is correct or if it should also move elsewhere in the code base.

pond:
Note that CI will fail.

Committed changes as given but needed to add a missing `def` and getting a test failure now anyway. The new code is not equivalent to the old. This make take me substantial time to work out as I'm a bit confused by how that method is meant to work, if being called _inside_ an example, so it can't possibly set a before-each or around-all since the example is already running; but then again, it's running that group example separately, so maybe that's now.

JonRowe:
@pirj please don't merge this until I've had a chance to look at it, might be a bit slow because I'm away this week

pond:
@JonRowe Any update?

pond:
@JonRowe / @pirj Just bumping this as it looks like it might have been forgotten.

spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -56,5 +58,83 @@ class CurrentSample < ActiveSupport::CurrentAttributes
...
+                expect(CurrentAttrsBetweenHooks.request_id).to eq('123')
+                CurrentAttrsBetweenHooks.request_id = '234'
+              end
+
+              it 'runs normally' do
+                expect(CurrentAttrsBetweenHooks.request_id).to eq('234')
+              end
+            end
+
+          expect(
```
pirj, line 133: Cosmetic
```suggestion
expect(group.run(failure_reporter)).to be(true), failure_reporter.exceptions
```
The second arg sent to `to` is a custom failure message.

pirj (approved):
I don’t have the bjections to this approach. It may happen that a Current would leak from a Rails e.g to a one that does test rails code but is not marked as such. We could say that our users should mark those groups accordingly, but it’s not a rare situation when people don’t, and this may cause them surprises. Like any shared state, which this Current is.

I wonder if we could instruct the suite to include an config-level around hook for rails example groups. It would take injecting into the users’s rspec conf object. I can’t tell if we already do this, or leave configuration to the user completely.

pond:
Could we at least merge this in the mean time, so I don't have to keep my Gemfile locked to 6.1.2?

its-fern:
+1 to @pond 's comment - we're currently pinned to 6.1.2 due to this issue and would love to upgrade soon!

JonRowe:
Its not forgotten I'd just like for rspec-rails to not be responsible here, I'm thinking about alternative approaches to integrating the minitest hooks that match up with our lifecycles, or preventing around hooks from calling these apis due to this error.

pond:
I note as time goes by that there are now things like conflicts arising, but I'm not minded to spend time "maintaining" this PR unless I know it's of any interest. It sounds like @JonRowe you're thinking of a different approach anyway - perhaps I should just stay  locked to RSpec 6.1.2 until that arises, and maybe just close this PR?

PR #2777, opened 2024-08-08
Extract `cell_selector` method in scaffold index view spec generator

mjankowski:
Basically same change as described in https://github.com/rspec/rspec-rails/pull/2619 - re-attempting only because of last comment - https://github.com/rspec/rspec-rails/pull/2619#issuecomment-1250240003 - which seemed to indicate a desire to get the logic into the generator and out of the generated spec file.

I read the linked discussion and it wasn't totally clear to me which portions involved people talking slightly past each other (about CI vs developer env) and which were genuine issues (that I may have not resolved here, if they exist?) so apologies if that's the case, and I can close and/or attempt to resolve. From my current understanding this does seem safe -- in CI environments or dev environments for working on rspec-rails itself, you'd presumably either have a fresh/clean run and/or clobber first and it should be ok; and for local rails-app development, if you had a prior-version generated spec+view they'd work in conjunction with each other, and this would continue to be true if you switched versions and started generating new ones, I think?

I agree with one of the comments in the linked discussion about how you would likely change this value pretty early on, but also that having the rails version conditional in the generated spec is slightly confusing when you first see it.

### DIFF
--- a/lib/generators/rspec/scaffold/templates/index_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/index_spec.rb
@@ -18,7 +18,7 @@ RSpec.describe "<%= ns_table_name %>/index", <%= type_metatag(:view) %> do
 
   it "renders a list of <%= ns_table_name %>" do
     render
-    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
+    cell_selector = <%= Rails::VERSION::STRING >= '7' ? "'div>p'" : "'tr>td'" %>
 <% for attribute in output_attributes -%>
     assert_select cell_selector, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
 <% end -%>
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,6 +8,7 @@ Enhancements:
 * Verify ActiveJob arguments by comparing to the method signature. (Oli Peate, #2745)
 * Add suggestion to rails_helper.rb to skip when not in test most. (Glauco Custódio, #2751)
 * Add `at_priority` qualifier to `have_enqueued_job` set of matchers. (mbajur, #2759)
+* Remove Rails version-specific conditional from index scaffold generation. (Matt Jankowski, #2777)
 
 Bug Fixes:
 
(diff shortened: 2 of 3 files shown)


pirj (approved):
There’s a lengthy discussion [here](https://github.com/rspec/rspec-rails/pull/2552#discussion_r781619422) about this. I support the change.

lib/generators/rspec/scaffold/templates/index_spec.rb
```diff
@@ -18,9 +18,12 @@
 
   it "renders a list of <%= ns_table_name %>" do
     render
-    cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
 <% for attribute in output_attributes -%>
     assert_select cell_selector, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
```
JonRowe, line 22: Personally, I don't see why this needs to be a method, I would just do:
```suggestion
    assert_select <%= Rails::VERSION::STRING >= '7' ? "'div>p'" : "'tr>td'" %>, text: Regexp.new(<%= value_for(attribute) %>.to_s), count: 2
```

mjankowski: I may have misinterpreted your comment in the linked issue proposing we add a method (I assumed you meant in the generated view spec ... maybe you meant an internal method in the scaffold class that holds this logic and generates the string into the view?)

In either case, there are some other specs which reference `cell_selector` directly ... they keep passing if we turn the local var into a method in the generated view spec, but they'd fail if we change to direct string output here (could be solved by also having this conditional logic inline in the scaffold spec, and/or adding a method like this with duplicated logic to the spec).

I chose the smallest-diff path, but could do another pass to keep it as just a generated string in the view spec output, with conditional logic either in a shared method (template file and scaffold spec) or repeated in those spots.

Relatedly ... another thing which will wind up fixing this issue is whenever Rails 6.1 is EOL, there will presumably be a sweep/cleanup rspec-rails to remove a bunch of this logic, drop old ruby/rails support, etc?

pirj: Do you mind this adjustment? It warrants an insta-merge.
Another alternative that's a minimal diff and an insta-merge would be:
```diff
- cell_selector = Rails::VERSION::STRING >= '7' ? 'div>p' : 'tr>td'
+ cell_selector = <%= Rails::VERSION::STRING >= '7' ? "'div>p'" : "'tr>td'" %>
```

mjankowski: Updated to use that latter approach (leave `cell_selector` as local var) to reduce the number of other specs that need changing.

pirj:
Thank you, @mjankowski !

lib/generators/rspec/scaffold/templates/index_spec.rb
JonRowe: For clarity / future travellers, I originally meant a method in the generator not the spec, but this is fine as is 👍

JonRowe:
Released in 6.1.4

PR #2783, opened 2024-08-26
Allow sqlite3 gem to float to version 2

p8:
Rails main requires `sqlite3` version 2:
https://github.com/rails/rails/pull/51958

PR #2781, opened 2024-08-26
Register stats directories with Rails::CodeStatistics.register_directory

p8:
Rails `main` uses Thor for the `bin/rails stats` command instead of
Rake. This means stats directories need to be added in the Railtie.

The global constant STATS_DIRECTORIES defined by Rails, has been
deprecated in favor of Rails::CodeStatistics.register_directory.

https://github.com/rails/rails/blob/8c7754dfdf39ed94cc93bbc40ee721c311b6d32c/railties/CHANGELOG.md?plain=1#L1-L11

PR #2784, opened 2024-08-27
Fix rails 8 build with 2781 and 2783

JonRowe:
Attempt to see if combining #2781 and #2783 have a green build.


p8:
Looks good to me!

PR #2785, opened 2024-08-27
rails_helper template: Restore old order of support files

franzliedke:
Before #2678, the suggested code snippet for loading support files loaded files in directories (e.g. `spec/support/a/b.rb`) *after* files with the same basename as those directories (e.g. `spec/support/a.rb`).

This tends to be the preferrable load order, given how Ruby modules and classes are usually structured.

The root cause is a difference between the alphabetic ordering of string sorting compared to `Pathname#<=>`.

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -25,7 +25,7 @@ require 'rspec/rails'
 # directory. Alternatively, in the individual `*_spec.rb` files, manually
 # require only the support files necessary.
 #
-# Rails.root.glob('spec/support/**/*.rb').sort.each { |f| require f }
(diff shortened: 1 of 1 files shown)


franzliedke:
@JonRowe suggested this approach in https://github.com/rspec/rspec-rails/pull/2678#issuecomment-2311836926.

franzliedke:
Side note: Is there a template for a default `spec_helper.rb` (for non-Rails projects) somewhere? 🧐

pirj:
> Is there a template for a default spec_helper.rb

[Here you go](https://github.com/rspec/rspec-core/blob/2ba0f10fe68948e5ee9addd569b4694f0245aa71/lib/rspec/core/project_initializer/spec/spec_helper.rb#L1), @franzliedke.

pirj:
Thank you!

JonRowe:
Released in 6.1.5 (and later)

PR #2787, opened 2024-09-01
Update rspec-rails.gemspec with supported Rails description.

simi:


JonRowe:
Thank you for highlighting this! I've merged a similar change removing the version from the description in favour of our readme / the version of the gem required by the gem spec.

PR #2789, opened 2024-09-02
Register Rails::CodeStatistics directory with test_directory arg

p8:
There is no need to edit the TEST_TYPES constant directly.


JonRowe:
Released in 7.0.1

p8:
Thanks!

PR #2790, opened 2024-09-02
Drop support for rails versions < 7 in scaffold generator

mjankowski:
With rspec-rails 7.0.0 release, my assumption is that it's safe/desired to drop support for any rails-conditional stuff earlier than 7.0?

This is a followup on https://github.com/rspec/rspec-rails/pull/2777 which removes the conditional and just uses the style generated by rails 7.0 and higher.

Side note - opened this PR just b/c the prior one was mine and this change made sense to me from previous PR when I saw the 7.0.0 release version support. Would other PRs of this nature (cleaning up conditional support checks) be welcome?


JonRowe (approved):

JonRowe:
Released in 7.0.1

PR #2793, opened 2024-09-04
Update method name in HaveEnqueuedMail (job_match? to job_matches?)

davidrunger:
This change fixes a bug wherein a spec that expects no mail to have been enqueued will fail (falsely claiming that a mail _was_ enqueued) if any _non-mail_ job is enqueued, even if indeed no _mail_ job is actually enqueued.

This bug was introduced in ab6e6e83 / #2780 , which was first included in the 7.0.0 release.

In that change, the `RSpec::Rails::Matchers::ActiveJob::Base` `job_match?` instance method was renamed to `job_matches?`. The problem is that `RSpec::Rails::Matchers::HaveEnqueuedMail`, which inherits from `RSpec::Rails::Matchers::ActiveJob::Base` (via
`RSpec::Rails::Matchers::ActiveJob::HaveEnqueuedJob`), intends to override that method, which it does by defining its own `job_match?` instance method. However, because this `job_match?` method name is no longer the same as the (now so called) `job_matches?` method that it intends to override, the `job_match?` method of
`RSpec::Rails::Matchers::HaveEnqueuedMail` is no longer having the intended override effect on the behavior of the matcher, producing the aforementioned bug.

This change resolves the issue by updating the method name in the `RSpec::Rails::Matchers::HaveEnqueuedMail` subclass from `job_match?` to `job_matches?`, reflecting that renaming that was done in the parent class in ab6e6e83 .

### DIFF
--- a/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
+++ b/spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
@@ -20,6 +20,10 @@ if RSpec::Rails::FeatureCheck.has_active_job?
     def test_email; end
   end
 
+  class NonMailerJob < ActiveJob::Base
+    def perform; end
+  end
+
   if RSpec::Rails::FeatureCheck.has_action_mailer_unified_delivery?
     class UnifiedMailer < ActionMailer::Base
       self.delivery_job = ActionMailer::MailDeliveryJob
@@ -93,6 +97,10 @@ RSpec.describe "HaveEnqueuedMail matchers", skip: !RSpec::Rails::FeatureCheck.ha
 
     it "passes when negated with 0 arguments" do
       expect { }.not_to have_enqueued_email
+    end
+
+    it "passes when negated with 0 arguments and a non-mailer job is enqueued" do
+      expect { NonMailerJob.perform_later }.not_to have_enqueued_email
     end
 
     it "passes when only given mailer argument" do
@@ -305,11 +313,6 @@ RSpec.describe "HaveEnqueuedMail matchers", skip: !RSpec::Rails::FeatureCheck.ha
     end
 
     it "generates a failure message with unmatching enqueued mail jobs" do
-      non_mailer_job = Class.new(ActiveJob::Base) do
-        def perform; end
-        def self.name; "NonMailerJob"; end
-      end
-
       send_time = Date.tomorrow.noon
       queue = 'urgent_mail'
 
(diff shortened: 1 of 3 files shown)


spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -305,11 +313,6 @@ def test_email; end
     end
 
     it "generates a failure message with unmatching enqueued mail jobs" do
-      non_mailer_job = Class.new(ActiveJob::Base) do
-        def perform; end
-        def self.name; "NonMailerJob"; end
-      end
```
davidrunger, line 311: (Since now there are two specs in this file (i.e. the spec being modified here, and the new spec that I added above) that have occasion to refer to a non-mailer job class, I am extracting this previously anonymous/local job class up to an actual class constant defined in the setup at the top of this file. The downside of this is that the global Ruby namespace in specs is now polluted by an additional constant, `NonMailerJob`, but personally I think that's worthwhile, for the upside of any specs that need to reference such a job being able to conveniently do so.)

spec/rspec/rails/matchers/have_enqueued_mail_spec.rb
```diff
@@ -95,6 +99,10 @@ def test_email; end
       expect { }.not_to have_enqueued_email
     end
 
+    it "passes when negated with 0 arguments and a non-mailer job is enqueued" do
+      expect { NonMailerJob.perform_later }.not_to have_enqueued_email
```
davidrunger, line 103: (This new spec fails without the `job_match?` => `job_matches?` rename done in this PR.)

davidrunger:
For an earlier approved build ([here](https://github.com/rspec/rspec-rails/actions/runs/10701028512)) before I force-pushed a change or two, two matrix items failed:

1. Ruby: 3.3, Rails: ~> 7.2.0
2. Ruby: 3.3, Rails: ~> 7.1.0

However, these are also failing on `main` (e.g. as can be seen in what is currently the most recent build for `main`, [here](https://github.com/rspec/rspec-rails/actions/runs/10700191903)), so I don't think that those failures are related to or caused by this PR, and I expect that those will fail again, if another build is approved to run.

JonRowe:
Good catch can you rebase this please

davidrunger:
@JonRowe Thank you for fixing the failing specs on `main`, and thank you for taking a look at this PR.

I just rebased with latest `main` and force-pushed.

JonRowe:
Thanks, merged despite build failure as that has been fixed in #2798

JonRowe:
Released in 7.0.2 and 7.1.0

PR #2794, opened 2024-09-04
Silence warning in mailer preview path check

JonRowe:


davidrunger:
@JonRowe : A possible alternative is here: https://github.com/rspec/rspec-rails/pull/2796 , although I'm not sure whether you have already considered that approach and believe your approach here to be superior, and I'm also not 100% sure that will actually get all of the specs passing (although I think it will).

PR #2795, opened 2024-09-04
Improve implicit description for RSpec::Rails::Matchers::ActionCable::HaveBroadcastedTo

boardfish:
Previously, tests defined like this:

```rb
it do
  expect { some_action }.to have_broadcasted_to("stream_name").from_channel(SomeChannel).with("some_data")
end
```

would implicitly be given the description `have broadcasted to`, which is derived from the name of the matcher method.
This didn't carry any extra information about the data we're expecting to be broadcast from the channel - behaviour
which would be very welcome in combination with Turbo Streams. With a matcher for Turbo Streams, we might benefit from
a description like `"broadcast exactly 1 messages to stream with turbo-stream[action="append"][target="some_list"]"`.

### DIFF
--- a/lib/rspec/rails/matchers/action_cable.rb
+++ b/lib/rspec/rails/matchers/action_cable.rb
@@ -47,7 +47,6 @@ module RSpec
       #     expect {
       #       ActionCable.server.broadcast "messages", text: 'Hi!'
       #     }.to have_broadcasted_to("messages").with(text: 'Hi!')
-
       def have_broadcasted_to(target = nil)
         check_action_cable_adapter
 
--- a/lib/rspec/rails/matchers/action_cable.rb
+++ b/lib/rspec/rails/matchers/action_cable.rb
@@ -3,6 +3,8 @@ require "rspec/rails/matchers/action_cable/have_broadcasted_to"
 module RSpec
   module Rails
     module Matchers
+      extend RSpec::Matchers::DSL
+
       # Namespace for various implementations of ActionCable features
       #
       # @api private
@@ -45,12 +47,16 @@ module RSpec
       #     expect {
       #       ActionCable.server.broadcast "messages", text: 'Hi!'
       #     }.to have_broadcasted_to("messages").with(text: 'Hi!')
+
       def have_broadcasted_to(target = nil)
         check_action_cable_adapter
 
         ActionCable::HaveBroadcastedTo.new(target, channel: described_class)
       end
-      alias_method :broadcast_to, :have_broadcasted_to
+
+      alias_matcher :broadcast_to, :have_broadcasted_to do |desc|
+        desc.gsub("have broadcasted", "broadcast")
+      end
 
     private
 
--- a/lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
+++ b/lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
@@ -51,6 +51,10 @@ module RSpec
             exactly(:thrice)
           end
 
+          def description
+            "have broadcasted #{base_description}"
+          end
+
           def failure_message
             "expected to broadcast #{base_message}".tap do |msg|
               if @unmatching_msgs.any?
@@ -140,18 +144,21 @@ module RSpec
               end
           end
 
-          def base_message
+          def base_description
             "#{message_expectation_modifier} #{@expected_number} messages to #{stream}".tap do |msg|
               msg << " with #{data_description(@data)}" unless @data.nil?
-              msg << ", but broadcast #{@matching_msgs_count}"
             end
+          end
+
+          def base_message
+            "#{base_description}, but broadcast #{@matching_msgs_count}"
           end
 
           def data_description(data)
             if data.is_a?(RSpec::Matchers::Composable)
               data.description
             else
-              data
+              data.inspect
             end
           end
 
--- a/spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
+++ b/spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
@@ -226,5 +226,36 @@ RSpec.describe "have_broadcasted_to matchers", skip: !RSpec::Rails::FeatureCheck
         end
       end
     end
+
+    it "has an appropriate description" do
+      expect(have_broadcasted_to("my_stream").description).to eq("have broadcasted exactly 1 messages to my_stream")
+    end
+
+    it "has an appropriate description when aliased" do
+      expect(broadcast_to("my_stream").description).to eq("broadcast exactly 1 messages to my_stream")
+    end
+
+    it "has an appropriate description when stream name is passed as an array" do
+      expect(have_broadcasted_to(%w[my_stream stream_2]).from_channel(channel).description).to eq("have broadcasted exactly 1 messages to broadcast:my_stream:stream_2")
+    end
+
+    it "has an appropriate description not mentioning the channel when qualified with `#from_channel`" do
+      expect(have_broadcasted_to("my_stream").from_channel(channel).description).to eq("have broadcasted exactly 1 messages to my_stream")
+    end
+
+    it "has an appropriate description including the expected contents when qualified with `#with`" do
+      expect(have_broadcasted_to("my_stream").from_channel(channel).with("hello world").description).to eq("have broadcasted exactly 1 messages to my_stream with \"hello world\"")
+    end
+
+    it { expect("foo").to eq("foo") }
+
+    it "has an appropriate description including the matcher's description when qualified with `#with` and a composable matcher" do
+      expect(
+        have_broadcasted_to("my_stream")
+          .from_channel(channel)
+          .with(a_hash_including(a: :b))
+          .description
+      ).to eq("have broadcasted exactly 1 messages to my_stream with a hash including {:a => :b}")
+    end
   end
 end
--- a/spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
+++ b/spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
@@ -246,9 +246,6 @@ RSpec.describe "have_broadcasted_to matchers", skip: !RSpec::Rails::FeatureCheck
     it "has an appropriate description including the expected contents when qualified with `#with`" do
       expect(have_broadcasted_to("my_stream").from_channel(channel).with("hello world").description).to eq("have broadcasted exactly 1 messages to my_stream with \"hello world\"")
     end
-
-    it { expect("foo").to eq("foo") }
-
     it "has an appropriate description including the matcher's description when qualified with `#with` and a composable matcher" do
       expect(
         have_broadcasted_to("my_stream")


spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
         end
       end
     end
+
+    it "has a description when not chained" do
```
pirj, line 230: “Chained” can mean two things: adding a qualifier like from_message, and compounding with and/or. In this context, when you say that it is not chained, it is unclear what is meant.

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+
+    it "has a description when stream name is passed as an array" do
+      expect(have_broadcasted_to(%w[stream stream_2]).from_channel(channel).description).to eq("have broadcasted exactly 1 messages to broadcast:stream:stream_2")
+    end
+
+    it "has a description when stream name is passed as an array and aliased" do
+      expect(broadcast_to(%w[stream stream_2]).from_channel(channel).description).to eq("broadcast exactly 1 messages to broadcast:stream:stream_2")
+    end
+
+    it "has a description when chained with `from_channel" do
```
pirj, line 246: Misses backtick?

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
         end
       end
     end
+
+    it "has a description when not chained" do
+      expect(have_broadcasted_to("stream").description).to eq("have broadcasted exactly 1 messages to stream")
```
pirj, line 231: “stream” reads too generic, maybe “my_stream” to make it clear that it’s the same as specified?

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+    end
+
+    it "has a description when not chained and aliased" do
+      expect(broadcast_to("stream").description).to eq("broadcast exactly 1 messages to stream")
+    end
+
+    it "has a description when stream name is passed as an array" do
+      expect(have_broadcasted_to(%w[stream stream_2]).from_channel(channel).description).to eq("have broadcasted exactly 1 messages to broadcast:stream:stream_2")
```
pirj, line 239: Does the channel affect the description in some way?

lib/rspec/rails/matchers/action_cable.rb
```diff
@@ -50,7 +50,13 @@ def have_broadcasted_to(target = nil)
...
         ActionCable::HaveBroadcastedTo.new(target, channel: described_class)
       end
-      alias_method :broadcast_to, :have_broadcasted_to
+
+      def broadcast_to(target = nil)
+        RSpec::Matchers::AliasedMatcher.new(
+          have_broadcasted_to(target),
+          ->(old_desc) { old_desc.gsub("have broadcasted", "broadcast") }
+        )
+      end
```
JonRowe, line 59: You can't use that class here, its a private api of rspec-expectations, instead:

```suggestion
      alias_matcher :broadcast_to, :have_broadcasted_to { |desc| desc.gsub("have broadcasted", "broadcast") }
```

You may need to ensure the matcher DSL is included which is fine to do.

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
boardfish: It's required because it can't be implicitly inferred by the matcher. It's not included in the failure message, so I didn't include it in the implicit description either. I think it might be a useful thing to include in both, though?

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
         end
       end
     end
+
+    it "has a description when not chained" do
+      expect(have_broadcasted_to("stream").description).to eq("have broadcasted exactly 1 messages to stream")
+    end
+
+    it "has a description when not chained and aliased" do
```
JonRowe, line 234:
```suggestion
    it "has an appropriate description when aliased" do
```

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
JonRowe, line 230:
```suggestion
    it "has an appropriate description" do
```

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+      expect(broadcast_to("stream").description).to eq("broadcast exactly 1 messages to stream")
+    end
+
+    it "has a description when stream name is passed as an array" do
+      expect(have_broadcasted_to(%w[stream stream_2]).from_channel(channel).description).to eq("have broadcasted exactly 1 messages to broadcast:stream:stream_2")
+    end
+
+    it "has a description when stream name is passed as an array and aliased" do
+      expect(broadcast_to(%w[stream stream_2]).from_channel(channel).description).to eq("broadcast exactly 1 messages to broadcast:stream:stream_2")
+    end
```
JonRowe, line 244: I don't think this test is really necessary given the implementation

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
JonRowe, line 246:
```suggestion
    it "has a description not mentioning the channel when constrained with `from_channel`" do
```

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+
+    it "has a description when chained with `from_channel" do
+      # We don't include the channel in the failure message, so we don't include it in the description either.
```
JonRowe, line 247: Inline into the doc string

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+      expect(have_broadcasted_to("stream").from_channel(channel).description).to eq("have broadcasted exactly 1 messages to stream")
+    end
+
+    it "has a description when chained with `from_channel" do
+      expect(have_broadcasted_to("stream").from_channel(channel).description).to eq("have broadcasted exactly 1 messages to stream")
+    end
```
JonRowe, line 253: Duplicate spec

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+
+    it "has a description when chained with `with`" do
```
JonRowe, line 255:
```suggestion
    it "has a description including the contents when constrained with `with`" do
```

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,43 @@ def broadcast(stream, msg)
...
+      expect(have_broadcasted_to("stream").from_channel(channel).with("hello world").description).to eq("have broadcasted exactly 1 messages to stream with \"hello world\"")
+    end
+
+    it "has a description when chained with `with` and a composable matcher" do
```
JonRowe, line 259:
```suggestion
    it "has a description including the matchers description when constraining with a composable matcher" do
```

JonRowe (changes requested):
Good improvements here, just some minor tweaks to the specs and a removal of a private api in favour of the public one

boardfish:
Thank you both for your speedy reviews! I think I've addressed everything you've mentioned 👍

lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
```diff
@@ -140,9 +144,14 @@ def set_expected_number(relativity, count)
...
+          def base_description
             "#{message_expectation_modifier} #{@expected_number} messages to #{stream}".tap do |msg|
               msg << " with #{data_description(@data)}" unless @data.nil?
+            end
+          end
+
+          def base_message
+            base_description.tap do |msg|
               msg << ", but broadcast #{@matching_msgs_count}"
             end
```
JonRowe, line 156: This minorly improve allocations 
```suggestion
            "#{base_description}, but broadcast #{@matching_msgs_count}"
```

lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
```diff
@@ -151,7 +160,7 @@ def data_description(data)
             if data.is_a?(RSpec::Matchers::Composable)
               data.description
             else
-              data
+              data.inspect
```
JonRowe, line 163: What did this actually come from, I'm on the fence wether letting user values user `to_s` is fine or not.

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,34 @@ def broadcast(stream, msg)
         end
       end
     end
+
+    it "has an appropriate description when not qualified with a chain method" do
```
JonRowe, line 230: Honestly I'd prefer leaving off the `when not ...` of all of these as the complexity would just grow over time.
```suggestion
    it "has an appropriate description" do
```

lib/rspec/rails/matchers/action_cable.rb
```diff
@@ -45,12 +45,17 @@ module ActionCable
       #     expect {
       #       ActionCable.server.broadcast "messages", text: 'Hi!'
       #     }.to have_broadcasted_to("messages").with(text: 'Hi!')
+      extend RSpec::Matchers::DSL
```
JonRowe, line 48: This needs to moved to the top of the module, under the module doc comment, so that yard works properly.

JonRowe (changes requested):

lib/rspec/rails/matchers/action_cable/have_broadcasted_to.rb
boardfish: I would expect a matcher to use this if mentioning an object. The default description for the `eq` matcher does this:

```rb
it { expect("foo").to eq("foo") }
```

Resultant description:

```
is expected to eq "foo"
```

This adds some clarity to what the object is in context.

spec/rspec/rails/matchers/action_cable/have_broadcasted_to_spec.rb
```diff
@@ -226,5 +226,36 @@ def broadcast(stream, msg)
...
+    it "has an appropriate description not mentioning the channel when qualified with `#from_channel`" do
+      expect(have_broadcasted_to("my_stream").from_channel(channel).description).to eq("have broadcasted exactly 1 messages to my_stream")
+    end
+
+    it "has an appropriate description including the expected contents when qualified with `#with`" do
+      expect(have_broadcasted_to("my_stream").from_channel(channel).with("hello world").description).to eq("have broadcasted exactly 1 messages to my_stream with \"hello world\"")
+    end
+
+    it { expect("foo").to eq("foo") }
+
```
JonRowe, line 251:
```suggestion

```

lib/rspec/rails/matchers/action_cable.rb
```diff
@@ -45,12 +47,16 @@ module ActionCable
       #     expect {
       #       ActionCable.server.broadcast "messages", text: 'Hi!'
       #     }.to have_broadcasted_to("messages").with(text: 'Hi!')
+
```
JonRowe, line 50:
```suggestion
```

JonRowe (approved):
LGTM thank you for this

JonRowe:
Released in 7.1.0

PR #2796, opened 2024-09-04
Add 'ostruct' to example app Gemfile

davidrunger:
Currently, specs are failing for the "Ruby: 3.3, Rails: ~> 7.1.0" and "Ruby: 3.3, Rails: ~> 7.2.0" matrix builds. This is caused by Ruby 3.3.5 (released yesterday, 2024-09-03) now warning if `ostruct` is loaded from the standard library, rather than from a gem.

This change fixes the failing specs by adding the `ostruct` gem to the `Gemfile` of the example app, which causes the problematic warning not to be emitted by Ruby.

### DIFF
--- a/example_app_generator/generate_app.rb
+++ b/example_app_generator/generate_app.rb
@@ -25,6 +25,7 @@ in_root do
   gsub_file "Gemfile", /.*puma.*/, ''
   gsub_file "Gemfile", /.*bootsnap.*/, ''
 
+  append_to_file 'Gemfile', "gem 'ostruct'\n"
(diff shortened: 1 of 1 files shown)


davidrunger:
I think that #2794 aims to solve the same problem that this PR does (i.e. currently failing specs for some matrix builds on `main`), so, if that PR is merged, then that will render this PR unnecessary, and it should be closed.

JonRowe:
:wave: Thanks but its actually not just ostruct and is actually Rails issue upstream

PR #2804, opened 2024-09-17
Comment out `infer_spec_type_from_file_location!` in generated helper.

### LINKED ISSUE #2803: Remove `infer_spec_type_from_location!` from generator
Be consistent. Don't make infer spec types the default, and then also generate the types in the spec files. Also don't describe it as opt-in, when it is on by default.

rubocop-rails has [a cop](https://docs.rubocop.org/rubocop-rspec_rails/cops_rspecrails.html#rspecrailsinferredspectype) that helps with this inconsistency and removes the types in the spec files
> After setting up rspec-rails, you will have enabled config.infer_spec_type_from_file_location! by default in spec/rails_helper.rb. This cop works in conjunction with this config. 

Remove `infer_spec_type_from_location!` from https://github.com/rspec/rspec-rails/blob/124cbcb2c689e6e9addaaac482b6ff2ec2342684/lib/generators/rspec/install/templates/spec/rails_helper.rb#L83

It is described as the old behavior.
https://github.com/rspec/rspec-rails/blob/124cbcb2c689e6e9addaaac482b6ff2ec2342684/Changelog.md?plain=1#L779
> The old behaviour

But it is not very old or opt-in, because it is still enabled by default in the generator.
[This PR](https://github.com/rspec/rspec-rails/pull/1002#issue-31913596) describes it as "truly opt-in".
But, for that to be true, `infer_spec_type_from_location!` would have to be commented out or removed in the generator.

Make it configurable whether `type: ` is generated by the spec generators
https://github.com/rspec/rspec-rails/blob/124cbcb2c689e6e9addaaac482b6ff2ec2342684/lib/generators/rspec/model/templates/model_spec.rb#L4

**Additional context**

JonRowe:
We would prefer that people tag their specs manually (or via generators) and whilst our generators do generate tags we never removed the generated `infer_spec_type_from_file_location!`  as a default.

Fixes #2803

### DIFF
--- a/lib/generators/rspec/install/templates/spec/rails_helper.rb
+++ b/lib/generators/rspec/install/templates/spec/rails_helper.rb
@@ -67,20 +67,22 @@ RSpec.configure do |config|
   # config.use_transactional_fixtures = true
 
 <% end -%>
-  # RSpec Rails can automatically mix in different behaviours to your tests
-  # based on their file location, for example enabling you to call `get` and
-  # `post` in specs under `spec/controllers`.
+  # RSpec Rails uses metadata to mix in different behaviours to your tests,
+  # for example enabling you to call `get` and `post` in request specs. e.g.:
   #
-  # You can disable this behaviour by removing the line below, and instead
-  # explicitly tag your specs with their type, e.g.:
-  #
-  #     RSpec.describe UsersController, type: :controller do
+  #     RSpec.describe UsersController, type: :request do
   #       # ...
   #     end
   #
   # The different available types are documented in the features, such as in
   # https://rspec.info/features/7-0/rspec-rails
-  config.infer_spec_type_from_file_location!
+  #
+  # You can also this infer these behaviours automatically by location, e.g.
+  # /spec/models would pull in the same behaviour as `type: :model` but this
+  # behaviour is considered legacy and will be removed in a future version.
+  #
+  # To enable this behaviour uncomment the line below.
+  # config.infer_spec_type_from_file_location!
 
   # Filter lines from Rails gems in backtraces.
   config.filter_rails_from_backtrace!
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,8 @@ Enhancements:
 
 * Improve implicit description for ActionCable matchers `have_broadcasted_to` /
   `have_broadcast`. (Simon Fish, #2795)
+* Comment out `infer_spec_type_from_file_location!` in newly generated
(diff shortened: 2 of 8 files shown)


pirj (approved):
Lgtm

JonRowe:
Released in 7.1.0

JonRowe:
A combination of factors, it is better to be explicit about things to improve the "documentation" effect, it makes it less surprising when you include extra helpers into different types of specs and its less brittle as an implementation.

BenMorganMY:
I think this really goes against the naming conventions that Rails instills.

JonRowe:
> I think this really goes against the naming conventions that Rails instills.

Rails doesn't automatically configure things based on directories either, you inherit from classes e.g. `ActionController::Base`

fonji:
> > I think this really goes against the naming conventions that Rails instills.
> 
> Rails doesn't automatically configure things based on directories either, you inherit from classes e.g. `ActionController::Base`

I think it goes against [the convention over configuration part of the doctrine](https://rubyonrails.org/doctrine#convention-over-configuration). Maybe that's what @BenMorganMY wanted to talk about.

BenMorganMY:
@fonji yep, this is primarily about ensuring convention over configuration. It's redundant for a Rails application to also have to specify the spec type and will end up encouraging bad configuration behaviour. This should not be the default in a Rails application.

lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -67,20 +67,22 @@
...
   #       # ...
   #     end
   #
   # The different available types are documented in the features, such as in
   # https://rspec.info/features/7-0/rspec-rails
-  config.infer_spec_type_from_file_location!
+  #
+  # You can also this infer these behaviours automatically by location, e.g.
+  # /spec/models would pull in the same behaviour as `type: :model` but this
+  # behaviour is considered legacy and will be removed in a future version.
```
schlick, line 82: Why not keep it as a configurable option?

JonRowe: This is a template, to configure it you just remove the comment part of  this line in your project.

schlick: My concern was that this comment meant that `config.infer_spec_type_from_file_location!` will be removed in a future version

JonRowe: There is no timeline for its removal, if we did we'd probably just pull it into its own gem

PR #2808, opened 2024-10-25
Respect verify partial doubles config when verifying job/mailer arguments

### LINKED ISSUE #2801: Add configuration to disable ActiveJob job signature matching check

The 7.0 version of RSpec Rails introduced https://github.com/rspec/rspec-rails/pull/2745 by @odlp which is overall a great addition, but doesn't provide a configuration to disable this check.

This is a problem for one of our app that uses `before_enqueue` and `around_perform` callbacks that modifies the arguments passed to the job. 

I would like a way to disable this check.

I'm not sure there are other solutions short of updating all the jobs in our app.

**Additional context**

Here are the callbacks being used:

```ruby

class ApplicationJob < ActiveJob::Base
  # Automatically retry jobs that encountered a deadlock
  # retry_on ActiveRecord::Deadlocked

  # Most jobs are safe to ignore if the underlying records are no longer available
  discard_on ActiveJob::DeserializationError

  before_enqueue do |job|
    if Actor.current?
      job.arguments << { actor: Actor.current }
    end
  end

  around_perform do |job, block|
    arg = job.arguments.find { |a| a.is_a?(Hash) && a.key?(:actor) }
    if arg
      job.arguments.delete(arg)
      actor = arg[:actor]
    end

    if actor
      Actor.with(actor) do
        block.call
      end
    else
      block.call
    end
  end
end
```

odlp:
Following on from #2745, this PR skips the job/mailer verification that arguments match the signature when:

- RSpec Mocks is configured with `verify_partial_doubles` false
- The examples/expectations are wrapped with [`#without_partial_double_verification`](https://rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:without_partial_double_verification)

This direction was suggested by @JonRowe [here](https://github.com/rspec/rspec-rails/issues/2801#issuecomment-2387109797):

> I'm leaving this open because I think this should respect the existing configuration for verifying doubles, we should check to see if we should verify arguments.

Closes #2801.


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -189,6 +191,13 @@ def detect_args_signature_mismatch(jobs)
             nil
           end
 
+          def skip_signature_verification?
+            return true unless defined?(RSpec::Mocks)
```
odlp, line 195: Trying to honour that these matchers might be used without RSpec Mocks, as indicated by this comment here:

https://github.com/rspec/rspec-rails/blob/1ba134da2c8a48c3e4a26d6bb9104ef33d91781f/lib/rspec/rails/matchers/have_enqueued_mail.rb#L1-L3

But I'm not sure if this guard is strictly necessary...

JonRowe: You can remove this check, that is about avoiding conflict with other mocking frameworks, you can safely load configuration for the later check.

odlp: Thanks for clarifying. I've removed this guard.

I tried requiring `rspec/mocks/configuration` but it doesn't seem possible to require in isolation - we then get errors regarding uninitialized constants for (Mocks) Syntax, ExampleMethods etc. Instead of requiring all of `rspec/mocks` I've left this without a require for now... happy to adjust as preferred.

JonRowe (changes requested):
Rather than creating a new helper, I would rather the specs be setup with an around block that restores the original behaviour, and explicitly setting the values to `true`/`false` in a `before` in the `context`, I don't mind if you make a helper for the "restore" part (in an around hook) but I would strongly prefer it if the contexts set the correct values directly in a before.

JonRowe:
> FWIW I find a block-based helper superior when temporarily mutating global state because that interface guarantees the state is reset (like [Timecop safe mode](https://github.com/travisjeffery/timecop?tab=readme-ov-file#timecopsafe_mode) for example).

> I've worked in many codebases which mutate global state in before hooks and don't have a corresponding after or around hook to reset said state, leading to order-dependent test failures. Appreciate this codebase is safeguarded with good code review with high standards... but to err is human and so on...

An around hook is essentially a block based helper, the problem I had with it is the implementation obscured the actual config setting that was being set, as our specs act as part of our documentation its less obvious whats being tweaked, an alternative would be to use the `without_partial_double_verification` block helper thats actually part of `rspec-mocks` if you prefer.

odlp:
@JonRowe 

> An around hook is essentially a block based helper

Very true.

> an alternative would be to use the [without_partial_double_verification](https://rspec.info/documentation/3.13/rspec-mocks/RSpec/Mocks/ExampleMethods.html#without_partial_double_verification-instance_method) block helper

I do [use](https://github.com/rspec/rspec-rails/pull/2808/commits/67a538fd307c4243642226b6792cf769f96a4fd0#diff-1c76f986480ad13c755c6a7241d8ab75c01dad37f20afde838ebb36ec0680212) this in the other specs - but it sets a distinct flag `Mocks.configuration.temporarily_suppress_partial_double_verification` rather than `Mocks.configuration.verify_partial_doubles` (it's a pity in a way there are two flags at play here).

Overall I think the PR is fine now; I'm happy to adopt the house style and appreciate the trade-offs you mention between obscuring the config setting vs safety.

JonRowe:
Released in 7.1.0

Issue #2810, opened 2024-11-08
Lacks a corresponding generator to the Rails 8 authentication system generator


tnagatomi:
## What Ruby, Rails and RSpec versions are you using?

Ruby version: 3.3.6
Rails version: 8.0.0
RSpec version: rspec-rails 7.0.1

## Observed behaviour

rspec-rails lacks a corresponding generator to the authentication system generator added by Rails 8.0.0 which can be run by `bin/rails generate authentication`, and produce an error.

## Expected behaviour

rspec-rails generates boilerplate specs corresponding to the authentication system generator when you run `bin/rails generate authentication`.

## Can you provide an example reproduction?

1. Initialize a fresh rails app with `rails new --skip-test sample-app`
2. Add `rspec-rails` to the Gemfile and run `bundle install`
3. Run `bin/rails generate authentication`
4. Command ends with `error  rspec [not found]` like below:

```console
❯ bin/rails generate authentication
      invoke  erb
      create    app/views/passwords/new.html.erb
      create    app/views/passwords/edit.html.erb
      create    app/views/sessions/new.html.erb
      create  app/models/session.rb
      create  app/models/user.rb
      create  app/models/current.rb
      create  app/controllers/sessions_controller.rb
      create  app/controllers/concerns/authentication.rb
      create  app/controllers/passwords_controller.rb
      create  app/channels/application_cable/connection.rb
      create  app/mailers/passwords_mailer.rb
      create  app/views/passwords_mailer/reset.html.erb
      create  app/views/passwords_mailer/reset.text.erb
      create  test/mailers/previews/passwords_mailer_preview.rb
      insert  app/controllers/application_controller.rb
       route  resources :passwords, param: :token
       route  resource :session
        gsub  Gemfile
      bundle  install --quiet
    generate  migration CreateUsers email_address:string!:uniq password_digest:string! --force
       rails  generate migration CreateUsers email_address:string!:uniq password_digest:string! --force
      invoke  active_record
      create    db/migrate/20241108074415_create_users.rb
    generate  migration CreateSessions user:references ip_address:string user_agent:string --force
       rails  generate migration CreateSessions user:references ip_address:string user_agent:string --force
      invoke  active_record
      create    db/migrate/20241108074416_create_sessions.rb
       error  rspec [not found]
```

JonRowe:
PRs welcome 😂

jeromedalbert:
I took a stab at it in https://github.com/rspec/rspec-rails/pull/2811.

PR #2811, opened 2024-11-11
Add Rails 8 authentication generator

jeromedalbert:
Fixes https://github.com/rspec/rspec-rails/issues/2810.

## Problem

When running Rails 8's `bin/rails generate authentication`, generation ends with with the following error: `error  rspec [not found]`.

This is because the Rails `AuthenticationGenerator` [contains](https://github.com/rails/rails/pull/53165) [`hook_for :test_framework`](https://github.com/rails/rails/blob/d5be42897e09850ffbd31394ea9f805b9af055d5/railties/lib/rails/generators/rails/authentication/authentication_generator.rb#L55), which calls `rspec-rails`, but it errors out since `rspec-rails` does not have an authentication generator.

## Solution

Add authentication generator that creates a user test and possibly a fixture, similarly to Rails' [`TestUnit::Generators::AuthenticationGenerator`](https://github.com/rails/rails/blob/d5be42897e09850ffbd31394ea9f805b9af055d5/railties/lib/rails/generators/test_unit/authentication/authentication_generator.rb). The fixture is basically the same as the [Rails one](https://github.com/rails/rails/blob/d5be42897e09850ffbd31394ea9f805b9af055d5/railties/lib/rails/generators/test_unit/authentication/templates/test/fixtures/users.yml.tt) except that it has a comment similar to `rspec-rails`' model [fixture](https://github.com/rspec/rspec-rails/blob/01704c50c146f720db914724c25681781ecefb23/lib/generators/rspec/model/templates/fixtures.yml).

An alternative solution could be to somehow disable the authentication generator for `rspec-rails` so nothing is generated while not having any errors. This might be a viable option if Rails adds more generators in the future and playing catch-up is undesired.

### DIFF
--- a/lib/generators/rspec/authentication/authentication_generator.rb
+++ b/lib/generators/rspec/authentication/authentication_generator.rb
@@ -1,0 +1,25 @@
+require 'generators/rspec'
+
+module Rspec
+  module Generators
+    # @private
+    class AuthenticationGenerator < Base
+      def initialize(args, *options)
+        args.replace(['User'])
+        super
+      end
+
+      def create_user_spec
+        template 'user_spec.rb', target_path('models', 'user_spec.rb')
+      end
+
+      hook_for :fixture_replacement
+
+      def create_fixture_file
+        return if options[:fixture_replacement]
+
+        template 'users.yml', target_path('fixtures', 'users.yml')
+      end
+    end
+  end
+end
--- a/lib/generators/rspec/authentication/templates/user_spec.rb
+++ b/lib/generators/rspec/authentication/templates/user_spec.rb
@@ -1,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe User, <%= type_metatag(:model) %> do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
--- a/lib/generators/rspec/authentication/templates/users.yml
+++ b/lib/generators/rspec/authentication/templates/users.yml
@@ -1,0 +1,11 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
+
+<%% password_digest = BCrypt::Password.create("password") %>
+
+one:
+  email_address: one@example.com
+  password_digest: <%%= password_digest %>
+
+two:
+  email_address: two@example.com
+  password_digest: <%%= password_digest %>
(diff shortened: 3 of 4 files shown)


lib/generators/rspec/authentication/authentication_generator.rb
```diff
@@ -0,0 +1,25 @@
...
+      def initialize(args, *options)
+        args.replace(['User'])
+        super
+      end
+
+      def create_user_spec
+        template 'user_spec.rb', target_path('models', 'user_spec.rb')
+      end
+
+      hook_for :fixture_replacement
```
jeromedalbert, line 16: I added this to mimic the model generator, but maybe this not desired as that means extra work for fixture replacement gems like factory bot. I made https://github.com/thoughtbot/factory_bot_rails/pull/519.

jeromedalbert:
I am not able to run specs on my local because I get the following error on `bundle install` (unrelated to this PR):

```
Fetching https://github.com/rspec/rspec-support.git
Fetching https://github.com/rspec/rspec-mocks.git
Fetching https://github.com/rspec/rspec-expectations.git
Fetching gem metadata from https://rubygems.org/..........
Resolving dependencies...
Could not find compatible versions

Because every version of rspec-expectations depends on rspec-support = 3.14.0.pre
  and every version of rspec-core depends on rspec-support = 3.8.pre,
  every version of rspec-expectations is incompatible with rspec-core >= 0.
So, because Gemfile depends on rspec-core >= 0
  and Gemfile depends on rspec-expectations >= 0,
  version solving has failed.
```

I just pushed a tentative fix that probably won't work, but at this rate it will take a lot of time to fix tests if someone needs to approve every time I push. I am open to suggestions.

jeromedalbert:
Fixed my unit test; I believe it should work this time.

jeromedalbert:
There is an unrelated Rails 7.2.0 spec failure about `scaffold_generator_spec.rb:69`, and the same test is currently failing on the `main` branch. 

So it seems that this PR is good to go, tests-wise.

JonRowe:
I think this is a good start, I'm sort of wondering if it makes sense to add some scaffold specs showing that login works but I think I'm right in thinking the generator as is doesn't generate any tests at all?

jeromedalbert:
> I think I'm right in thinking the generator as is doesn't generate any tests at all?

Correct, I was a bit surprised by this, but the Rails 8 generator doesn't generate any controller tests or integration tests or anything. The only test it generates is an empty `test/models/user_test.rb` `ActiveSupport::TestCase`, and a fixture.

I think that if Rails generates more tests in a future version, such as login tests, then we should do the same in `rspec-rails`.

JonRowe:
Thanks!

Issue #2817, opened 2024-11-30
SQL queries matcher support


hugopeixoto:
[Rails 7 added support for a few query assertions](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/testing/query_assertions.rb):

```ruby
assert_queries_count(1) { Post.first }
assert_queries_count { Post.first }
assert_queries_count(1, include_schema: true) { Post.columns }

assert_no_queries { post.comments }
assert_no_queries(include_schema: true) { Post.columns }

assert_queries_match(/LIMIT \?/, count: 1) { Post.first }
assert_queries_match(/LIMIT \?/) { Post.first }
assert_queries_match(/FROM pg_attribute/i, include_schema: true) { Post.columns }

assert_no_queries_match(/SELECT/i) { post.comments }
assert_no_queries_match(/FROM pg_attribute/i, include_schema: true) { Post.columns }
```

Does an equivalent matcher belong in `rspec-rails`?

I've found these rails assertions useful to ensure performance issues (like N+1 queries) don't appear and to make regression tests for when they do.

In projects that use rspec-rails, a matcher that allows to test for the number of queries executed would be helpful.

I have implemented a custom matcher that supports a subset of these:

```ruby
expect { Post.first }.to execute_queries(1)
expect { Post.first }.to execute_no_queries
```

These could be expanded to support matching, no count, and `include_schema`:

```ruby
expect { Post.first }.to execute_queries
expect { Post.first }.to execute_queries(1).matching(/LIMIT \?/)
expect { Post.first }.to execute_no_queries.including_schema
```

I'm working on making a PR to rspec-rails to support these, opening an issue now to iterate on the syntax and to check if it belongs here.

hugopeixoto:
PR opened at #2818

JonRowe:
Thanks I've seen, will try to review tonight

PR #2818, opened 2024-12-01
Add execute_queries matcher

hugopeixoto:
This matcher is analog to the query assertions available since rails 7.0.

#2817


hugopeixoto:
There are a bunch of checks failing, but they don't seem related to this PR. For example:

```
+Finished in 0.00374 seconds (files took 2.32 seconds to load)
+1 example, 1 failure
+
+Failed examples:
+
+rspec ./spec/lib/fixture_set_blob.rb:4 # blob creates a blob from a sample file
```

hugopeixoto:
Maybe I could do `match_unless_raises` and delegate it to the query assertions instead of reimplementing it. This would be better because I wouldn't be duplicating the `SQLCounter` query.

Also, I'm missing docs. The matcher should at least be added to the readme, and I need to figure out how those work in rspec.

lib/rspec/rails/matchers/execute_queries.rb
```diff
@@ -0,0 +1,152 @@
...
+      # Use the `matching` method to specify a regular expression to filter the
+      # queries.
+      #
+      # Use the `including_schema` method to include schema related queries.
+      #
+      # @example
+      #     expect { Post.first }.to execute_no_queries
+      #     expect { Post.first }.to execute_no_queries.matching(/SELECT/)
+      #     expect { Post.columns }.to execute_no_queries.including_schema
+      def execute_no_queries
```
pirj, line 116: I would be leaning towards ` expect { }.not_to execute_queries`, but an obvious use for this is compound expectations (and/or).

pirj:
> added to the readme

Check the /features directory for examples.


> delegate it to the query assertions instead of reimplementing it

That’s how we do it for most other matchers.

Issue #2820, opened 2024-12-16
Ruby 3.4 support?


voxik:
Trying to run rspec-rails test suite against Ruby 3.4, I encounter test failures such as:

~~~
   8) be_a_new matcher #with right class and new record with composable matchers one attribute is a composable matcher fails
     Failure/Error:
       expect {
         expect(record).to be_a_new(record.class).with(
           foo: a_string_matching("bar"))
       }.to raise_error("attribute {\"foo\"=>(a string matching \"bar\")} was not set on #{record.inspect}")

       expected Exception with "attribute {\"foo\"=>(a string matching \"bar\")} was not set on #<#<Class:0x00007fe77b608208>:0x00007fe77b756c68 @attributes={foo: \"foo\", bar: \"bar\"}>", got #<RSpec::Expectations::ExpectationNotMetError: attribute {"foo" => (a string matching "bar")} was not set on #<#<Class:0x00007fe77b608208>:0x00007fe77b756c68 @attributes={foo: "foo", bar: "bar"}>> with backtrace:
         # ./spec/rspec/rails/matchers/be_a_new_spec.rb:75:in 'block (7 levels) in <top (required)>'
         # ./spec/rspec/rails/matchers/be_a_new_spec.rb:77:in 'block (6 levels) in <top (required)>'
         # ./spec/spec_helper.rb:83:in 'block (3 levels) in <top (required)>'
         # ./spec/spec_helper.rb:77:in 'block (2 levels) in <top (required)>'
     # ./spec/rspec/rails/matchers/be_a_new_spec.rb:77:in 'block (6 levels) in <top (required)>'
     # ./spec/spec_helper.rb:83:in 'block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:77:in 'block (2 levels) in <top (required)>'
~~~

The error message seems to be originating from here:

https://github.com/rspec/rspec-rails/blob/609bb83a76c11c67076ff2619fd5f47edd6325ce/lib/rspec/rails/matchers/be_a_new.rb#L29-L43

Specifically the `#{describe_unmatched_attributes.inspect}`. Here are some details from `binding.irb`:

~~~
rom: /builddir/build/BUILD/rubygem-rspec-rails-7.1.0-build/rspec-rails-7.1.0/usr/share/gems/gems/rspec-rails-7.1.0/lib/rspec/rails/matchers/be_a_new.rb @ line 36 :

    31:             unless actual.is_a?(expected) && actual.new_record?
[...]

PR #2821, opened 2024-12-16
[WIP] Enable CI for Ruby head

voxik:
This is just to demonstrate the Ruby 3.4 errors discussed in #2820

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,9 @@ jobs:
              RAILS_VERSION: 'main'
 
          # Rails 8.0 builds >= 3.2
+         - ruby: 3.4.0-rc1
+           env:
+             RAILS_VERSION: '~> 8.0.0'
          - ruby: 3.3
            env:
              RAILS_VERSION: '~> 8.0.0'
--- a/spec/rspec/rails/matchers/be_a_new_spec.rb
+++ b/spec/rspec/rails/matchers/be_a_new_spec.rb
@@ -71,10 +71,16 @@ RSpec.describe "be_a_new matcher" do
           end
 
           it "fails" do
+            message =
+              if RUBY_VERSION >= '3.4'
+                "attribute {\"foo\" => (a string matching \"bar\")} was not set on #{record.inspect}"
+              else
+                "attribute {\"foo\"=>(a string matching \"bar\")} was not set on #{record.inspect}"
+              end
             expect {
               expect(record).to be_a_new(record.class).with(
                 foo: a_string_matching("bar"))
-            }.to raise_error("attribute {\"foo\"=>(a string matching \"bar\")} was not set on #{record.inspect}")
+            }.to raise_error(message)
           end
 
           context "matcher is wrong type" do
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -45,6 +45,9 @@ jobs:
              RAILS_VERSION: 'main'
 
          # Rails 8.0 builds >= 3.2
+         - ruby: head
```
JonRowe, line 48: I'd prefer if you used one of the preview versions

voxik:
~~~
  1) have_broadcasted_to matchers have_broadcasted_to has an appropriate description including the matcher's description when qualified with `#with` and a composable matcher
     Failure/Error:
       expect(
         have_broadcasted_to("my_stream")
           .from_channel(channel)
           .with(a_hash_including(a: :b))
           .description
       ).to eq("have broadcasted exactly 1 messages to my_stream with a hash including #{{:a => :b}}")

       expected: "have broadcasted exactly 1 messages to my_stream with a hash including {:a=>:b}"
            got: "have broadcasted exactly 1 messages to my_stream with a hash including {:a => :b}"
~~~

Eh, ok. Failure due to:

https://github.com/rspec/rspec-rails/blob/609bb83a76c11c67076ff2619fd5f47edd6325ce/lib/rspec/rails/matchers/base_matcher.rb#L138

voxik:
Yeah, now sure about Rubocop. Maybe using the `"#{{}}"` syntax is not good idea and everything should be conditionalized?  But something along these lines was kind of suggested [here](https://bugs.ruby-lang.org/issues/20857#note-14)

spec/rspec/rails/matchers/be_a_new_spec.rb
```diff
@@ -178,7 +200,7 @@ def new_record?; false; end
             expect(record).to be_a_new(String).with(foo: 'bar')
           }.to raise_error(
             "expected #{record.inspect} to be a new String and " +
-            %(attribute {"foo"=>"bar"} was not set on #{record.inspect})
+            %(attribute #{{"foo"=>"bar"}} was not set on #{record.inspect})
```
JonRowe, line 203: I would extract this to a variable which is defined based on a condition and pass that to `raise_error`

voxik: Could you please elaborate a bit more about the reasons? As I see it, the main advantage of this approach is that the string is still very much reminds the error output. OTOH, if this is extracted into variable, it might have the benefit that Rubocop is happier. Besides that, I'd need to come up with some reasonable variable name, which is always the worst task 🙈

And also is this remark specific to this hunk or in general?

JonRowe: Its in general for these interpolated hunks after your rubocop comment e.g.

```
message =
  if RUBY_VERSION >= '3.4'
    %(attribute {foo: "bar"} was not set on #{record.inspect})
  else
    %(attribute {"foo"=>"bar"} was not set on #{record.inspect})
  end

...

}.to raise_error(message)
```

JonRowe:
If you rebase this the main branch fixed the rubygems errors

voxik:
Rubocop is happy now 👍

But I am not exactly sure about the remaining failures for Ruby 3.4 / Rails 8.0 🤷

benoittgt:
I've probably fixed the two issues in https://github.com/rspec/rspec-rails/pull/2825

JonRowe:
Thanks @voxik

PR #2823, opened 2024-12-17
Ensure mocks config available

### LINKED ISSUE #2822: Failures when using an alternate mock framework

Ruby version: 3.3.3
Rails version: 7.1.4.2
RSpec version: 3.13

## Observed behavior

I use Mocha for mocks with RSpec. In my spec_helper.rb file, I have
```ruby
RSpec.configure do |config|
  config.mock_with :mocha
end
```
Using rspec-rails 7.1.0, whenever I run a spec that uses `have_enqueued_job`, I get the following error:
```
     NoMethodError:
       undefined method `configuration' for module RSpec::Mocks
```
This was introduced in https://github.com/rspec/rspec-rails/commit/2f519d426243f0d2d52bf8df5c3cd18bdb65329e. The issue is not present in rspec-rails 7.0.2.

JonRowe:
Should fix #2822

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -192,6 +192,8 @@ module RSpec
           end
 
           def skip_signature_verification?
+            return true unless defined?(::RSpec::Mocks) && (::RSpec::Mocks.respond_to?(:configuration))
+
             !RSpec::Mocks.configuration.verify_partial_doubles? ||
               RSpec::Mocks.configuration.temporarily_suppress_partial_double_verification
           end
--- a/spec/rspec/rails/matchers/active_job_spec.rb
+++ b/spec/rspec/rails/matchers/active_job_spec.rb
@@ -415,6 +415,18 @@ RSpec.describe "ActiveJob matchers", skip: !RSpec::Rails::FeatureCheck.has_activ
           }
         end
       end
+
+      context "without rspec-mocks loaded" do
+        before do
+          # Its hard for us to unload this, but its fairly safe to assume that we can run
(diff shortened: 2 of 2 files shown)


lib/rspec/rails/matchers/active_job.rb
```diff
@@ -1,5 +1,6 @@
 require "active_job/base"
 require "active_job/arguments"
```
pirj, line 2: We check if mocks are defined [here](https://github.com/rspec/rspec-rails/blob/609bb83a76c11c67076ff2619fd5f47edd6325ce/lib/rspec/rails/active_record.rb#L10). Should we do the same here instead of requiring what’s not used?

JonRowe: The difference is here we are actually using it

pirj: We’re only using it if verified partial doubles are checked, right? Which is not the case when RSpec is configured to mock with an alternative framework. Or am I missing something? It looks like before #2808 everything worked just fine, but apparently rspec-mocks were not required and RSpec::Mocks was not defined, and no problem.

jeffse:
@JonRowe Yes, this works for me!

lib/rspec/rails/matchers/active_job.rb
JonRowe: We don't want to skip the check just because someone is using another mocking framework

pirj: If someone is using another framework, the result of the check will always be false. So it’s not that we skip it, we shortcut it.

PR #2828, opened 2025-01-10
Update instructions for gems installations from git repository

rubysamurai:
Use RSpec monorepo for all related RSpec libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,11 @@ See the [`7-1-maintenance` branch on Github](https://github.com/rspec/rspec-rail
    # (requires main-branch versions of all related RSpec libraries)
    group :development, :test do
      %w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
-       gem lib, git: "https://github.com/rspec/#{lib}.git", branch: 'main'
+       if lib == 'rspec-rails'
+         gem lib, git: "https://github.com/rspec/rspec-rails"
+       else
+         gem lib, git: "https://github.com/rspec/rspec", glob: "#{lib}/#{lib}.gemspec"
+       end
      end
    end
    ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,7 +48,11 @@ See the [`7-1-maintenance` branch on Github](https://github.com/rspec/rspec-rail
    # (requires main-branch versions of all related RSpec libraries)
    group :development, :test do
      %w[rspec-core rspec-expectations rspec-mocks rspec-rails rspec-support].each do |lib|
-       gem lib, git: "https://github.com/rspec/#{lib}.git", branch: 'main'
+       if lib == 'rspec-rails'
+         gem lib, git: "https://github.com/rspec/rspec-rails"
+       else
+         gem lib, git: "https://github.com/rspec/rspec", glob: "#{lib}/#{lib}.gemspec"
```
pirj, line 54: Above it says that it requires "main" branch, but I believe this is the case with this notation as well.

It also corresponds to how we do it here ourselves https://github.com/rspec/rspec-rails/blob/607293139d952e0344564afeaabfb2789701492c/Gemfile-rspec-dependencies#L7

pirj (approved):
Thanks!

README.md
JonRowe: To be honest this is out of date, you don't need to run against the main branch of the other libraries to run rspec-rails from git.

pirj: And we don't have to run our rspec-rails CI against them, too, right?

pirj: We do [need git main on CI](https://github.com/rspec/rspec-rails/blob/607293139d952e0344564afeaabfb2789701492c/rspec-rails.gemspec#L49), but only on CI.

@rubysamurai Would you be up to simplify this section of the README?

rubysamurai: @pirj I just did. Thanks for the review!

JonRowe:
Thanks, merged despite build failure because I fixed that elsewhere and this is only readme

PR #2847, opened 2025-05-16
Fix comment by removing extra word.

ballou88:
Was reading through the helper and was initially confused when reading this sentence, I believe that extra `this` should be removed.


JonRowe:
Yep, this appears to be a typo, thanks!

PR #2849, opened 2025-06-04
Introduce have_reported_error matcher

### LINKED ISSUE #2827: Support for Rails 7.1 error reporter
Rails 7.1 introduced the Rails error reporter: https://guides.rubyonrails.org/error_reporting.html
When using the Rails error reporter rescued errors are quietly handled. Unless I'm overlooking something I don't see support as of it yet.

Would like to be able to test if these errors are reported.

Rails minitest has support here: https://edgeapi.rubyonrails.org/classes/ActiveSupport/Testing/ErrorReporterAssertions.html#method-i-assert_error_reported

Test side effects like no errors being raised and testing third party error reporters directly.

**Additional context**

skatkov:
fixes #2827

rspec-rails is missing support for Rails ErrorReporter, this was introduced to rails in v7 and has been evolving ever since. With my client, we have moved to using ErrorReporter as a unified error reporting interface, so we can easily move from one error tracking software to another with minimal code changes. And we had a need to test this interface with rspec, so we implemented our own matcher to handle this.

I'm suggesting our internal implementation as is. This will probably is not suitable as is for this gem, but I'm open to further improving this implementation.

## Example usage

```
@example Checking for any error
  expect { Rails.error.report(StandardError.new) }.to have_reported_error

@example Checking for specific error class
  expect { Rails.error.report(MyError.new) }.to have_reported_error(MyError)

@example Checking for specific error instance with message
  expect { Rails.error.report(MyError.new("message")) }.to have_reported_error(MyError.new("message"))

@example Checking error attributes
  expect { Rails.error.report(StandardError.new, context: "test") }.to have_reported_error.with(context: "test")

@example Checking error message patterns
  expect { Rails.error.report(StandardError.new("test message")) }.to have_reported_error(/test/)

@example Negation
  expect { "safe code" }.not_to have_reported_error
```


spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      subscribers.delete(subscriber)
+    end
+
+    allow(mock_error_reporter).to receive(:report) do |error, **attrs|
+      subscribers.each { |subscriber| subscriber.report(error, **attrs) }
+    end
+  end
+
+  after do
+    subscribers.clear
```
pirj: This is unlikely to be needed, as ‘let’ will initialize a bew array each time, won’t it?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
+require "rspec/rails/matchers/have_reported_error"
+
+RSpec.describe "have_reported_error matcher" do
+  class TestError < StandardError; end
+  class AnotherTestError < StandardError; end
+
+  let(:mock_error_reporter) { double("ErrorReporter") }
+  let(:subscribers) { [] }
+
+  before do
```
pirj: Can we do without this? Why is this mocking needed?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+    subscribers.clear
+  end
+
+  describe "basic functionality" do
+    it "passes when an error is reported" do
+      test_block = proc do
+        Rails.error.report(StandardError.new("test error"))
+      end
+
+      expect(test_block).to have_reported_error
```
pirj: This suspiciously looks like passing an argument, and like the “value expectations” syntax.

Do you think a proper “this matcher doesn’t work with value expectations “ error will be shown to users on an attempt to use it mistakenly?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+    end
+
+    it "fails when no error is reported" do
+      test_block = proc { "no error" }
+      matcher = have_reported_error
+
+      expect(matcher.matches?(test_block)).to be false
```
pirj: Is there a reason not to use the purposed matcher’s usage form like:
```ruby
expect {
  expect { “no error reported” }.to have_reported_error
}.to fail_with(/failure message/)
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      end
+
+      expect(mock_error_reporter).to receive(:unsubscribe)
+
+      expect {
+        have_reported_error.matches?(test_block)
+      }.to raise_error("unexpected error")
+    end
+  end
+
```
pirj, line 235: This might be tested directly, by adding a spec with expect { } with this matcher.

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      expect(test_block).to have_reported_error(StandardError)
+    end
+
+    it "works with matcher chaining" do
+      test_block = proc do
+        Rails.error.report(TestError.new("test"), user_id: 123)
+      end
+
+      expect(test_block).to have_reported_error(TestError).and have_reported_error
```
pirj: Does it really pass when two errors were reported and we expect to have those two reported? Does the matcher fail if either of the errors was not reported?

features/matchers/README.md
```diff
@@ -24,3 +24,22 @@ expect(response).to render_template(template_name)
 # and it is not persisted
 expect(assigns(:widget)).to be_a_new(Widget)
 ```
+
+### error reporting
+
+```ruby
+# passes if Rails.error.report was called with any error
+expect { Rails.error.report(StandardError.new) }.to have_reported_error
```
pirj, line 32: I think this file is supposed to contain just the outline of the matchers defined here (and at a glance it’s rather incomplete). What do you think of keeping just one example of usage of this matcher? The rest will be taken into the same documentation from the feature file.

pirj, line 32: Is it a realistic example that a literal ‘Rails.error.report’ will be called? Or it’s rather some application logic that will bubble up the exception to eg a controller filter that would do that?

Would it be reasonable to give such a real-world-like example here, eg (borrowed from Rails guides) have_reported_error { process_service } ?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,193 @@
...
+      #   expect { Rails.error.report(StandardError.new, context: "test") }.to have_reported_error.with(context: "test")
+      #
+      # @example Checking error message patterns
+      #   expect { Rails.error.report(StandardError.new("test message")) }.to have_reported_error(/test/)
+      #
+      # @example Negation
+      #   expect { "safe code" }.not_to have_reported_error
+      #
+      # @param expected_error [Class, Exception, Regexp, Symbol, nil] the expected error to match
+      def have_reported_error(expected_error = nil)
```
pirj: We usually provide an alias to accommodate both styles, like ‘send_email’ and ‘have_sent_email’. Would you be opposed to have ‘reports_error’ as an alias?

pirj (commented):
Thank you!

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,237 @@
+require "rspec/rails/matchers/have_reported_error"
```
JonRowe: You shouldn't need to require parts of `rspec/rails` manually?

spec/rspec/rails/matchers/have_reported_error_spec.rb
skatkov: I'm not quite sure how to handle this case. Should this not allow chaining?

pirj: I’m on the fence with this. It might be useful, but is it worth the effort in this introductory PR?
The matcher _can_ be made chainable, but is it worth the effort?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,186 @@
...
+            actual_error.is_a?(@expected_error)
+          when Exception
+            actual_error.is_a?(@expected_error.class) &&
+              (@expected_error.message.empty? || actual_error.message == @expected_error.message)
+          when Regexp
+            actual_error.message&.match(@expected_error)
+          when Symbol
+            actual_error == @expected_error
+          else
+            true
```
pirj: It slipped my mind, what might this other value be?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,186 @@
...
+          end
+        end
+
+        def attributes_match_if_specified?
+          return true if @attributes.empty?
+          return false if @error_subscriber.events.empty?
+
+          event_context = @error_subscriber.events.last[1].with_indifferent_access["context"]
```
pirj: Does it lock and match only on the last one reported?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,186 @@
...
+          ::Rails.error.subscribe(@error_subscriber)
+
+          block&.call
+
+          return false if @error_subscriber.events.empty? && !@expected_error.nil?
+          return false unless error_matches_expectation?
+
+          return attributes_match_if_specified?
+        ensure
+          ::Rails.error.unsubscribe(@error_subscriber) if @error_subscriber
```
pirj:
```suggestion
          ::Rails.error.unsubscribe(@error_subscriber)
```
I can’t imagine that it is nil

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,186 @@
...
+
+        def attributes_match_if_specified?
+          return true if @attributes.empty?
+          return false if @error_subscriber.events.empty?
+
+          event_context = @error_subscriber.events.last[1].with_indifferent_access["context"]
+          attributes_match?(event_context)
+        end
+
+        def actual_error
```
pirj: Cosmetic. [0] and [1] look cryptic, is it worth using a struct?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,186 @@
...
+          when Symbol
+            actual_error == @expected_error
+          else
+            true
+          end
+        end
+
+        def attributes_match_if_specified?
+          return true if @attributes.empty?
+          return false if @error_subscriber.events.empty?
```
pirj: Is this needed? There’s a similar check on line 41

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: @pirj yeah, it does. I've been debating this, but this works in our case just fine. 

But questions keep popping in my head. Should we look through array of errors? what to do if there will be more than 1? Should we match amount of errors too? 

I don't have good answers for these. So this is my dumbest approach -- checking last error only.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: sounds good!

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: It's defensive programming at it's best :) removed

features/matchers/README.md
skatkov: It is pretty realistic as a generic error reporting interface. Guides are very explicit about this interface and offers it usage.

spec/rspec/rails/matchers/have_reported_error_spec.rb
skatkov: @pirj I agree, that we can drop this for now.

But not sure how to properly disable chaining, while leaving support for "with"?

skatkov: I disabled "and" chaining, but not sure how canonical this method is to rspec-rails :)

spec/rspec/rails/matchers/have_reported_error_spec.rb
skatkov: Good idea! I'm raising an ArgumentError now.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: good catch! removed

spec/rspec/rails/matchers/have_reported_error_spec.rb
pirj: It’s a bummer that [include](https://github.com/rspec/rspec-rails/blob/6135d5d0bdfd80d6e3051428639facbab6bd831f/lib/rspec/rails/matchers/base_matcher.rb#L8) Composable is in the base class.

lib/rspec/rails/matchers/have_reported_error.rb
pirj: We should match like we do with scheduled jobs, if there is a matching one among reported, it matches.

I don’t see a reason to add count matches now. This can be added later.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+          @events << ErrorEvent.new(error, attrs.with_indifferent_access)
+        end
+      end
+
+      # Matcher class for `have_reported_error`. Should not be instantiated directly.
+      #
+      # @private
+      # @see RSpec::Rails::Matchers#have_reported_error
+      class HaveReportedError < RSpec::Rails::Matchers::BaseMatcher
+        def initialize(expected_error = nil)
```
pirj: We usually make a distinction between an explicitly passed nil (sometimes by accident), and no argument. You can see the UNSPECIFIED const in the BaseMatcher for an example.

features/matchers/README.md
```diff
@@ -24,3 +24,10 @@ expect(response).to render_template(template_name)
 # and it is not persisted
 expect(assigns(:widget)).to be_a_new(Widget)
 ```
+
+### error reporting
+
+```ruby
+# passes if Rails.error.report was called with specific error instance and message
```
JonRowe:
```suggestion
# passes when `Rails.error.report` is called with specific error instance and message
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
+Feature: `have_reported_error` matcher
+
+  The `have_reported_error` matcher is used to check if an error was reported
+  to Rails error reporting system (`Rails.error`). It can match against error
+  classes, instances, messages, and attributes.
+
+  The matcher supports several matching strategies:
+  * Any error reported
+  * Specific error class
```
JonRowe:
```suggestion
  * A specific error class
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+  * Error message patterns using regular expressions
+  * Error attributes using `.with()`
+  * Symbol errors
+
+  The matcher is available in all spec types where Rails error reporting is used.
+
+  Background:
+    Given a file named "app/models/user.rb" with:
+      """ruby
+      class User < ApplicationRecord
```
JonRowe, line 20: You could clean the custom error up a bit...
```suggestion
      class User < ApplicationRecord
        class ValidationError < StandardError; end
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+        def self.process_custom_error
+          Rails.error.report(ValidationError.new("Email is invalid"))
+        end
+      end
+      """
+    And a file named "app/errors/validation_error.rb" with:
+      """ruby
+      class ValidationError < StandardError
+      end
+      """
```
JonRowe:
```suggestion
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+
+RSpec.describe "have_reported_error matcher" do
+  class TestError < StandardError; end
+  class AnotherTestError < StandardError; end
+
+  it "has an reports_error alias" do
+    expect {Rails.error.report(StandardError.new("test error"))}.to reports_error
+  end
+
+  it "warns that passing value expectation doesn't work" do
```
JonRowe:
```suggestion
  it "warns when used as a value expectation" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
+require "rspec/rails/matchers/have_reported_error"
+
+RSpec.describe "have_reported_error matcher" do
+  class TestError < StandardError; end
+  class AnotherTestError < StandardError; end
+
+  it "has an reports_error alias" do
```
JonRowe:
```suggestion
  it "is aliased as reports_error" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    expect {Rails.error.report(StandardError.new("test error"))}.to reports_error
+  end
+
+  it "warns that passing value expectation doesn't work" do
+    expect {
+      expect(Rails.error.report(StandardError.new("test error"))).to have_reported_error
+    }.to raise_error(ArgumentError, "this matcher doesn’t work with value expectations")
+  end
+
+  describe "basic functionality" do
```
JonRowe:
```suggestion
  context "without constraint" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+        expect { "no error" }.to have_reported_error
+      }.to fail_with(/Expected the block to report an error, but none was reported./)
+    end
+
+    it "passes when negated and no error is reported" do
+      expect { "no error" }.not_to have_reported_error
+    end
+  end
+
+  describe "error class matching" do
```
JonRowe: All these describes would be better off as a contexts.
```suggestion
  context "constrained to a specific error class" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    it "fails when wrong error class is reported" do
+      expect {
+        expect {
+          Rails.error.report(AnotherTestError.new("wrong error"))
+        }.to have_reported_error(TestError)
+      }.to fail_with(/Expected error to be an instance of TestError, but got AnotherTestError/)
+    end
+  end
+
+  describe "error instance matching" do
```
JonRowe:
```suggestion
  context "constrained to a matching error (class and message)" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    it "fails when error instance has different message" do
+      expect {
+        expect {
+          Rails.error.report(TestError.new("actual message"))
+        }.to have_reported_error(TestError.new("expected message"))
+      }.to fail_with(/Expected error to be TestError with message 'expected message', but got TestError with message: 'actual message'/)
+    end
+  end
+
+  describe "regex pattern matching" do
```
JonRowe:
```suggestion
  context "constrained by regex pattern matching" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    it "passes when error message matches pattern" do
```
JonRowe:
```suggestion
    it "passes when an error message matches the pattern" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      }.to have_reported_error(TestError.new("exact message"))
+    end
+
+    it "passes when error instance has empty expected message" do
+      expect {
+        Rails.error.report(TestError.new("any message"))
+      }.to have_reported_error(TestError.new(""))
+    end
+
+    it "fails when error instance has different message" do
```
JonRowe:
```suggestion
    it "fails when the error has different message to the expected" do
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
+require "rspec/rails/matchers/base_matcher"
+
```
JonRowe: This is already required so shouldn't be necessary.

```suggestion
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+module RSpec
+  module Rails
+    module Matchers
+      # @private
```
JonRowe:
```suggestion
      # @api private
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+        end
+
+        def report(error, **attrs)
+          @events << ErrorEvent.new(error, attrs.with_indifferent_access)
+        end
+      end
+
+      # Matcher class for `have_reported_error`. Should not be instantiated directly.
+      #
+      # @private
```
JonRowe:
```suggestion
      # @api private
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+        end
+
+        def and(_)
+          raise ArgumentError, "Chaining is not supported"
+        end
+
+        def matches?(block)
+          if block.nil?
+            raise ArgumentError, "this matcher doesn’t work with value expectations"
+          end
```
JonRowe: Instead you should set `def supports_value_expectations?; false; end`

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+
+          @error_subscriber = ErrorSubscriber.new
+          ::Rails.error.subscribe(@error_subscriber)
+
+          block&.call
```
JonRowe: Won't be nil
```suggestion
          block.call
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+          when Exception
+            desc = "report a #{@expected_error.class} error"
+            desc += " with message '#{@expected_error.message}'" unless @expected_error.message.empty?
+          when Regexp
+            desc = "report an error with message matching #{@expected_error}"
+          when Symbol
+            desc = "report #{@expected_error}"
+          end
+          desc += " with #{@attributes}" unless @attributes.empty?
+          desc
```
JonRowe: I'd prefer to see this refactored to return the description rather than mutate one.

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect(Rails.error.report(StandardError.new("test error"))).to have_reported_error
+    }.to raise_error(ArgumentError, "this matcher doesn’t work with value expectations")
+  end
+
+  describe "basic functionality" do
+    it "passes when an error is reported" do
+      expect {Rails.error.report(StandardError.new("test error"))}.to have_reported_error
+    end
+
+    it "fails when no error is reported" do
```
JonRowe:
```suggestion
    it "fails when no errors are reported" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect {
+        expect { "no error" }.to have_reported_error
+      }.to fail_with(/Expected the block to report an error, but none was reported./)
+    end
+
+    it "passes when negated and no error is reported" do
```
JonRowe:
```suggestion
    it "passes when negated and no errors are reported" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect { "no error" }.not_to have_reported_error
+    end
+  end
+
+  describe "error class matching" do
+    it "passes when correct error class is reported" do
```
JonRowe:
```suggestion
    it "passes when an error with the correct class is reported" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect { Rails.error.report(TestError.new("test error")) }.to have_reported_error(TestError)
+    end
+
+    it "fails when wrong error class is reported" do
```
JonRowe:
```suggestion
    it "fails when an error with the wrong class is reported" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect {
+        expect {
+          Rails.error.report(AnotherTestError.new("wrong error"))
+        }.to have_reported_error(TestError)
+      }.to fail_with(/Expected error to be an instance of TestError, but got AnotherTestError/)
+    end
+  end
+
+  describe "error instance matching" do
+    it "passes when error instance matches exactly" do
```
JonRowe:
```suggestion
    it "passes with an error that matches exactly" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect {
+        Rails.error.report(TestError.new("exact message"))
+      }.to have_reported_error(TestError.new("exact message"))
+    end
+
+    it "passes when error instance has empty expected message" do
```
JonRowe:
```suggestion
    it "passes any error of the same class if the expected message is empty" do
```

I would also argue this should fail, if you don't care about the message use the class assertion?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+  end
+
+  describe "regex pattern matching" do
+    it "passes when error message matches pattern" do
+      expect {
+        Rails.error.report(StandardError.new("error with pattern"))
+      }.to have_reported_error(/with pattern/)
+    end
+
+    it "fails when error message does not match pattern" do
```
JonRowe:
```suggestion
    it "fails when no error messages match the pattern" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      expect {
+        expect {
+          Rails.error.report(StandardError.new("error without match"))
+        }.to have_reported_error(/different pattern/)
+      }.to fail_with(/Expected error message to match/)
+    end
```
JonRowe, line 102: What happens here when there are two errors and one matches but not the other?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,195 @@
...
+      # This matcher asserts that ActiveSupport::ErrorReporter has received an error report.
+      #
+      # @example Checking for any error
+      #   expect { Rails.error.report(StandardError.new) }.to have_reported_error
+      #
+      # @example Checking for specific error class
+      #   expect { Rails.error.report(MyError.new) }.to have_reported_error(MyError)
+      #
+      # @example Checking for specific error instance with message
+      #   expect { Rails.error.report(MyError.new("message")) }.to have_reported_error(MyError.new("message"))
```
JonRowe: I think this should be broken into either two arguments e.g.
```
have_reported_error(MyError, "message")
have_reported_error(MyError, /me/)
```
Or instead use `with(message:)`

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+              severity: "critical",
+              user_id: 123
+            )
+          }.to have_reported_error(ArgumentError)
+            .with(context: "validation", severity: "critical")
+        end
+      end
+      """
+    When I run `rspec spec/models/user_spec.rb`
+    Then the examples should all pass
```
JonRowe: Incorrect line ending here.

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+      end
+      """
+
+  Scenario: Checking for any error being reported
+    Given a file named "spec/models/user_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe User do
+        it "reports an error during processing" do
```
JonRowe:
```suggestion
        it "reports errors" do
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+          expect {
+            User.process_data
+          }.to have_reported_error
+        end
+      end
+      """
+    When I run `rspec spec/models/user_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking for specific error class
```
JonRowe:
```suggestion
  Scenario: Checking for a specific error class
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+          expect {
+            User.process_with_context
+          }.to have_reported_error(ArgumentError.new("Invalid input"))
+        end
+      end
+      """
+    When I run `rspec spec/models/user_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking error message patterns with regular expressions
```
JonRowe:
```suggestion
  Scenario: Checking error messages using regular expressions
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+    Given a file named "spec/models/user_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe User do
+        it "reports error with message matching pattern" do
```
JonRowe:
```suggestion
        it "reports errors with a message matching a pattern" do
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+          expect {
+            User.process_data
+          }.to have_reported_error(/Processing/)
+        end
+
+        it "reports error with message containing 'failed'" do
+          expect {
+            User.process_data
+          }.to have_reported_error(/failed$/)
+        end
```
JonRowe, line 134: This doesn't add anything, its the same as the scenario above.

```suggestion
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+      end
+      """
+    When I run `rspec spec/models/user_spec.rb`
+    Then the examples should all pass
+
+  Scenario: Checking error attributes using `with`
```
JonRowe:
```suggestion
  Scenario: Constraining error matches to their attributes using `with`
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,239 @@
...
+            expect {
+              post "/users", params: { user: { name: "Test" } }
+            }.to have_reported_error.with(context: "user_creation")
+          end
+        end
+      end
+      """
+    When I run `rspec spec/requests/users_spec.rb`
+    Then the examples should all pass
+
```
JonRowe: This isn't a controller / request specific matcher so these feel out of place, why these, why not every aspect of Rails... I'd just cut them.

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+  end
+
+  describe "failure messages for attribute mismatches" do
```
JonRowe:
```suggestion
  describe "#failure_message" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    it "provides detailed failure message when attributes don't match" do
```
JonRowe:
```suggestion
    it "provides details about mismatched attributes" do
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    it "shows actual context values when attributes don't match" do
+      expect {
+        expect {
+          Rails.error.report(StandardError.new("test"), context: { user_id: 123, context: "actual" })
+        }.to have_reported_error.with(user_id: 456)
+      }.to fail_with(/actual values are {"user_id" => 123, "context" => "actual"}/)
+    end
+  end
+
+  describe "attribute matching with .with chain" do
```
JonRowe:
```suggestion
  describe "#with" do
```

skatkov:
@JonRowe thanks for review.

> The one change I do want to see though is dropping instance matching, I don't think it makes sense over just providing class / message / using with.

Which of these you would rather go with?
`.have_reported_error(StandardError, /wrong attributes/).with(param1: 'test')`

or 

`.have_reported_error(StandardError).with(message:/wrong attributes/, params1: 'test1')`

??

JonRowe:
Given that you already have `with`, I think:
```
.have_reported_error(StandardError).with(/wrong attributes/, params1: 'test1')
```

pirj:
It makes sense to separate matching of args passed to `report` and exception’s own args, including the message. Both can receive args, right? It would open a way for ambiguity and confusion.
`raise_error` accepts `have_attributes` as an argument matcher. It supports `with_message`. Wdyt if we reserve `with` for additional args passed to `report`, but keep it separate from `with_message`?

skatkov:
> It makes sense to separate matching of args passed to `report` and exception’s own args, including the message. Both can receive args, right? It would open a way for ambiguity and confusion. 

The only argument that the exception class accepts is "message". We can of course subclass it and support way more than that, but I'm not sure if we should support this in a matcher?


I have a feeling, that having a matcher interface similar to `raise_error` would make it simpler for people to memorize it.  I don't mind changing `.with` to `.have_attributes` as well.

JonRowe:
I hadn't clocked that the extra attributes were in addition to the error, I think matching the `raise_error` interface in terms of `report_error(ErrorClass, message_or_regexp)` makes sense with the existing `with` becoming something clearer, `with_metadata`? What does Rails call the extra data?

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: Strangely enough, this doesn't seem to be working in my case.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -3,6 +3,11 @@
 module RSpec
   module Rails
     module Matchers
+      # @api private
+      # Sentinel value to distinguish between no argument passed vs explicitly passed nil.
+      # This follows the same pattern as RSpec's raise_error matcher.
+      UndefinedValue = Object.new.freeze
```
pirj: Such a constant is already defined in the basematcher, do we need another one?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -28,17 +33,20 @@ def report(error, **attrs)
...
         #   Expected message when first param is a class
-        def initialize(expected_error_or_message = nil, expected_message = nil)
+        def initialize(expected_error_or_message = UndefinedValue, expected_message = nil)
           @actual_error = nil
           @attributes = {}
           @error_subscriber = nil
           
           case expected_error_or_message
-          when nil
+          when nil, UndefinedValue
```
pirj: We usually make a distinction between those two, where nil is typically a sign of a mistake. For some matchers, we warn if nil is passed. Do you think there can be legitimate cases of passing nil as an argument?

skatkov: Yeah, similar to error_raised you can just expect an error to be raise, but don't care which one.

pirj: The thing is `raise_error()` is to expect any exception, no matter which; but `raise_error(nil)` is usually an indication of a mistake in the spec itself. See https://github.com/rspec/rspec-expectations/issues/1142
[This warning](https://github.com/rspec/rspec/blob/95fbf88255de04e4ba8288172292bda494c4f187/rspec-expectations/lib/rspec/matchers/built_in/raise_error.rb#L204) may explain it better.

Similarly, `expect { report… }.to have_reported_error()` should pass when any kind of error has been reported, while passing a nil as the first argument should emit a warning.

skatkov: Ok! Adjusted.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+                           ""
+                         end
+
+          attributes_desc = @attributes.empty? ? "" : " with #{@attributes}"
+
+          base_desc + message_desc + attributes_desc
+        end
+
+        def failure_message
+          if !@error_subscriber.events.empty? && !@attributes.empty?
```
pirj: Cosmetic: `any?` or `present?`?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+            event_context = @error_subscriber.events.last.attributes[:context]
```
pirj: Won’t it be confusing to use `last` when several errors were reported?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+              return "Expected error to be an instance of #{@expected_error} with message '#{@expected_message}', but got: #{reported_errors}"
+            elsif @expected_error
+              return "Expected error to be an instance of #{@expected_error}, but got: #{reported_errors}"
+            elsif @expected_message.is_a?(Regexp)
+              return "Expected error message to match #{@expected_message}, but got: #{reported_errors}"
+            elsif @expected_message.is_a?(String)
+              return "Expected error message to be '#{@expected_message}', but got: #{reported_errors}"
+            end
+          else
+            if @expected_error && !actual_error.is_a?(@expected_error)
```
pirj: actual_error should be is_a? expected_class, no? Just by looking at the implementation. Or what is the case when they won’t match?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+          @expected_error.nil? || error.is_a?(@expected_error)
+        end
+
+        # Check if the given error message matches the expected message pattern
+        def error_message_matches?(error)
+          return true if @expected_message.nil?
+
+          case @expected_message
+          when Regexp
+            error.message&.match(@expected_message)
```
pirj: Cosmetic: is safe nav necessary?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        }.to have_reported_error(StandardError, /different pattern/)
+      }.to fail_with(/Expected error to be an instance of StandardError with message/)
+    end
+  end
+
+  describe "#failure_message" do
+    it "provides details about mismatched attributes" do
+      expect {
+        expect {
+          Rails.error.report(StandardError.new("test"), context: { user_id: 123, context: "actual" })
```
pirj: Nested `context` key is confusing here. Do you mind to change the inner one to something else?

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        }.to have_reported_error.with_context(user_id: 456, context: "expected")
+      }.to fail_with(/Expected error attributes to match {user_id: 456, context: "expected"}, but got these mismatches: {user_id: 456, context: "expected"} and actual values are {"user_id" => 123, "context" => "actual"}/)
```
pirj: Can you please fix specs?

Expected error attributes to match {user_id: 456, context: "expected"}, but got these mismatches: {user_id: 456, context: "expected"} and actual values are {"user_id" => 123, "context" => "actual"}/, got #<RSpec::Expectations::ExpectationNotMetError: Expected error attributes to match {:user_id=>456, :context=>"expected"}, but got these mismatches: {:user_id=>456, :context=>"expected"} and actual values are {"user_id"=>123, "context"=>"actual"}

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+          error_count = @error_subscriber.events.count
+          error_word = 'error'.pluralize(error_count)
+          verb = error_count == 1 ? 'has' : 'have'
+
+          "Expected the block not to report any errors, but #{error_count} #{error_word} #{verb} been reported."
+        end
+
+        private
+
+        def error_matches_expectation?
```
pirj: Is the last condition necessary? We already check if it’s `empty?` before calling this method.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
+module RSpec
+  module Rails
+    module Matchers
+      # @api private
+      # Sentinel value to distinguish between no argument passed vs explicitly passed nil.
+      # This follows the same pattern as RSpec's raise_error matcher.
+      UndefinedValue = Object.new.freeze
+
+      # @api private
+      class ErrorSubscriber
```
pirj: They have https://github.com/rails/rails/blob/57093484e7926fb33ff266aa25b3c6241b235a68/activesupport/lib/active_support/testing/error_reporter_assertions.rb#L6, can we reuse it?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+            @expected_error = nil
+            @expected_message = expected_error_or_message
+          else
+            @expected_error = expected_error_or_message
+            @expected_message = expected_message
+          end
+        end
+
+        def with_context(expected_attributes)
+          @attributes.merge!(expected_attributes)
```
pirj, line 42: Cosmetic: Do we need to merge? Maybe just assign?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+          self
+        end
+
+        def and(_)
+          raise ArgumentError, "Chaining is not supported"
+        end
+
+        def matches?(block)
+          if block.nil?
+            raise ArgumentError, "this matcher doesn't work with value expectations"
```
pirj: Does this check have the same effect as defining a method `supports_value_expectations?` that returns false? It should be.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+          end
+
+          warn_about_nil_error! if @warn_about_nil_error
+
+          @error_subscriber = ErrorSubscriber.new
+          ::Rails.error.subscribe(@error_subscriber)
+
+          block.call
```
pirj: This will become something like `reports = ErrorCollector.record(&block)`

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+
+        def error_class_matches?(error)
+          @expected_error.nil? || error.is_a?(@expected_error)
+        end
+
+        # Check if the given error message matches the expected message pattern
+        def error_message_matches?(error)
+          return true if @expected_message.nil?
+
+          case @expected_message
```
pirj, line 152: There should be existing examples amongst our matchers of something like `value_matches?` that would also allow the use of argument matchers like `a_string_including("network")`.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+        def attributes_match_if_specified?
+          return true if @attributes.empty?
+          return false unless matching_event
+
+          event_context = matching_event.attributes[:context]
+          attributes_match?(event_context)
+        end
+
+        def actual_error
+          @actual_error ||= matching_event&.error
```
pirj: Is it worth memoizing?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+      #
+      # @example Checking error attributes
+      #   expect { Rails.error.report(StandardError.new, context: "test") }.to have_reported_error.with_context(context: "test")
+      #
+      # @example Negation
+      #   expect { "safe code" }.not_to have_reported_error
+      #
+      # @param expected_error_or_message [Class, String, Regexp, nil] the expected error class, message string, or message pattern
+      # @param expected_message [String, Regexp, nil] the expected error message to match
+      def have_reported_error(expected_error_or_message = UndefinedValue, expected_message = nil)
```
pirj: We have default attributes both here and in the initializer. Worth leaving just here?

pirj (commented):
A fee minor things and simplifications and it looks good to go.

Let’s leave out matching multiple with chains, severity qualifiers, and all extras for later.

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,199 @@
...
+    Given a file named "app/models/user.rb" with:
+      """ruby
+      class User < ApplicationRecord
+        class ValidationError < StandardError; end
+        def self.process_data
+          Rails.error.report(StandardError.new("Processing failed"))
+        end
+
+        def self.process_with_context
+          Rails.error.report(ArgumentError.new("Invalid input"), context: { context: "user_processing", severity: :error })
```
pirj: Nested context here, too. Rename inner to `topic:` or `:section`?

pirj: The `severity:` here is inside the `context:`, is this correct?

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,199 @@
...
+    Given a file named "spec/models/user_spec.rb" with:
+      """ruby
+      require "rails_helper"
+
+      RSpec.describe User do
+        it "does not report any errors for safe operations" do
+          expect {
+            # Safe operation that doesn't report errors
+            "safe code"
+          }.not_to have_reported_error
```
pirj, line 194: I wonder if the negated form should warn if qualifiers were passed to avoid mistakenly letting unmatched reports through:
```ruby
          expect {
            User.process_custom_error
          }.not_to have_reported_error("Email is TYPO")
```

Meaning that if the matcher is used in the negated form, no class, message, or attributes qualifiers are acceptable.

See https://github.com/rspec/rspec/blob/95fbf88255de04e4ba8288172292bda494c4f187/rspec-expectations/lib/rspec/matchers/built_in/raise_error.rb#L162

lib/rspec/rails/matchers/have_reported_error.rb
pirj: On a second thought, this may make matching more complicated, as if just one argument is passed, and it’s an argument matcher, what should we match it against, the exception itself, or the exception’s message?

This makes it more compelling to build the `with_message` qualifier. But it then limits the argument to the matcher to only accept the class. But this is what the Rails’ assertion does anyway.

pirj:
Ping

spec/rspec/rails/matchers/have_reported_error_spec.rb
skatkov: Uh, resolved.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: RSpec is actually allowing the value expectation to reach the `matches?` method, even with `supports_value_expectations?` defined as false. This mechanism is NOT preventing the value expectation from reaching `matches?`. The explicit check was necessary in my testing! But maybe I missed something?

pirj: Ha! Nice catch. The `RSpec.deprecate` call somehow didn't fail the build. I'll have a closer look.

The `block` is not `nil` though for `expect(:foo).to reports_error`, so this check is useless in its current implementation.
I suggest removing it, and relying on the `supports_value_expectations?`.

pirj: I recall now that we kept in 3.x the possibility to pass a lambda to `expect`, or define `subject` as a lambda and use `is_expected.to` with block matchers. We just [print a deprecation message now](https://github.com/rspec/rspec-expectations/pull/1139).

Swallowed deprecation messages are due to https://github.com/rspec/rspec-rails/pull/2857

pirj: After the aftermentioned fix, with

```ruby
        def supports_value_expectations?
          false
        end
```

The following passes:
```ruby
  it "warns when used as a value expectation" do
    expect {
      expect(Rails.error.report(StandardError.new("test error"))).to have_reported_error
    }.to raise_error(/implicit block expectation.+deprecated/)
  end
```

We can change the code when that swallowed deprecation fix is merged.

features/matchers/README.md
```diff
@@ -24,3 +24,30 @@ expect(response).to render_template(template_name)
...
+# passes when any error with exact message is reported
+expect { Rails.error.report(StandardError.new("exact message")) }.to have_reported_error("exact message")
+
+# passes when any error with message matching pattern is reported
+expect { Rails.error.report(StandardError.new("test message")) }.to have_reported_error(/test/)
+
+# passes when error is reported with specific context attributes
+expect { Rails.error.report(StandardError.new, context: { user_id: 123 }) }.to have_reported_error.with_context(user_id: 123)
+
+
```
JonRowe: Remove these two blank lines, I'd do it in a suggestion but Github isn't co-operating.

features/matchers/README.md
JonRowe: This is very verbose for us... These lines take up half this file when our primary documentation is the feature file itself...

log/development.log
JonRowe, line 1: Don't commit this.

lib/rspec/rails/matchers/have_reported_error.rb
JonRowe: No this is fine, I think it would be expected, consider if you compose off one.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,254 @@
...
+          end
+        end
+
+        def with_context(expected_attributes)
+          @attributes.merge!(expected_attributes)
+          self
+        end
+
+        def and(_)
+          raise ArgumentError, "Chaining is not supported"
```
JonRowe, line 47: Why? If you include the composition helper this should just work ™ ?

lib/rspec/rails/matchers/have_reported_error.rb
JonRowe: This can be removed now and any test marked as pending with the reason instead.

spec/rspec/rails/matchers/have_reported_error_spec.rb
JonRowe: Ping?

lib/rspec/rails/matchers/have_reported_error.rb
pirj: Agreed, chaining another `with_context(...)` on an existing `have_reported_error.with_context(user_id: 123)`, should not overwrite the first one, at least without warning that it does.
@skatkov can you please add a spec example to solidify this behaviour?

features/matchers/README.md
pirj:
```suggestion
```

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: Done.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: Currently, `.merge!` will overwrite previously defined `with_context(..)`. So I modified this code to detect possible conflicting keys and raise a warning.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: Resolved!

features/matchers/have_reported_error_matcher.feature
skatkov: It's not :)

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: Oh, tried this previously, but now this works as expected!

skatkov: oh, yeah, the problem here is that some chaining is allowed (`.with_context` as example). But I don't want to allow `expect {}.to have_reported_error("error message").and have_reported_error`.

I'm not sure how to handle such cases correctly.

features/matchers/have_reported_error_matcher.feature
skatkov: It's a good idea. If `raise_error` already works like that, I'd like to keep it similar here as well.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: This particular method checks that the error matches one that we expect, not just that the error occurred.

lib/rspec/rails/matchers/have_reported_error.rb
skatkov: yeah, sometimes `error` object could be nil.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+        # @param expected_message [String, Regexp, nil]
+        #   Expected message when first param is a class
+        def initialize(expected_error_or_message, expected_message)
+          @attributes = {}
+          @warn_about_nil_error = expected_error_or_message.nil?
+
+          case expected_error_or_message
+          when UNDEFINED
+            @expected_error = nil
+            @expected_message = expected_message
```
pirj: If the first argument was not passed at all and defaulted to `UNDEFINED`, the second, `expected_message` will also be `nil`?
```suggestion
            @expected_error = @expected_message = nil
```

lib/rspec/rails/matchers/have_reported_error.rb
pirj: There are still defaults both here and [there](https://github.com/rspec/rspec-rails/pull/2849/files/c1816e2f2b14a547e26ddfc0b2b3d9a02c4b9286#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R38):
```
def initialize(expected_error_or_message = UndefinedValue, expected_message = nil)
```
Let's remove there.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,271 @@
...
+      # @see RSpec::Rails::Matchers#have_reported_error
+      class HaveReportedError < RSpec::Rails::Matchers::BaseMatcher
+        # Uses UndefinedValue as default to distinguish between no argument
+        # passed vs explicitly passed nil.
+        #
+        # @param expected_error_or_message [Class, String, Regexp, nil]
+        #   Error class, message string, or message pattern
+        # @param expected_message [String, Regexp, nil]
+        #   Expected message when first param is a class
+        def initialize(expected_error_or_message = UndefinedValue, expected_message = nil)
```
pirj:
```suggestion
        def initialize(expected_error_or_message, expected_message)
```

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        }.to have_reported_error.with_context(user_id: 456)
+      }.to fail_with(/actual values are {"user_id" => 123, "context" => "actual"}/)
+    end
+  end
+
+  describe "#with_context" do
+    it "passes when attributes match exactly" do
+      expect {
+        Rails.error.report(StandardError.new("test"), context: { user_id: 123, context: "test" })
+      }.to have_reported_error.with_context(user_id: 123, context: "test")
```
pirj: Nested `context` strikes back.

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+    end
+
+    it "passes with partial attribute matching" do
+      expect {
+        Rails.error.report(
+          StandardError.new("test"), context: { user_id: 123, context: "test", extra: "data" }
```
pirj: Here, too

spec/rspec/rails/matchers/have_reported_error_spec.rb
```diff
@@ -0,0 +1,213 @@
...
+        Rails.error.report(AnotherTestError.new("error with pattern"))
+      }.to have_reported_error(/with pattern/)
+    end
+
+    it "fails when no error with exact message is reported" do
+      expect {
+        expect {
+          Rails.error.report(StandardError.new("actual message"))
+        }.to have_reported_error("expected message")
+      }.to fail_with(/Expected error message to be 'expected message', but got: StandardError/)
```
pirj: Is that all the failure message? Does it include the actual exception message? Or is it just skipped for brevity in this expectation?

features/matchers/README.md
pirj:
```suggestion
```

features/matchers/have_reported_error_matcher.feature
pirj: It's still there

pirj:
```suggestion
          Rails.error.report(ArgumentError.new("Invalid input"), context: { section: "user_processing" }, severity: :error)
```

lib/rspec/rails/matchers/have_reported_error.rb
pirj: I've unresolved a few notes that were not answered/addressed. Please check.

lib/rspec/rails/matchers/have_reported_error.rb
pirj: How? Please add a spec for this case.

features/matchers/README.md
```diff
@@ -24,3 +24,28 @@ expect(response).to render_template(template_name)
...
+expect { Rails.error.report(MyError.new("test message")) }.to have_reported_error(MyError, /test/)
+
+# passes when any error with exact message is reported
+expect { Rails.error.report(StandardError.new("exact message")) }.to have_reported_error("exact message")
+
+# passes when any error with message matching pattern is reported
+expect { Rails.error.report(StandardError.new("test message")) }.to have_reported_error(/test/)
+
+# passes when error is reported with specific context attributes
+expect { Rails.error.report(StandardError.new, context: { user_id: 123 }) }.to have_reported_error.with_context(user_id: 123)
```
pirj:
```suggestion
```

features/matchers/have_reported_error_matcher.feature
```diff
@@ -0,0 +1,199 @@
...
+    Given a file named "app/models/user.rb" with:
+      """ruby
+      class User < ApplicationRecord
+        class ValidationError < StandardError; end
+        def self.process_data
+          Rails.error.report(StandardError.new("Processing failed"))
+        end
+
+        def self.process_with_context
+          Rails.error.report(ArgumentError.new("Invalid input"), severity: :error , context: { topic: "user_processing"})
```
pirj:
```suggestion
          Rails.error.report(ArgumentError.new("Invalid input"), severity: :error, context: { topic: "user_processing" })
```

lib/rspec/rails/matchers/have_reported_error.rb
pirj: [Here](https://github.com/rspec/rspec-rails/pull/2849/files#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R58) we check if `@reports.empty?` and return.

Right after the check, we call `error_matches_expectation?`.

The last condition check `@reports.count.positive?` [here](https://github.com/rspec/rspec-rails/pull/2849/files#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R139) is redundant.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+          error_word = 'error'.pluralize(error_count)
+          verb = error_count == 1 ? 'has' : 'have'
+
+          "Expected the block not to report any errors, but #{error_count} #{error_word} #{verb} been reported."
+        end
+
+        private
+
+        def error_matches_expectation?
+          return true if @expected_error.nil? && @expected_message.nil? && @reports.count.positive?
```
pirj:
```suggestion
          return true if @expected_error.nil? && @expected_message.nil?
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+            elsif @expected_message
+              return "Expected error message to #{@expected_message.is_a?(Regexp) ? "match" : "be" } #{@expected_message}, but got: '#{actual_error.message}'"
+            else
+              return "Expected specific error, but got #{actual_error.class} with message: '#{actual_error.message}'"
+            end
+          end
+        end
+
+        def failure_message_when_negated
+          warn_about_negated_with_qualifiers! if has_qualifiers?
```
pirj: ⚠️ This is a bad place to warn.
There might be no failure due to a false positive.
Please move to `does_not_match?` like [here](https://github.com/rspec/rspec-expectations/blob/ba31727e856de42abb5a2e6566855f0831e1a619/lib/rspec/matchers/built_in/raise_error.rb#L84).

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+        end
+
+        def matching_report
+          @matching_report ||= find_matching_report
+        end
+
+        def find_matching_report
+          @reports.find do |report|
+            error_class_matches?(report.error) && error_message_matches?(report.error)
+          end
```
pirj: What if there were two errors reported, same class, same message, but the attributes (context) only match on the second one, not the first one that `find` finds?

There is some repetition between this method and [those lines](https://github.com/rspec/rspec-rails/pull/2849/files#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R141-R143).

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+
+        def attributes_match?(actual)
+          @attributes.all? do |key, value|
+            actual_value = actual[key] || actual[key.to_s] || actual[key.to_sym]
+            if value.respond_to?(:matches?)
+              value.matches?(actual_value)
+            else
+              actual_value == value
+            end
+          end
```
pirj, line 182: We have [this](https://github.com/rspec/rspec/blob/12effb26900129ba2c977585ade9938d18f753ce/rspec-support/lib/rspec/support/fuzzy_matcher.rb#L10), should fit nicely here:
```suggestion
          values_match?(@attributes, actual)
```

There is a difference, though - we don't coerce keys to strings. I'm more inclined to be strict here and don't coerce unless you have a good argument for being indifferent to keys' type.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+        def unmatched_attributes(actual)
+          @attributes.reject do |key, value|
+            actual_value = actual[key] || actual[key.to_s] || actual[key.to_sym]
+            if value.respond_to?(:matches?)
+              value.matches?(actual_value)
+            else
+              actual_value == value
+            end
+          end
+        end
```
pirj: Maybe just report the entire actual context for simplicity, not the diff?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+                           ""
+                         end
+
+          attributes_desc = @attributes.empty? ? "" : " with #{@attributes}"
+
+          base_desc + message_desc + attributes_desc
+        end
+
+        def failure_message
+          if !@reports.empty? && !@attributes.empty?
```
pirj:
```suggestion
          if @reports.any? && @attributes.any?
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+            report_context = @reports.last.context
```
pirj: `last` is incorrect if several errors could have been reported.

pirj: If there were no errors reported matching by class/message, even if the expectation has a `with_context` qualifier, don't even mention context attributes in the failure message.

But if **some** reports did match by class/message (or **ALL** if the expectation did not specify class/message/regexp), list those errors with their attributes.

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+        def attributes_match_if_specified?
+          return true if @attributes.empty?
+          return false unless matching_report
+
+          report_context = matching_report.context
+          attributes_match?(report_context)
+        end
+
+        def actual_error
+          @actual_error ||= matching_report&.error
```
pirj:
```suggestion
          matching_report&.error
```
Memoization feels redundant

lib/rspec/rails/matchers/have_reported_error.rb
pirj: `matching_reports` (plural) should do better

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,272 @@
...
+            elsif @expected_error
+              return "Expected error to be an instance of #{@expected_error}, but got: #{reported_errors}"
+            elsif @expected_message.is_a?(Regexp)
+              return "Expected error message to match #{@expected_message}, but got: #{reported_errors}"
+            elsif @expected_message.is_a?(String)
+              return "Expected error message to be '#{@expected_message}', but got: #{reported_errors}"
+            end
+          else
+            if @expected_error && !actual_error.is_a?(@expected_error)
+              return "Expected error to be an instance of #{@expected_error}, but got #{actual_error.class} with message: '#{actual_error.message}'"
```
pirj: How does this work?
`actual_error` is `matching_report&.error` where `matching_report` [is](https://github.com/rspec/rspec-rails/pull/2849/files#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R181-R182):
```ruby
          @reports.find do |report|
            error_class_matches?(report.error) && error_message_matches?(report.error)
```
and `error_class_matches?` [is](https://github.com/rspec/rspec-rails/pull/2849/files#diff-5095b0a095ba1b0511036361c8930a18ff4367591813576e280b69bb823bf655R146):
```ruby
@expected_error.nil? || error.is_a?(@expected_error)
```

Above, we check that `@expected_error` is not `nil`.

How on Earth can `actual_error` not be of `@expected_error` class?

Reduced:
```ruby
actual_error.is_a?(@expected_error) && !actual_error.is_a?(@expected_error)
```

How does [that spec](https://github.com/rspec/rspec-rails/pull/2849/files#diff-f50b1ae989c61af9699f390696ecc8bd71e21a274a6b75094395cb58cb9d4ef1R88) pass?

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,270 @@
...
+            raise ArgumentError, "Attribute keys #{conflicting_keys.inspect} are already defined. " \
+                                "Chaining with_context calls should not overwrite existing attributes. " \
+                                "Use a single with_context call with all attributes instead."
+          end
+          @attributes.merge!(expected_attributes)
+          self
+        end
+
+        def and(_)
+          raise ArgumentError, "Chaining is not supported"
```
pirj, line 47:
```suggestion
        def and(_)
          raise ArgumentError, "Chaining is not supported"
        end

        def or(_)
          raise ArgumentError, "Chaining is not supported"
```

lib/rspec/rails/matchers/have_reported_error.rb
```diff
@@ -0,0 +1,270 @@
...
+          RSpec.warn_with("Using `expect { }.not_to have_reported_error(error_class_or_message)` " \
+                         "or with `.with_context()` is prone to false positives, since any error " \
+                         "that doesn't match the specific error class, message, or context can " \
+                         "cause the expectation to pass. Instead consider using " \
+                         "`expect { }.not_to have_reported_error` with no qualifiers to ensure " \
+                         "that no errors are reported at all.")
+        end
+
+        def has_qualifiers?
+          !@expected_error.nil? || !@expected_message.nil? || !@attributes.empty?
```
pirj:
```suggestion
          @expected_error || @expected_message || @attributes.any?
```

pirj:
If you revase, the problem with swallowed depreciation warnings should be fixed, and you don’t have to manually check if a block was passed to match, leaving `supports_value_expectations() => false` to do the job

spec/rspec/rails/matchers/have_reported_error_spec.rb
skatkov: It should pass if any reported error fully matches the expectation.

I updated the matcher so it checks all captured reports rather than effectively depending on the first candidate, and added specs for:
- multiple reported errors where a later class/message match should pass
- multiple reported errors where only the later report has matching context

PR #2856, opened 2025-07-18
Try to drop seleinum

voxik:


pirj:
Why does this command fail for you? What if you run this manually?
```
 /usr/share/gems/gems/selenium-webdriver-4.27.0/bin/linux/selenium-manager --browser chrome --language-binding ruby --output json
```

JonRowe:
Apologies for the delay here, as far as I can tell / remember the point of these is to ensure the tests are running on a javascript capable runner, but as the test on line 33 shows, the default is a selenium driver...

So I guess my question is why they work when removing the explicit driver selection when the default should also be chromedriver...

I'm not adversed to merging this, but if theres a more complex behaviour going on I'd like to understand it.

PR #2857, opened 2025-08-06
Reduce sandboxing to prevent swallowed deprecations

pirj:
We used to sandbox everything, but it turns out that deprecations (`RSpec.deprecate("anything")`) were swallowed, as the temporary config has no deprecation stream set, or stderr is masked.
Found here https://github.com/rspec/rspec-rails/pull/2849#discussion_r2256730128

We're almost back to square one with deprecations as we were before https://github.com/rspec/rspec-rails/pull/2365

However, I'm not entirely happy with this approach. We still mute deprecations for those Rails-related example group specs and a few others, and there are chances deprecations will pop up there eventually.

:question: Would it be better to replay the configuration block on top of the sandboxed configuration? Probably not, and it might not work if stderr is intercepted.

Please consider this WIP

### DIFF
--- a/spec/rspec/rails/view_rendering_spec.rb
+++ b/spec/rspec/rails/view_rendering_spec.rb
@@ -1,5 +1,5 @@
 module RSpec::Rails
-  RSpec.describe ViewRendering do
+  RSpec.describe ViewRendering, :with_isolated_config do
     let(:group) do
       RSpec::Core::ExampleGroup.describe do
         def controller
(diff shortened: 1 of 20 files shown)


spec/rspec/rails/view_rendering_spec.rb
```diff
@@ -53,7 +53,7 @@ def example.controller
           expect(group.new.render_views?).to be_falsey
         end
 
-        it "overrides the global config if render_views is enabled there" do
+        it "overrides the global config if render_views is enabled there", :with_isolated_config do
```
JonRowe: This is doubled up no?

pirj:
> we should at least extend the sandboxed config to not swallow output

@JonRowe Done

pirj:
CI green, @JonRowe how do you feel about this change?

PR #2858, opened 2025-08-08
Add .tool-versions to .gitignore and create AuthenticationGenerator

rodrigotoledo:
## Description

This change addresses an issue where running the `factory_bot` generator could result in a `factory_bot [not found]` error.  
The problem occurred due to missing generator definitions and template files.

With this update, the authentication generator is properly defined and includes the necessary templates, ensuring that the `users.rb` factory file is generated regardless of whether the project is using RSpec (`spec/factories`) or Minitest (`test/factories`).  
This guarantees consistent behavior and avoids generator errors in environments without a `spec` directory.

## Background

Previously, projects without a `spec` directory or with incomplete generator/template setup would fail when invoking `factory_bot` generators, making it impossible to scaffold the necessary factory files automatically.

## Changes Introduced

- Added the missing `factory_bot:authentication` generator definition.
- Included `users.rb` template under `templates` directory.
- Implemented logic to detect the test framework and place the generated factory file in the correct directory (`spec/factories` or `test/factories`).
- Ensured generator runs successfully regardless of the presence of a `spec` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ vendor
 Gemfile.lock
 bin
 .rvmrc
+.tool-versions
 *.rbc
 .yardoc
 .bundle
--- a/lib/generators/factory_bot/authentication/authentication_generator.rb
+++ b/lib/generators/factory_bot/authentication/authentication_generator.rb
@@ -1,0 +1,25 @@
+require 'generators/factory_bot'
+require 'factory_bot_rails'
+
+module FactoryBot
+  module Generators
+    class AuthenticationGenerator < Rails::Generators::Base
+      # Use the Source_root Setter correctly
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ vendor
 Gemfile.lock
 bin
 .rvmrc
+.tool-versions
```
pirj, line 10: We don't mind adding this line, though.

I, however, prefer adding those to my [global git config](https://github.com/pirj/dotfiles/blob/c358379ce0b7a3675920d1309f8649f30814c36f/.config/git/ignore#L7).

JonRowe: I think these are best off ignored in global configs

PR #2863, opened 2025-08-18
Add request specs generation to authentication generator

DiegoPisa2003:
Rails' built-in authentication generator (TestUnit) generates four test files when creating authentication scaffolding. However, the RSpec equivalent was only generating two of these tests - the model spec and fixtures. This PR adds the missing request specs for sessions and passwords functionality, bringing feature parity between TestUnit and RSpec authentication generators.

## What's Changed

- Generator Enhancements:
  - Added `--request-specs` option (defaults to true) to control request spec generation
  - Added `create_session_request_spec` method to generate session authentication tests
  - Added `create_password_request_spec` method to generate password reset tests

- New Test Templates:
  - `session_spec.rb` - Tests for user login, logout, and session management
  - `password_spec.rb` - Tests for password reset flow including email sending and token validation

- Test Coverage:
  - Added comprehensive test coverage for the new functionality
  - Tests verify files are created by default
  - Tests verify `--request-specs=false` option works correctly
  - Tests ensure compatibility with fixture replacement options
  
## References

[Based on Rails' TestUnit authentication generator](https://github.com/rails/rails/blob/main/railties/lib/rails/generators/test_unit/authentication/authentication_generator.rb)

### DIFF
--- a/spec/generators/rspec/authentication/authentication_generator_spec.rb
+++ b/spec/generators/rspec/authentication/authentication_generator_spec.rb
@@ -45,8 +45,8 @@ RSpec.describe Rspec::Generators::AuthenticationGenerator, type: :generator do
       expect(file('spec/rails_helper.rb')).to contain(
         "  # Include authentication support module for request specs\n  config.include AuthenticationSupport, type: :request\n"
       )
-      expect(file('spec/rails_helper.rb')).not_to contain(
-        "  # Rails.root.glob('spec/support/**/*.rb').sort_by(&:to_s).each { |f| require f }\n"
+      expect(file('spec/rails_helper.rb')).to contain(
+        /^#{Regexp.escape("Rails.root.glob('spec/support/**/*.rb').sort_by(&:to_s).each { |f| require f }")}$/
       )
 
       expect(File.exist?(file('spec/support/authentication_support.rb'))).to be true


lib/generators/rspec/authentication/templates/session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      delete session_path
+
+      expect(response).to redirect_to(new_session_path)
+      expect(cookies[:session_id]).to be_empty
+    end
+  end
+
+  private
+
+  def sign_in_as(user)
```
JonRowe: Given Rails's own preference for helpers, I feel this should be generated in a module thats included or configured for automatic inclusion.

lib/generators/rspec/authentication/templates/session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    Current.session = user.sessions.create!
+
+    ActionDispatch::TestRequest.create.cookie_jar.tap do |cookie_jar|
+      cookie_jar.signed[:session_id] = Current.session.id
+      cookies[:session_id] = cookie_jar[:session_id]
+    end
```
JonRowe: I'd also prefer if this did a simple post rather than reaching into internals, seems like it could be brittle?

lib/generators/rspec/authentication/templates/password_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+
+RSpec.describe "Passwords", <%= type_metatag(:request) %> do
+  fixtures :users
+
+  let(:user) { users(:one) }
```
JonRowe: See below

lib/generators/rspec/authentication/templates/session_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe "Sessions", <%= type_metatag(:request) %> do
+  fixtures :users
+
+  let(:user) { users(:one) }
```
JonRowe: Although we do generate fixtures when thats enabled, we don't use them anywhere else, I feel like this should be gated on fixtures being turned on but I'm not sure what the alternative should be?

JonRowe (changes requested):
Great start, I'd love to see this land, just a couple of points to address

lib/generators/rspec/authentication/templates/session_spec.rb
DiegoPisa2003: Yes, you are right, it is cleaner and it is closer to the Rails way. What do you think about adding this two methods to the `authentication_generator`:

`create_authentication_support`:
Creates the `spec/support/authentication_support.rb` file with the `sign_in_as(user)` helper

`configure_authentication_support`:
Automatically modifies `spec/rails_helper.rb` to include the helper in all request specs

lib/generators/rspec/authentication/templates/session_spec.rb
DiegoPisa2003: We could remove fixture usage and use direct model creation like the scaffold generator  [lib/generators/rspec/scaffold/templates/request_spec.rb](https://github.com/rspec/rspec-rails/blob/main/lib/generators/rspec/scaffold/templates/request_spec.rb):

```ruby
# TODO: Replace with your factory or model creation method
# For example, with FactoryBot: let(:user) { create(:user) }
# or with fixtures: let(:user) { users(:one) }
let(:user) { User.create!(email_address: "test@example.com", password: "password") }
```

This option is completely agnostic to fixtures vs factory replacements

spec/generators/rspec/authentication/authentication_generator_spec.rb
```diff
@@ -19,6 +23,48 @@
...
+      File.write(File.join(destination_root, 'spec', 'rails_helper.rb'), rails_helper_content)
+
+      run_generator
+
+      expect(file('spec/rails_helper.rb')).to contain(
+        "  # Include authentication support module for request specs\n  config.include AuthenticationSupport, type: :request\n"
+      )
+      expect(file('spec/rails_helper.rb')).not_to contain(
+        "  # Rails.root.glob('spec/support/**/*.rb').sort_by(&:to_s).each { |f| require f }\n"
+      )
```
JonRowe: Being pedantic, this only asserts the line was removed, I think prehaps match on the line without the #
```suggestion
      expect(file('spec/rails_helper.rb')).to contain(
        /^#{Regexp.escape("Rails.root.glob('spec/support/**/*.rb').sort_by(&:to_s).each { |f| require f }"}$/
      )
```

DiegoPisa2003: Oh yeah you are right, thanks. I will apply your suggestion

PR #2869, opened 2025-10-27
Avoid loading active record fixtures

moofkit:
Even when `use_active_record = false` is set, rspec-rails still included `ActiveRecord::TestFixtures` and intercepted `before` and `after` hooks, attempting to access the databases associated with active record models. This caused issues in non-standard Rails projects where ActiveRecord is present but not using the Rails testing framework directly.

This PR allows projects with ActiveRecord present to cleanly opt-out of fixture-based testing without database hook interference.

### Changes
- moved `ActiveRecord::TestFixtures` inclusion inside the `use_active_record?` conditional check
- created `ActiveRecord.md` guide covering fixtures, transactions, and configuration options. Merged `Transactions.md` to the new document
- updated `rails_helper.rb` comments to be aligned with updated documentation.


example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
```diff
@@ -1,21 +1,6 @@
...
-      module TestFixtures
-      end
-    end
-  end
-end
-
 require 'rails_helper'
 
-RSpec.describe 'Example App', :use_fixtures, type: :model do
+RSpec.describe 'Example App', type: :model do
```
pirj, line 3: It you revert this spec change, will the old spec still work?

If it won’t, why?

example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
```diff
@@ -1,21 +1,6 @@
...
   it "does not set up fixtures" do
```
pirj, line 4: This means “we don’t set up fixtures even if AR is loaded and there’s a metadata tag”, right? I hat changed?

lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -45,12 +45,11 @@
     Rails.root.join('spec/fixtures')
   ]
 
-  # If you're not using ActiveRecord, or you'd prefer not to run each of your
-  # examples within a transaction, remove the following line or assign false
-  # instead of true.
+  # If you'd prefer not to run each of your examples within a transaction,
```
pirj, line 48: Why remove the mention of AR here?

example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
moofkit: I'm not sure that using `:use_fixtures` changes anything at all.
Please look at this code:
https://github.com/rspec/rspec-rails/blob/ef12dbcdae4f209027aa311d3dcc88fe131c2a46/lib/rspec/rails/configuration.rb#L73-L82

It includes `RSpec::Rails::FixtureSupport` unconditionally, even if `:use_fixtures` is not provided.
So the purpose of this meta-tag is a little vague.

Thanks for the suggestion. I'll revert this change.

lib/generators/rspec/install/templates/spec/rails_helper.rb
moofkit: Well, originally this statement is not quite correct, specifically the `If you're not using ActiveRecord` part.
The `use_transactional_fixtures` setting doesn't make any sense without the ActiveRecord, as it's a part of its configuration in specs. But you are right, it needs to mention the setting is related to the ActiveRecord.
Thanks

example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
pirj: The global config.include with metadata is a conditional include as far as I recall

lib/generators/rspec/install/templates/spec/rails_helper.rb
```diff
@@ -45,12 +45,11 @@
...
+  # remove the following line or assign false instead of true.
   config.use_transactional_fixtures = true
 
-  # You can uncomment this line to turn off ActiveRecord support entirely.
+  # You can uncomment this line to turn off ActiveRecord support entirely
```
JonRowe, line 52: Revert this change please.

lib/generators/rspec/install/templates/spec/rails_helper.rb
JonRowe: I'd prefer it if this was reverted, you should delete this setting if you're not using ActiveRecord, thats why it's written that way.

features/ActiveRecord.md
```diff
@@ -0,0 +1,181 @@
+# Active Record
```
JonRowe, line 1: I'd prefer it if you reverted this change, if you want to add to the documentation in another PR or in the original file thats fine, but this is really hard to review what if anything you've changed

example_app_generator/no_active_record/spec/verify_no_fixture_setup_spec.rb
```diff
@@ -1,21 +1,6 @@
...
-# so that it's reachable from RSpec::Rails::FixtureSupport.
-# NOTE: this has to be defined before requiring `rails_helper`.
-module RSpec
-  module Rails
-    module ActiveRecord
-      module TestFixtures
-      end
-    end
-  end
-end
```
JonRowe, line 14: This needs to remain, see the documentation comment, it exists to test that stuff doesn't happen when the flag is set to false and the module is defined, this is because even though we can exclude active record, it often still exists due to dependencies, but hasn't been configured...

PR #2870, opened 2025-10-28
Fix ActiveJob matchers to use job-specific queue adapters

lparry:
In an app with multiple different queue backends, each may be configured with it's own distinct queue adapter

eg.

```ruby
ActiveJob::Base.queue_adapter = :test
QueJob.queue_adapter = :custom_adapter
SidekiqJob.queue_adapter = :test
```

ActiveJob matchers were always checking for the exact instance `ActiveJob::Base.queue_adapter`, even when testing job subclasses that defined their own queue_adapter. This caused matchers to check the wrong test adapter for queued/performed jobs, and meant that the "test adapter is configured" check was not actually checking the correct bit of configuration.

I have changed queue_adapter method to check @job.queue_adapter when a specific job class is being tested, falling back to ActiveJob::Base's queue_adapter only when @job is nil, and made `have_enqueued_job` accept an options `job` arg to try and check the correct spot more often.

This fixes issues when testing multiple job classes with potentially different queue adapters.

### DIFF
--- a/lib/rspec/rails/matchers/active_job.rb
+++ b/lib/rspec/rails/matchers/active_job.rb
@@ -273,7 +273,7 @@ module RSpec
           end
 
           def queue_adapter
-            ::ActiveJob::Base.queue_adapter
+            @job ? @job.queue_adapter : ::ActiveJob::Base.queue_adapter
           end
         end
         # rubocop: enable Metrics/ClassLength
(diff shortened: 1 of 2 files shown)


lparry:
Found this only sorted out a portion of our newly failing tests

Issue #2876, opened 2025-12-20
Add support for Rails 8.1 Structured Event Reporter


keshavbiswa:
**Is your feature request related to a problem? Please describe.**
With the introduction of Structured Event Reporting, we should have matchers to support it just like what we have for rails: https://github.com/keshavbiswa/rails/blob/main/activesupport/test/testing/event_reporter_assertions_test.rb

**Describe the solution you'd like**
Matchers for HaveReportedEvent.

**Describe alternatives you've considered**

**Additional context**

PR #2877, opened 2025-12-20
Add RSpec matchers for Rails 8.1+ EventReporter

keshavbiswa:
Add RSpec matchers for Rails 8.1+ EventReporter

Introduces matchers for testing ActiveSupport::EventReporter:
- have_reported_event: Assert an event was reported
- with_payload: Match event payload
- with_contexts: Match event contexts
- with_tags: Match event tags
- have_reported_no_event: Assert no matching events
- have_reported_events: Assert multiple events
- with_debug_event_reporting: Helper to enable debug mode in tests

Related issue: #2876

### DIFF
--- a/lib/rspec/rails/matchers/event_reporter.rb
+++ b/lib/rspec/rails/matchers/event_reporter.rb
@@ -8,59 +8,6 @@ module RSpec
       # @api private
       module EventReporter
         # @api private
-        # Internal subscriber that collects events during test execution.
-        module EventCollector
-          @subscribed = false
-          @mutex = Mutex.new
-
-          class << self
-            # @api private
-            # Receives events from the Rails EventReporter subscriber.
-            def emit(event)
-              event_recorders&.each do |recorder|
-                recorder << Event.new(event)
-              end
-              true
-            end
-
-            # @api private
-            # Records events emitted during the block execution.
-            def record
-              subscribe
-              events = []
-              event_recorders << events
-              begin
-                yield
-                events
-              ensure
-                event_recorders.delete_if { |r| events.equal?(r) }
-              end
-            end
-
-            private
-
-            def subscribe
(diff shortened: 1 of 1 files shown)


lib/rspec/rails/matchers/event_reporter.rb
```diff
@@ -0,0 +1,558 @@
...
+        #
+        # @see RSpec::Rails::Matchers#have_reported_event
+        class HaveReportedEvent < Base
+          def initialize(expected_name)
+            super()
+            @expected_name = expected_name
+          end
+
+          def matches?(block)
+            @events = EventCollector.record(&block)
```
moofkit: I believe the matcher could be simplified by utilizing `ActiveSupport::Testing::EventReporterAssertions::EventCollector.record(&block)` instead of implementing `EventCollector` from scratch.

keshavbiswa: Ahhh good point. Didn't think of it earlier, would have saved a lot of time.

keshavbiswa: @moofkit Removed it.

lib/rspec/rails/matchers/event_reporter.rb
```diff
@@ -0,0 +1,510 @@
...
+            @event_data = event_data
+          end
+
+          # @api private
+          # Returns a human-readable representation of the event.
+          def inspect
+            "#{event_data[:name]} (payload: #{event_data[:payload].inspect}, tags: #{event_data[:tags].inspect}, context: #{event_data[:context].inspect})"
+          end
+
+          def matches?(name, payload = nil, tags = nil, context = nil)
```
JonRowe: I feel like these would be better as keyword attributes

Issue #2878, opened 2026-02-05
Add option to rspec:install generator to configure for Selenium in Docker


unikitty37:
**Is your feature request related to a problem? Please describe.**
When I run `rails new` and don't skip tests, system tests, or Docker, it creates a Docker container for Selenium and installs this in `test/application_system_test_case.rb`:

```ruby
require "test_helper"

class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
  if ENV["CAPYBARA_SERVER_PORT"]
    served_by host: "rails-app", port: ENV["CAPYBARA_SERVER_PORT"]

    driven_by :selenium, using: :headless_chrome, screen_size: [ 1400, 1400 ], options: {
      browser: :remote,
      url: "http://#{ENV["SELENIUM_HOST"]}:4444"
    }
  else
    driven_by :selenium, using: :headless_chrome, screen_size: [ 1400, 1400 ]
  end
end
```

`rails g rspec:install` doesn't do anything to set Selenium up, and getting system tests working is a nightmare as a result.

**Describe the solution you'd like**
It should be possible to pass the container name to the install generator, using syntax along the lines of `rails g rspec:install --selenium-container=selenium` and the generator should then provide a basic setup similar to the one `rails new` does for minitest.

**Describe alternatives you've considered**
Right now I'm considering just using Playwright for system tests, because I know I can get that working.

**Additional context**

JonRowe:
:wave: A PR would be accepted; I don't use Docker in my workflow and don't have the bandwidth to investigate this myself.

PR #2886, opened 2026-02-24
Add option to rspec:install generator to configure for Selenium in Docker #2878


nicolasva:
issue : #2878

**changes :** 
4 files modified/created:

1. [install_generator.rb]
class_option :selenium_container (string, nil by default)
Public method copy_system_test_configuration that generates the support file when the option is provided
Private helper selenium_container
2. [system_test_configuration.rb]— ERB template that generates an RSpec.configure block with:
served_by + remote driven_by when CAPYBARA_SERVER_PORT is set (Docker mode)
Fallback to driven_by :selenium, using: :headless_chrome locally
3. [install_generator_spec.rb] — 7 new tests covering:
File generation when option is provided
Container hostname, served_by, remote driven_by, local fallback
Custom hostname (e.g. --selenium-container=chrome)
No file generated without the option
4. [Changelog.md] — New entry under the Development section

**Usage:**
rails generate rspec:install --selenium-container=selenium
Then uncomment the spec/support/**/*.rb glob in rails_helper.rb to auto-load the configuration.

### DIFF
--- a/spec/generators/rspec/install/install_generator_spec.rb
+++ b/spec/generators/rspec/install/install_generator_spec.rb
@@ -1,4 +1,5 @@
 # Generators are not automatically loaded by Rails
+require 'erb'
 require 'generators/rspec/install/install_generator'
 require 'support/generators'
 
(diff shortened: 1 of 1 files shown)


spec/generators/rspec/install/install_generator_spec.rb
```diff
@@ -107,6 +107,47 @@ def filter_rails_from_backtrace
...
+      expect(system_test_config).to match(/SELENIUM_HOST.*chrome/)
+    end
+  end
+
+  context "without --selenium-container option" do
+    it "does not generate spec/support/system_test_configuration.rb" do
+      run_generator
+      expect(File.exist?(file('spec/support/system_test_configuration.rb'))).to be false
+    end
+  end
```
JonRowe, line 140: I think I'd rather see one big expected config block with the right values interpolated, and/or a check against the Rails generator output itself to make sure we stay in sync...

nicolasva: Here's the summary of the changes made to address JonRowe's feedback:

Problem: The original specs had 6 separate it blocks, each calling run_generator independently and checking a single thing via regex. This was inefficient and fragile — if the template changed, only specific regex patterns would catch it.

Changes made ([install_generator_spec.rb]

Single generator run — replaced repeated run_generator calls with a before hook, so the generator runs once per context.

Full content comparison against the actual template — replaced 5 individual regex checks with one it block that reads the real ERB template from lib/generators/rspec/install/templates/, renders it with result_with_hash(selenium_container: container_name), and compares the full output with eq. This ensures the test stays in sync with the template automatically — if the template changes, the test will catch any discrepancy.

Custom container name — kept as a nested context with an overridden let(:container_name), reusing the same before hook.

Added require 'erb' at the top of the spec file.

@JonRowe

Issue #2887, opened 2026-02-25
Add Turbo/Hotwire testing support with matchers for turbo streams and frames


bastin27:
**Is your feature request related to a problem? Please describe.**

Rails 7+ uses Turbo as the default for form submissions and page updates via turbo-rails. Currently, rspec-rails has no built-in support for testing Turbo Stream and Turbo Frame responses, forcing users to write manual assertions against raw HTML or content types.

**Describe the solution you'd like**

Matchers for Turbo Stream and Turbo Frame responses:

- _have_turbo_stream(action:, target:)_ - asserts a <turbo-stream> element exists with the given action and target/targets, with optional .with_count(n) chaining
- _be_turbo_stream_ - asserts the response has Turbo Stream content type (text/vnd.turbo-stream.html)
- _have_turbo_frame(id)_ - asserts a <turbo-frame> element exists with the given id
Similar to what Rails provides in its own test helpers: https://github.com/hotwired/turbo-rails/blob/main/lib/turbo/test_assertions.rb

Additionally, the scaffold generator should be updated to include turbo_stream format tests for create, update, and destroy actions when Turbo is available.

**Describe alternatives you've considered**

- Using _assert_turbo_stream_response_ from turbo-rails directly, but this is Minitest-only and not available in RSpec
- Manually parsing response.body with Nokogiri in every test
- Third-party gems, but this is core enough to belong in rspec-rails

**Additional context**

Reference implementation for Rails structured event reporting matchers follows a similar pattern: https://github.com/keshavbiswa/rails/blob/main/activesupport/test/testing/event_reporter_assertions_test.rb

PR #2888, opened 2026-02-25
Add RSpec matchers for turbo stream and turbo frames

bastin27:
## Summary

 - Add have_turbo_stream matcher to assert <turbo-stream> elements by action and target/targets, with optional .with_count(n) chaining                                                             
  - Add be_turbo_stream matcher to verify Turbo Stream content type (text/vnd.turbo-stream.html)                                                                                                    
  - Add have_turbo_frame matcher to assert <turbo-frame> elements by id                                                                                                                             
  - Add has_turbo? feature check for conditional loading when turbo-rails is available                                                                                                              
  - Update scaffold generator to include turbo_stream format tests for create, update, and destroy actions                                                                                          
  - Matchers leverage Rails assertions via match_unless_raises with assert_select, consistent with redirect_to and have_rendered patterns                                                           
  - Includes Cucumber feature specs and RSpec unit tests (31 examples)                                                                                                                              
                                                                                                                                                                                                    
  Test plan                                                                                                                                                                                         
  - [X] All 31 turbo matcher specs pass                                                                                                                                                                 
  - [X] Rubocop passes with no offenses                                                                                                                                                                 
  - [X] Verify Cucumber feature specs pass with turbo-rails available                                                                                                                                   
  - [X] Test scaffold generator output includes turbo_stream format tests                                                                                                                               
  - [X] Confirm matchers work in a real Rails 7+ app with Turbo enabled     

Related issue: https://github.com/rspec/rspec-rails/issues/2887

### DIFF
--- a/lib/generators/rspec/scaffold/templates/request_spec.rb
+++ b/lib/generators/rspec/scaffold/templates/request_spec.rb
@@ -89,12 +89,14 @@ RSpec.describe "/<%= name.underscore.pluralize %>", <%= type_metatag(:request) %
       end
     end
 
-    context "with turbo_stream format", if: defined?(Turbo) do
-      it "creates a new <%= class_name %> and responds with turbo stream" do
-        post <%= index_helper %>_url, params: { <%= singular_table_name %>: valid_attributes }, as: :turbo_stream
-        expect(response).to be_turbo_stream
+    <% if RSpec::Rails::FeatureCheck.has_turbo? -%>
+      context "with turbo_stream format", if: defined?(Turbo) do
+        it "creates a new <%= class_name %> and responds with turbo stream" do
+          post <%= index_helper %>_url, params: { <%= singular_table_name %>: valid_attributes }, as: :turbo_stream
+          expect(response).to be_turbo_stream
+        end
       end
-    end
+    <% end -%>
   end
 
   describe "PATCH /update" do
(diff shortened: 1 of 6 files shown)


spec/rspec/rails/matchers/turbo/have_turbo_stream_spec.rb
```diff
@@ -0,0 +1,182 @@
+require "rspec/rails/feature_check"
+require "rspec/rails/matchers/turbo/have_turbo_stream"
+
+RSpec.describe "have_turbo_stream matcher" do
+  include RSpec::Rails::Matchers::Turbo
+
+  def have_turbo_stream(**opts)
+    RSpec::Rails::Matchers::Turbo::HaveTurboStream.new(**opts)
+  end
```
JonRowe: As above, you should be able to call this already

spec/rspec/rails/matchers/turbo/have_turbo_frame_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "rspec/rails/matchers/turbo/have_turbo_frame"
+
+RSpec.describe "have_turbo_frame matcher" do
+  def have_turbo_frame(id)
+    RSpec::Rails::Matchers::Turbo::HaveTurboFrame.new(id)
+  end
```
JonRowe: This should be available already

lib/generators/rspec/scaffold/templates/request_spec.rb
```diff
@@ -88,6 +88,13 @@
         expect(response).to have_http_status(<%= Rack::Utils::SYMBOL_TO_STATUS_CODE.key(422).inspect %>)
       end
     end
+
+    context "with turbo_stream format", if: defined?(Turbo) do
+      it "creates a new <%= class_name %> and responds with turbo stream" do
+        post <%= index_helper %>_url, params: { <%= singular_table_name %>: valid_attributes }, as: :turbo_stream
+        expect(response).to be_turbo_stream
+      end
+    end
```
JonRowe: These should be wrapped in an if statement that only emits the fragment if Turbo is enabled e.g. (note the `-%>` which remove the if line whitespace from the output)

```suggestion
<% if RSpec::Rails::FeatureCheck.has_turbo? -%>

    context "with turbo_stream format" do
      it "creates a new <%= class_name %> and responds with turbo stream" do
        post <%= index_helper %>_url, params: { <%= singular_table_name %>: valid_attributes }, as: :turbo_stream
        expect(response).to be_turbo_stream
      end
    end
<% end -%>
```

PR #2901, opened 2026-05-22
Clear ActiveJob queues between examples in job specs

### LINKED ISSUE #2900: Clear enqueued jobs after each example to match Mailer behavior

I found it confusing why rspec-rails clears enqueued mailers by default, but does _not_ clear enqueued jobs automatically (between tests).

It's simple enough to register our own "clear" between examples, but it took a lot of debugging before finding that it was the queued job state between tests that was causing our assertions to fail.

I would like enqueued active jobs (via the test adapter), to be cleared after each test just like it already does for mailers: https://github.com/rspec/rspec-rails/pull/2293`

I do see value in allowing the jobs to remain enqueued (it allows `after` hooks to do additional checks or to perform any enqueued jobs). But the disparity with Mailers is the pain point. Mailers themselves are implemented as jobs, and the async behavior for enqueuing, and asserting on them is consistent with Jobs. It seems counter-intuitive for the clearing behavior to be different between them.

**Additional context**
Should this be implemented, a question I would have is whether the clearing should occur in the _after_ hook or _before_ hook. It seems to me that clearing in `before` leaves the most flexibility. If folks need to inspect the job queue in an after hook, that would be easy to do. However, if we do the cleanup in an after hook, then users would need to ensure any inspection is done _before_ our after hook fires. (Which would prevent inspecting in an `around` hook.) If we agree that `before` is preferable, should the Mail delivery queue also be cleared in `before` for the same reason? (Instead of `after`)

I would be happy to open the PR if this feature would be well-received.

hammadxcm:
Fixes #2900.

## Problem

`have_been_enqueued` and `have_been_performed` inspect the queue adapter's accumulated `enqueued_jobs` / `performed_jobs` arrays directly, without state-delta logic. When the queue adapter is configured once (the usual setup in `config/environments/test.rb`), the same `TestAdapter` instance is reused across the suite and jobs accumulate between examples. The matchers then see jobs left over from prior tests.

rspec-rails' own matcher suite doesn't expose this because it reassigns `ActiveJob::Base.queue_adapter = :test` in a `before` hook (`spec/rspec/rails/matchers/active_job_spec.rb:88`), which creates a fresh `TestAdapter` per example — masking the production behaviour. Issue #2900 has a clear writeup by @jasonkarns showing the leak in a real Rails app.

## Fix

Add a `before` hook in `RSpec::Rails::JobExampleGroup` that clears `enqueued_jobs` and `performed_jobs` on the queue adapter at the start of every example, mirroring what `ActiveJob::TestHelper#before_setup` does for `ActiveJob::TestCase`. Scoping to `JobExampleGroup` (rather than `RailsExampleGroup`) keeps the change minimal and matches how Rails layers the helper — only `ActiveJob::TestCase` gets the clearing behaviour.

The `respond_to?` guards are there in case a user has swapped in a non-test queue adapter.

## Tests

New spec `spec/rspec/rails/example/job_example_group_spec.rb` uses the same `order: :defined` sub-example-group pattern as the existing `CurrentAttributes` leak test in `rails_example_group_spec.rb`. It runs two examples in sequence — the first enqueues / performs a job, the second asserts both queues are empty — and fails without the fix.

## Test plan

- [x] `bundle exec rspec spec/rspec/rails/example/job_example_group_spec.rb` — 16/16 pass
- [x] `bundle exec rspec spec/rspec/rails/example/` — 280/280 pass
- [x] `bundle exec rspec spec/rspec/rails/matchers/active_job_spec.rb` — no regressions
- [x] `bundle exec rubocop` on changed files — clean

### DIFF
--- a/lib/rspec/rails/example/rails_example_group.rb
+++ b/lib/rspec/rails/example/rails_example_group.rb
@@ -16,9 +16,12 @@ module RSpec
       include RSpec::Rails::MinitestAssertionAdapter
       include RSpec::Rails::FixtureSupport
       include RSpec::Rails::TaggedLoggingAdapter
-      include ActiveJob::TestHelper if RSpec::Rails::FeatureCheck.has_active_job?
       include ActiveSupport::CurrentAttributes::TestHelper
       include ActiveSupport::ExecutionContext::TestHelper
+
+      if RSpec::Rails::FeatureCheck.has_active_job?
+        include ActiveJob::TestHelper
+      end
     end
   end
 end
--- a/lib/rspec/rails/example/rails_example_group.rb
+++ b/lib/rspec/rails/example/rails_example_group.rb
@@ -16,6 +16,7 @@ module RSpec
       include RSpec::Rails::MinitestAssertionAdapter
       include RSpec::Rails::FixtureSupport
       include RSpec::Rails::TaggedLoggingAdapter
+      include ActiveJob::TestHelper if RSpec::Rails::FeatureCheck.has_active_job?
       include ActiveSupport::CurrentAttributes::TestHelper
       include ActiveSupport::ExecutionContext::TestHelper
     end
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -59,7 +59,9 @@ module RSpec::Rails
 
     it 'will not leak enqueued ActiveJob jobs between examples', skip: !RSpec::Rails::FeatureCheck.has_active_job? do
       original_adapter = ActiveJob::Base.queue_adapter
+      original_logger = ActiveJob::Base.logger
       ActiveJob::Base.queue_adapter = :test
+      ActiveJob::Base.logger = Logger.new(nil)
 
       leaky_job = Class.new(ActiveJob::Base) do
         def perform; end
@@ -88,6 +90,7 @@ module RSpec::Rails
       ).to be true
     ensure
       ActiveJob::Base.queue_adapter = original_adapter
+      ActiveJob::Base.logger = original_logger
     end
   end
 end
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -56,5 +56,39 @@ module RSpec::Rails
         group.run(failure_reporter) ? true : failure_reporter.exceptions
       ).to be true
     end
+
+    if RSpec::Rails::FeatureCheck.has_active_job?
+      it 'will not leak enqueued ActiveJob jobs between examples' do
+        original_adapter = ActiveJob::Base.queue_adapter
+        ActiveJob::Base.queue_adapter = :test
+        leaky_job = Class.new(ActiveJob::Base) do
+          def perform; end
+
+          def self.name
+            'LeakyJob'
+          end
+        end
+
+        group =
+          RSpec::Core::ExampleGroup.describe("A group", order: :defined) do
+            include RSpec::Rails::RailsExampleGroup
+
+            it 'enqueues a job' do
+              leaky_job.perform_later
+              expect(enqueued_jobs.size).to eq(1)
+            end
+
+            it 'does not see the job enqueued in the prior example' do
+              expect(enqueued_jobs).to be_empty
+            end
+          end
+
+        expect(
+          group.run(failure_reporter) ? true : failure_reporter.exceptions
+        ).to be true
+      ensure
+        ActiveJob::Base.queue_adapter = original_adapter
+      end
+    end
   end
 end
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -57,38 +57,37 @@ module RSpec::Rails
       ).to be true
     end
 
-    if RSpec::Rails::FeatureCheck.has_active_job?
-      it 'will not leak enqueued ActiveJob jobs between examples' do
-        original_adapter = ActiveJob::Base.queue_adapter
-        ActiveJob::Base.queue_adapter = :test
-        leaky_job = Class.new(ActiveJob::Base) do
-          def perform; end
+    it 'will not leak enqueued ActiveJob jobs between examples', skip: !RSpec::Rails::FeatureCheck.has_active_job? do
+      original_adapter = ActiveJob::Base.queue_adapter
+      ActiveJob::Base.queue_adapter = :test
 
-          def self.name
-            'LeakyJob'
+      leaky_job = Class.new(ActiveJob::Base) do
+        def perform; end
+
+        def self.name
+          'LeakyJob'
+        end
+      end
+
+      group =
+        RSpec::Core::ExampleGroup.describe("A group", order: :defined) do
+          include RSpec::Rails::RailsExampleGroup
+
+          it 'enqueues a job' do
+            leaky_job.perform_later
+            expect(enqueued_jobs.size).to eq(1)
+          end
+
+          it 'does not see the job enqueued in the prior example' do
+            expect(enqueued_jobs).to be_empty
           end
         end
 
-        group =
-          RSpec::Core::ExampleGroup.describe("A group", order: :defined) do
-            include RSpec::Rails::RailsExampleGroup
-
-            it 'enqueues a job' do
-              leaky_job.perform_later
-              expect(enqueued_jobs.size).to eq(1)
-            end
-
-            it 'does not see the job enqueued in the prior example' do
-              expect(enqueued_jobs).to be_empty
-            end
-          end
-
-        expect(
-          group.run(failure_reporter) ? true : failure_reporter.exceptions
-        ).to be true
-      ensure
-        ActiveJob::Base.queue_adapter = original_adapter
-      end
+      expect(
+        group.run(failure_reporter) ? true : failure_reporter.exceptions
+      ).to be true
+    ensure
+      ActiveJob::Base.queue_adapter = original_adapter
     end
   end
 end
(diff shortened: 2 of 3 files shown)


pirj:
Hi
This will only affect specs with the `:job` tag, won’t it?

hammadxcm:
@pirj  Good catch — you're right, and that was too narrow. `have_been_enqueued`/`have_been_performed` are included globally via `config.include RSpec::Rails::Matchers`, so the leak isn't specific to `:job` specs; it shows up in any spec type that uses those matchers against a shared test adapter.

I've pushed a follow-up commit that moves the clearing out of `JobExampleGroup` and into a global `before` hook in `initialize_configuration`, right next to the existing `config.after { ActionMailer::Base.deliveries.clear }`:

```ruby
  config.include RSpec::Rails::JobExampleGroup, type: :job
  config.before do
    queue_adapter = ::ActiveJob::Base.queue_adapter
    queue_adapter.enqueued_jobs.clear  if queue_adapter.respond_to?(:enqueued_jobs)
    queue_adapter.performed_jobs.clear if queue_adapter.respond_to?(:performed_jobs)
  end
end
```

The `respond_to?` guards keep it a no-op for non-test adapters (`:inline`, `:async`, etc.), so only the test adapter's accumulating queues get reset. The spec now lives in `configuration_spec.rb` alongside the mailer-deliveries test and verifies isolation across two examples (it fails without the hook).

pirj:
What if we included the `ActiveJob::TestHelper#before_setup` instead of repeating what it does? We do this for a few other things already.

hammadxcm:
That's much cleaner — done. I've replaced the hand-rolled hook with `include ActiveJob::TestHelper` in `RailsExampleGroup`, right next to the existing `ActiveSupport::CurrentAttributes::TestHelper` and `ExecutionContext::TestHelper` includes:

```ruby
if defined?(ActiveJob)
  require 'active_job/test_helper'
  include ActiveJob::TestHelper
end
```

`MinitestLifecycleAdapter` already invokes `before_setup` via its `around` hook, so `TestHelper#before_setup` now resets the test adapter before each example — same mechanism `ActiveJob::TestCase` uses. Bonus: it also handles per-class adapter overrides via `enable_test_adapter`, which the hand-rolled version didn't, and its `clear_enqueued_jobs`/`clear_performed_jobs` are already guarded by `using_test_adapter?`, so non-test adapters are untouched.

The fix is now 4 lines in `rails_example_group.rb`. The spec moved to `rails_example_group_spec.rb` mirroring the `ActiveSupport::CurrentAttributes` leak test (two ordered examples; fails without the include). I squashed the branch to a single commit for a clean diff.

pirj:
Nice!
Do you think now our own job matchers’ specs don’t manually reset those queues?

sorry, misclicked

pirj:
This is a bugfix, but it may cause failures in consumers’ suites. @JonRowe

hammadxcm:
Thanks both! A few quick things:

**Our own matcher specs** — they don't go through `RailsExampleGroup`, so this change doesn't touch them. They already reset themselves with the `before { ActiveJob::Base.queue_adapter = :test }` at the top of the file. I checked, and the manual `enqueued_jobs.clear` lines are actually redundant even today — happy to tidy those up, though probably better as its own PR.

**Consumer impact** — fair point. Pulling in the whole `ActiveJob::TestHelper` adds a fair few methods (`enqueued_jobs`, `performed_jobs`, `queue_adapter`, `perform_enqueued_jobs`, the `assert_*` helpers…) to every rails example group, so there's some risk of clashing with people's own `let`s/helpers, and queues now reset between examples everywhere. Our full suite stays green, but that won't catch downstream surprises. If you'd rather play it safe, I'm glad to put it behind an opt-in setting or land it as a documented change in a minor — whatever you prefer.

**The red CI cell** — Ruby 3.1 / Rails 7.2 failed on a scaffold generator spec: `Thor::UndefinedCommandError: Could not find command "mountable_engine?"`. I don't think it's this change — the test stubs a *private* method on `NamedBase` and it leaks as a Thor command on 3.1, which is a mocks/Thor quirk in code this PR doesn't touch. Every other cell is green (including Rails main) and it passes locally on 3.2–3.4. Is that 3.1 cell a known flake, or should I dig in?

pirj:
> Consumer impact

It’s fine as is. Just a matter of release management.

> the manual enqueued_jobs.clear lines are actually redundant even today — happy to tidy those up, though probably better as its own PR.

Interesting. Is this because expectations don’t overlap between examples?


> The red CI cell

restarted. Might have been a temporary glitch.

jasonkarns:
> Interesting. Is this because expectations don’t overlap between examples?

It's because assigning the test adapter (https://github.com/rspec/rspec-rails/blob/main/spec/rspec/rails/matchers/active_job_spec.rb#L88-L90) instantiates a fresh adapter before each test so the queue _starts_ empty. (If you set a breakpoint in that before hook, you can see that the queue isn't empty prior to re-assigning/reseting the adapter.)


In my repro example here https://github.com/rspec/rspec-rails/compare/main...jasonkarns:rspec-rails:active-job-queue-clearing I modified the hook to only set the adapter once. I'll submit that as a PR on its own. Presently, the manual job clearing is redundant, _but explanatory_. It's not clear that the adapter assignent is doing the "clearing" as a side effect. Instead, I think the tests would be more robust if the adapter were set in a `before :all` hook like in my repro spike, which would make the explicit clearing non-redundant. (But still clear, precise, and exemplary to user/readers.)

lib/rspec/rails/example/rails_example_group.rb
```diff
@@ -16,6 +16,10 @@ module RailsExampleGroup
       include RSpec::Rails::MinitestAssertionAdapter
       include RSpec::Rails::FixtureSupport
       include RSpec::Rails::TaggedLoggingAdapter
+      if defined?(ActiveJob)
+        require 'active_job/test_helper'
```
jasonkarns: Is this require supposed to be necessary? The check for activejob should be sufficient, right? That's when we know the gem is available. And rails autoloading takes over from there.

I know the other includes here are from ActiveSupport and are truly required for rspec-rails to function, so it makes sense for them to be unconditional.

spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -56,5 +56,38 @@ class CurrentSample < ActiveSupport::CurrentAttributes
         group.run(failure_reporter) ? true : failure_reporter.exceptions
       ).to be true
     end
+
+    if defined?(ActiveJob)
```
jasonkarns: rspec rails has a dedicated feature check (`RSpec::Rails::FeatureCheck.has_active_job?`) should we be using that? I think if we use it, we'd be able to restructure this test to pull the LeakyJob class out of the block and eliminate the rubocop exception.

lib/rspec/rails/example/rails_example_group.rb
jasonkarns: Also, should it be using `RSpec::Rails::FeatureCheck.has_active_job?` (is that helper even available at this point in the boot phase?

jasonkarns (commented):
Nice

pirj:
> instantiates a fresh adapter before each test

>> the manual enqueued_jobs.clear lines are actually redundant even today

I was asking in this context. I thought this means that this re-assignment to a fresh queue before each test is no longer needed, which would be just perfect.

Can we set the test queue adapter just once in a `before(:all)`, would specs remain green?

hammadxcm:
CI's all green now after the restart — looks like the 3.1 cell was just a glitch, thanks for re-running it.

On the `before(:all)` idea: the per-example `queue_adapter = :test` is what's currently doing the reset — assigning the adapter hands you a fresh instance with empty queues each time, which is why the explicit `clear`s read as redundant today. If we set it once in a `before(:all)` instead, those `clear`s stop being redundant and start doing the real work (exactly @jasonkarns' point). I gave it a quick try, and even with the clears kept, a few of the `have_been_enqueued` signature-check specs went red — so it needs a bit more care than a straight swap.

Since @jasonkarns is already picking that up as its own PR, I'd rather keep this one focused on the fix and let his handle the matcher-spec tidy-up. Happy to help review it.

lib/rspec/rails/example/rails_example_group.rb
hammadxcm: Good call — it's not necessary. `ActiveJob` autoloads `:TestHelper` (active_job.rb), so referencing `ActiveJob::TestHelper` under the `has_active_job?` guard is enough. Dropped the `require`.

hammadxcm: Yes — and it is available here: this file requires `rspec/rails/matchers` at the top, which pulls in `rspec/rails/feature_check`. Switched to `RSpec::Rails::FeatureCheck.has_active_job?` to match `configuration.rb`.

spec/rspec/rails/example/rails_example_group_spec.rb
hammadxcm: Done — switched the guard to `has_active_job?` and replaced the in-block `class LeakyJob` with an anonymous `Class.new(ActiveJob::Base)` (with a `self.name`, matching the job-class style in active_job_spec.rb). That drops the `Lint/ConstantDefinitionInBlock` exception entirely.

hammadxcm:
Thanks for the review @jasonkarns — all three addressed and pushed:

- Dropped the `require 'active_job/test_helper'` (it's autoloaded).
- Switched both the include and the spec guard to `RSpec::Rails::FeatureCheck.has_active_job?` (available here via the `rspec/rails/matchers` require at the top).
- Rewrote the spec's `LeakyJob` as an anonymous `Class.new(ActiveJob::Base)`, which lets the `Lint/ConstantDefinitionInBlock` exception go away.

The lib change is now a single line. Full suite stays green.

JonRowe:
Thanks all!

PR #2907, opened 2026-08-17
Assign default_url_options instead of mutating it

### LINKED ISSUE #2906: Rails main freezes default_url_options, breaking feature specs
## Reproduction

With Rails `main` after [`rails/rails@baa1ca5f2f75477e4ce3e9f6237d1dc29c9596a7`](https://github.com/rails/rails/commit/baa1ca5f2f75477e4ce3e9f6237d1dc29c9596a7) and rspec-rails 8.0.4, loading a feature spec raises:

```text
FrozenError: can't modify frozen Hash: {}
```

The error is caused by [`feature_example_group.rb:19`](https://github.com/rspec/rspec-rails/blob/main/lib/rspec/rails/example/feature_example_group.rb#L19):

https://github.com/rspec/rspec-rails/blob/1dbe95df0b2fe0e9089e762e0b2804f3d7fb2fb9/lib/rspec/rails/example/feature_example_group.rb#L19

The Rails commit changed the default value:

```diff
- self.default_url_options = {}
+ self.default_url_options = {}.freeze
```

As a result, the mutation in `FeatureExampleGroup` fails while the example group is being defined.

## Expected behavior

Feature specs should load successfully with Rails `main`.

## Possible fix

Assign a new hash instead of mutating the existing value:

```ruby
self.default_url_options = default_url_options.merge(
  host: ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
)
```

Alternatively, Rails could provide a compatible API for this use case.

Ref: https://github.com/rails/rails/pull/58483

_This issue was posted by OpenAI gpt-5.6-luna. The content was reviewed by a human and verified against a real-world scenario failure._

federico-carrocera:
Rails main freezes the controller `default_url_options` (rails/rails#58483),
so the in-place writes in `FeatureExampleGroup` and `MailerExampleGroup` now
raise `FrozenError` while the example group is being defined. We hit this
bumping an app to Rails main; rspec-rails' own suite fails the same way.

```
 RSpec::Rails::FeatureExampleGroup includes Rails route helpers
     Failure/Error: include FeatureExampleGroup

     FrozenError:
       can't modify frozen Hash: {}
     # ./lib/rspec/rails/example/feature_example_group.rb:19:in 'block in <module:FeatureExampleGroup>'
     # .../active_support/concern.rb:150:in 'ActiveSupport::Concern#append_features'
```
Assign through the `class_attribute` writer instead, as the same Rails commit
did when migrating its own suite.

Fixes #2906.

### DIFF
--- a/lib/rspec/rails/example/feature_example_group.rb
+++ b/lib/rspec/rails/example/feature_example_group.rb
@@ -9,6 +9,9 @@ module RSpec
       # Default host to be used in Rails route helpers if none is specified.
       DEFAULT_HOST = "www.example.com"
 
+      # Default url options, merged into those of the example group.
+      DEFAULT_OPTIONS = { host: DEFAULT_HOST }.freeze
+
       included do
         app = ::Rails.application
         if app.respond_to?(:routes)
@@ -16,7 +19,7 @@ module RSpec
           include app.routes.mounted_helpers if app.routes.respond_to?(:mounted_helpers)
 
           if respond_to?(:default_url_options)
-            default_url_options[:host] ||= ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
+            self.default_url_options = DEFAULT_OPTIONS.merge(default_url_options)
           end
         end
       end
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,8 @@ Bug Fixes:
 
 * Include `ActiveJob::TestHelper` in all rails example groups to ensure jobs are reset
   between examples matching Rails in-build behaviour. (Hammad Khan, rspec/rspec-rails#2901)
+* Handle Rails 8.2 having a frozen hash for `default_url_options` by assigning rather than mutating it.
+  (Federico Carrocera, rspec/rspec-rails#2907)
 * Fix `have_stream` (and related) matcher(s) by handling changes to stream test harness in Rails 8.2.
   (Jon Rowe, Phil Pirozhkov, rspec/rspec-rails#2909)
 
--- a/lib/rspec/rails/example/mailer_example_group.rb
+++ b/lib/rspec/rails/example/mailer_example_group.rb
@@ -21,8 +21,8 @@ if defined?(ActionMailer)
 
         included do
           include ::Rails.application.routes.url_helpers
-          options = ::Rails.configuration.action_mailer.default_url_options || {}
-          options.each { |key, value| default_url_options[key] = value }
+          options = ::Rails.configuration.action_mailer.default_url_options
+          self.default_url_options = default_url_options.merge(options) if options.present?
         end
 
         # Class-level DSL for mailer specs.
--- a/spec/rspec/rails/example/feature_example_group_spec.rb
+++ b/spec/rspec/rails/example/feature_example_group_spec.rb
@@ -18,6 +18,38 @@ module RSpec::Rails
       expect(group.new.foo_url).to eq("http://www.example.com/foo")
     end
 
+    it "does not mutate a frozen `default_url_options`" do
+      with_isolated_stderr do
+        Rails.application.routes.draw do
+          get "/foo", as: :foo, to: "foo#bar"
+        end
+      end
+
+      group = RSpec::Core::ExampleGroup.describe do
+        include ::Rails.application.routes.url_helpers
+        self.default_url_options = {}.freeze
+        include FeatureExampleGroup
+      end
+
+      expect(group.new.foo_url).to eq("http://www.example.com/foo")
+    end
+
+    it "does not override a host set by the example group" do
+      with_isolated_stderr do
+        Rails.application.routes.draw do
+          get "/foo", as: :foo, to: "foo#bar"
+        end
+      end
+
+      group = RSpec::Core::ExampleGroup.describe do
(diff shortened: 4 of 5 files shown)


pirj:
A quick assessment of the CI state.

There seem to be a few things broken with dependencies,
1) [Rails main](https://github.com/rspec/rspec-rails/actions/runs/32050293502/job/95455023587#step:7:1620)
```
Failure/Error: next if klass <= ::ActionCable::Connection::Base
      NoMethodError:
        undefined method `<=' for nil
```
2) [Rails 8.0.5.1](https://github.com/rspec/rspec-rails/actions/runs/32050293502/job/95455023701#step:7:4916) ?
```
      Failure/Error: run_generator %w[posts --request_specs]
     Thor::UndefinedCommandError:
       Could not find command "mountable_engine?".
```
3) [Some weird gem issue](https://github.com/rspec/rspec-rails/actions/runs/32050293502/job/95455023755#step:7:4795)
```
Running avoid_fixture_name_collision.rb
/opt/hostedtoolcache/Ruby/3.4.10/x64/lib/ruby/gems/3.4.0/gems/bundler-4.0.18/lib/bundler/runtime.rb:71:in 'block (2 levels) in Bundler::Runtime#require': There was an error while trying to load the gem 'rails'. (Bundler::GemRequireError)
Gem Load Error is: cannot load such file -- concurrent/map
```
This is possibly due to the bundler cache yielding some garbage with git-sourced dependencies.
Fix idea:
```diff
      - uses: ruby/setup-ruby@v1
        with:
          ruby-version: ${{ matrix.ruby }}
+         bundler-cache: true
```
Note: setup-ruby/v1's [docs](https://github.com/ruby/setup-ruby/blob/95ef2b042f9d7a56d8268cba8559e2842e2ad01b/README.md?plain=1#L198) mention that:
> When using `bundler-cache: true` you might notice there is no good place to run `bundle config ...` commands.
> These can be replaced by `BUNDLE_*` environment variables, which are also faster.
But [we do set](https://github.com/rspec/rspec-rails/blob/1dbe95df0b2fe0e9089e762e0b2804f3d7fb2fb9/script/bundle#L8) them:
```
bundle config set --local path '.bundle/gems'
```
And this needs to be fixed as well.
4) [Something with AJ::GlobalId](https://github.com/rspec/rspec-rails/actions/runs/32050293502/job/95455023674#step:7:1680) in Rails main:
```
      ActiveJob::DeserializationError:
        Error while trying to deserialize arguments: Couldn't fetch record for gid://rspec-suite/GlobalIdModel/42: undefined method 'where' for class GlobalIdModel
      # ./.bundle/gems/ruby/3.4.0/gems/globalid-1.4.0/lib/global_id/locator.rb:191:in 'GlobalID::Locator.fetch_record'
```
[and](https://github.com/rspec/rspec-rails/actions/runs/32050293502/job/95455023674#step:7:1306) 
```
        expected: #<GlobalIdModel:0x00007f76ebfe9a48 @id="42", @global_id=#<GlobalID:0x00007f76ebfe6de8 @uri=#<URI::GID gid://rspec-suite/GlobalIdModel/42>>>
             got: {"_aj_globalid" => "gid://rspec-suite/GlobalIdModel/42"}
```
edit: 5) [missing to_i](https://github.com/rspec/rspec-rails/actions/runs/32064070693/job/95491802216?pr=2908#step:7:48)
```
      NoMethodError:
       undefined method `to_i' for an instance of RSpec::Matchers::BuiltIn::Eq
     # ./lib/rspec/rails/matchers/active_job.rb:35:in `at_priority'
```

None of the above seem to be related to the PR in question, but I'd prefer to see a green build without this noise.

lib/rspec/rails/example/feature_example_group.rb
```diff
@@ -16,7 +16,10 @@ module FeatureExampleGroup
           include app.routes.mounted_helpers if app.routes.respond_to?(:mounted_helpers)
 
           if respond_to?(:default_url_options)
-            default_url_options[:host] ||= ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
+            options = default_url_options
+            self.default_url_options = options.merge(
+              host: options[:host] || ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST
+            )
```
JonRowe: Rather than doing this, it would be cleaner to setup our own frozen default hash e.g.
```
DEFAULT_OPTIONS = {host: ::RSpec::Rails::FeatureExampleGroup::DEFAULT_HOST}.freeze
```
The merge into it, this means that its still possible for `options[:host]` to override our own, and negates the need to assign the variable ourselves.

```suggestion
            self.default_url_options = DEFAULT_OPTIONS.merge(options)
```

federico-carrocera: Uhhh yes that is cleaner will ammend the commit now thanks 😄

federico-carrocera: Pushed it there

federico-carrocera:
Thanks for fixing the conflict  @JonRowe  😄

JonRowe:
Thanks for the fix!

tagliala:
Thank you for the fix and the responsiveness, really appreciate

PR #2913, opened 2026-08-21
Handle templates returned directly by resolvers

### LINKED ISSUE #2905: `render_views: false` no longer nullifies templates on Rails main – `ResolverDecorator#nullify_templates` only handles `Enumerable`

Ruby version: 3.4.9
Rails version: main @ `99874313157adb0ef0f85487c5985dcfa88390e8` (post-8.1, pre-8.2)
RSpec version: rspec-rails 8.0.4 (rspec-core/rspec-expectations/rspec-mocks/rspec-support 3.13.x)

## Observed behaviour

Rails commit [rails/rails@9987431315](https://github.com/rails/rails/commit/99874313157adb0ef0f85487c5985dcfa88390e8) (PR ["Resolve the single best template without sorting the whole set"](https://github.com/rails/rails/pull/58279)) changed `ActionView::PathSet#find` to call `resolver.find(...)` directly and return that result, instead of the previous `resolver.find_all(...).first`. Correspondingly, `ActionView::Resolver#find` (new) and `FileSystemResolver#find` now return a single `Template` (or `nil`) rather than an `Array`.

`RSpec::Rails::ViewRendering::EmptyTemplateResolver::ResolverDecorator` implements the default `render_views: false` behavior in controller specs. It undefines all of `Resolver`'s methods and relies on `method_missing` to forward calls to the real resolver, then passes the result through `nullify_templates`:

```ruby
def nullify_templates(collection)
  return collection unless collection.is_a?(Enumerable)
  return collection unless collection.all? { |element| element.is_a?(::ActionView::Template) }

  EmptyTemplateResolver.nullify_template_rendering(collection)
end
```

Before the Rails change, `PathSet` only ever invoked `:find_all` on resolvers, which always returns an `Enumerable`, so `nullify_templates` worked correctly for every call it ever saw. After the Rails change, `PathSet#find` invokes `:find` on each resolver. `method_missing` forwards that call and the real resolver's `find` returns a bare `Template`. Since a bare `Template` is not `Enumerable`, `nullify_templates` returns it unchanged — the real template is used, so it renders for real even with `render_views: false`.

This silently breaks any controller spec that triggers a view render without every helper/instance-variable dependency of that view stubbed. Previously the render was always a no-op, so those dependencies were never exercised; now they raise inside real view code (`NoMethodError`, `ActionView::Template::Error`, etc.), or in the simplest case the view's actual content just leaks into the response body.

## Expected behaviour

`render_views: false` should continue to prevent real template rendering — the resolver decorator should nullify a template regardless of whether it's returned as a bare `Template` (via `find`) or wrapped in an `Enumerable` (via `find_all`).

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"

  git "https://github.com/rails/rails.git", ref: "99874313157adb0ef0f85487c5985dcfa88390e8" do
    gem "actionview"
    gem "actionpack"
    gem "activesupport"
    gem "railties"
  end

  gem "rspec-rails", require: false
  gem "rspec", require: false
end

require "fileutils"
require "tmpdir"

VIEWS_DIR = Dir.mktmpdir
FileUtils.mkdir_p("#{VIEWS_DIR}/test")
File.write("#{VIEWS_DIR}/test/index.html.erb", "<%= raise 'view actually rendered!' %>")

class TestApp < Rails::Application
  config.secret_key_base = "x" * 30
  config.eager_load = false
end
Rails.application.initialize!

class TestController < ActionController::Base
  self.view_paths = [VIEWS_DIR]

  def index
    render "index"
  end
end

TestApp.routes.draw { get "test", to: "test#index" }

require "rspec/rails"

RSpec.describe TestController, type: :controller do
  it "does not render the view when render_views is false (default)" do
    get :index
    expect(response).to have_http_status(:ok)
  end
end

exit RSpec::Core::Runner.run([])
```

This fails with `ActionView::Template::Error: view actually rendered!` — proving the (deliberately exception-raising) view template really executed, despite `render_views` defaulting to `false`. Changing the `ref:` to the parent commit, `a1906d9f81635b8a337cf87c8ea2a8de641ac14f`, makes the same script pass.

## Suggested fix

Update `nullify_templates` to also handle a bare `Template`:

```ruby
def nullify_templates(result)
  if result.is_a?(ActionView::Template)
    EmptyTemplateResolver.nullify_template_rendering([result]).first
  elsif result.is_a?(Enumerable) && result.all?(ActionView::Template)
    EmptyTemplateResolver.nullify_template_rendering(result)
  else
    result
  end
end
```

We've applied this as a local monkeypatch (a prepended module) in our app and confirmed it restores the expected `render_views: false` behavior against Rails main, and it's a no-op on Rails 8.1 — the new `Template` branch is simply never reached there, because `PathSet#find` still resolves via `find_all(...).first` rather than calling `resolver.find` directly, so `nullify_templates` is never called with a bare `Template` in the first place. Happy to open a PR with this change if useful.

saidM:
Rails main changed `ActionView::Resolver#find` to return a single template.

`ResolverDecorator#nullify_templates` only handled collections, so the template was returned unchanged and the real view could render when `render_views` was disabled.

This updates the resolver decorator to handle both single templates and collections.

Fixes #2905

## Testing

- Added a regression spec for a resolver returning a single template
- Ran `bundle exec rspec spec/rspec/rails/view_rendering_spec.rb`
- Ran RuboCop

### DIFF
--- a/lib/rspec/rails/view_rendering.rb
+++ b/lib/rspec/rails/view_rendering.rb
@@ -92,6 +92,10 @@ module RSpec
         private
 
           def nullify_templates(collection)
+            if collection.is_a?(::ActionView::Template)
+              return EmptyTemplateResolver.nullify_template_rendering([collection]).first
+            end
+
             return collection unless collection.is_a?(Enumerable)
             return collection unless collection.all? { |element| element.is_a?(::ActionView::Template) }
 
--- a/spec/rspec/rails/view_rendering_spec.rb
+++ b/spec/rspec/rails/view_rendering_spec.rb
@@ -165,6 +165,25 @@ module RSpec::Rails
           expect(custom_method_called).to eq(true)
         end
 
+        it "prevents rendering when a custom resolver returns a single template" do
+          template = ActionView::Template.new(
+            "rendered",
+            "custom resolver template",
+            ViewRendering::EmptyTemplateHandler,
+            virtual_path: "custom/template",
+            format: :html,
+            locals: []
+          )
+          resolver = Class.new(ActionView::Resolver) do
+            define_method(:find) { |*| template }
+          end.new
+
+          decorated_template = ViewRendering::EmptyTemplateResolver.build(resolver).find
+
+          expect(decorated_template).to be_an(ActionView::Template)
+          expect(decorated_template.source).to eq("")
+        end
+
         it "works with strings" do
           decorated = false
           ActionController::Base.view_paths = ActionView::PathSet.new(['app/views', 'app/legacy_views'])


pirj (approved):
Thank you!

saidM:
hey @JonRowe, could you please have a look?

PR #2918, opened 2026-08-27
Fix sticky queue_adapter in our specs

pirj:
Reading ActiveJob::Base.queue_adapter before assigning anything started breaking on Rails 8.2-pre

ActiveJob::TestHelper::TestQueueAdapter.adapter_for caches an adapter and makes further assigning to queue_adapter an NOP.

Rails commit: https://github.com/rails/rails/commit/77b6966720d4f2e2cf975de6ae9ccafc0cd0681e

It caused flakiness due to order-dependency. Passes whenever another file, eg matchers/active_job_spec.rb, has already assigned an adapter, and fails when it runs first. Seeds 999 and 12345 fail while eg seed 1 passes.

Only our internal build is affected. `active_job.set_configs` sets

    config.active_job.queue_adapter ||= (Rails.env.test? ? :test : :async)

and assigns it on boot, so an application always has an adapter on the class attribute and a bare read can never fill the cache.

In our specs, ActiveJob::Base._queue_adapter starts with nil.

Drop the read and the assignment, and hand the adapter over through queue_adapter_for_test, the hook ActiveJob::TestHelper calls per example.

Fixes eg https://github.com/rspec/rspec-rails/actions/runs/32901535965/job/97976281083?pr=2915#step:7:96

### DIFF
--- a/spec/rspec/rails/example/rails_example_group_spec.rb
+++ b/spec/rspec/rails/example/rails_example_group_spec.rb
@@ -58,9 +58,7 @@ module RSpec::Rails
     end
 
     it 'will not leak enqueued ActiveJob jobs between examples', skip: !RSpec::Rails::FeatureCheck.has_active_job? do
-      original_adapter = ActiveJob::Base.queue_adapter
       original_logger = ActiveJob::Base.logger
-      ActiveJob::Base.queue_adapter = :test
       ActiveJob::Base.logger = Logger.new(nil)
 
       leaky_job = Class.new(ActiveJob::Base) do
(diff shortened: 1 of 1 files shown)


spec/rspec/rails/example/rails_example_group_spec.rb
```diff
@@ -75,6 +73,10 @@ def self.name
         RSpec::Core::ExampleGroup.describe("A group", order: :defined) do
           include RSpec::Rails::RailsExampleGroup
 
+          def queue_adapter_for_test
+            ActiveJob::QueueAdapters::TestAdapter.new
```
JonRowe, line 77: This breaks the purpose of the spec, each example is a new instance so this is called twice and the queue never shared between the examples, unlike real life where it would be a single queue instance.

JonRowe:
:wave: Thanks for the investigation into the trigger of the bug, however when poking into this myself I realised the problem is the cache prevents the configuration of the queue, I'd argue this is a bug in the rails side, if an adapter is cached in a test its not possible to change it due to the overridden test implementation.

Given the nature of the test simply resetting the cache seems more prudent, closing in favour of #2919

